PR #775, opened 2026-04-10
fix(#765,#761,#762): deploy polish — pull sidecar, default TLS storage path, warn on health timeout

### LINKED ISSUE #761: fix: set default tls.storage_path for letsencrypt
When tls.provider: letsencrypt, tls.storage_path should default to /data/caddy (or wherever the Docker volume maps). Currently the user has to guess the right path.

### LINKED ISSUE #762: fix: deploy health check timeout should warn, not hard-fail
When vibew deploy health check times out, print the last few container logs and suggest vibew deploy logs instead of just exiting 1. The services might be healthy — the check might fail due to TLS/SNI issues.

### LINKED ISSUE #765: fix: vibew deploy should always pull latest sidecar image
docker compose up doesn't pull newer images if a local cache exists. vibew deploy should run docker compose pull for the sidecar image even in build mode, to ensure the latest VibeWarden version is used.

tibtof:
Closes #765
Closes #761
Closes #762

## Summary

- **#765 — always pull latest sidecar image**: `docker compose pull vibewarden` is now run before `docker compose up` in both build mode and image mode. Previously, build mode skipped the sidecar pull entirely, leaving a stale cached image. The insecure HTTP client previously scoped to `waitHealthy` is now built in `NewService` and injected via `httpDo`, which also fixes a pre-existing test bug where the health-check mock was bypassed.

- **#761 — default `tls.storage_path` for letsencrypt**: After unmarshalling in `config.Load`, when `tls.provider` is `letsencrypt` and `storage_path` is empty, it is defaulted to `/root/.local/share/caddy` (Caddy's standard path when running as root in Docker). This suppresses the "storage_path is required for certificate monitoring" warning on first boot without requiring the user to set it manually.

- **#762 — health check timeout warns instead of failing**: `waitHealthy` now returns void and prints a `Warning:` message on timeout or context cancellation instead of returning an error. The deploy still prints "Deploy complete." and the operator can run `vibew deploy status --target ...` to check manually.

## Test plan

- `TestService_Deploy_HealthCheckTimeout` — updated: now asserts `nil` error and presence of "Deploy complete" and "Warning" in output
- `TestService_Deploy_PullsSidecarBeforeBuild` — new: asserts `docker compose pull vibewarden` is called and ordered before `docker compose up --build`
- `TestService_Deploy_PullsSidecarInImageMode` — new: asserts both the targeted sidecar pull and the full pull occur in image mode
- `TestService_Deploy_HealthCheckWarnOnTimeout` — new: asserts warning message and `nil` error
- `TestService_Deploy_BuildMode` — updated: now asserts `docker compose pull vibewarden` is called, and that no broad `docker compose pull` fires in build mode
- `TestLoad_LetsEncryptDefaultStoragePath` — new: table-driven, covers letsencrypt-without-path, letsencrypt-with-explicit-path, and self-signed-without-path

All checks pass locally (`make check`, including `golangci-lint` and `-race` tests).

### DIFF
--- a/internal/app/deploy/service.go
+++ b/internal/app/deploy/service.go
@@ -345,9 +345,10 @@ func (s *Service) waitHealthy(ctx context.Context, healthURL string, out io.Writ
 
 		if time.Now().After(deadline) {
 			if lastErr != nil {
-				fmt.Fprintf(out, "Warning: health check timed out (last error: %v).\n", lastErr)
+				fmt.Fprintf(out, "Warning: health check timed out (last error: %v). Services may still be starting. Run: vibew deploy status --target ... to check.\n", lastErr)
+			} else {
+				fmt.Fprintln(out, "Warning: health check timed out. Services may still be starting. Run: vibew deploy status --target ... to check.")
 			}
-			fmt.Fprintln(out, "Warning: health check timed out. Services may still be starting. Run: vibew deploy status --target ... to check.")
 			return
 		}
 
--- a/internal/app/deploy/service.go
+++ b/internal/app/deploy/service.go
@@ -45,16 +45,23 @@ type Service struct {
 // NewService creates a Service.
 // executor handles SSH commands and rsync transfers.
 // generator is used to produce the .vibewarden/generated/ files before transfer.
-// httpDo is the HTTP function used for status/logs commands; pass nil to use
-// http.DefaultClient. Health checks during deploy use a separate insecure
-// client scoped to waitHealthy (cert may not be issued yet).
+// httpDo is the HTTP function used for all outbound HTTP calls, including the
+// deploy health check. Pass nil to use an insecure HTTP client that tolerates
+// self-signed and not-yet-issued TLS certificates (the default for deployment,
+// where the ACME cert may not have been issued yet).
 func NewService(
 	executor ports.RemoteExecutor,
 	generator ports.ConfigGenerator,
 	httpDo func(req *http.Request) (*http.Response, error),
 ) *Service {
 	if httpDo == nil {
-		httpDo = http.DefaultClient.Do
+		//nolint:gosec // G402: intentional — cert may not be issued yet during deploy
+		httpDo = (&http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
+			},
+			Timeout: 5 * time.Second,
+		}).Do
 	}
 	return &Service{
 		executor:  executor,
@@ -162,7 +169,16 @@ func (s *Service) Deploy(ctx context.Context, cfg *config.Config, opts RunOption
 	}
 
 	// Step 4: start Docker Compose on the remote.
-	// When app.build is set, build the image remotely instead of pulling.
+	// Always pull the latest sidecar image before starting, regardless of mode.
+	// In build mode the app image is built locally, but the sidecar image may be
+	// cached from an older version, so we explicitly pull it first.
+	fmt.Fprintln(out, "Pulling latest sidecar image on remote...")
+	pullSidecarCmd := fmt.Sprintf("cd %s && docker compose pull vibewarden", remoteDir)
+	if _, err := s.executor.Run(ctx, pullSidecarCmd); err != nil {
+		return fmt.Errorf("docker compose pull vibewarden: %w", err)
+	}
+
+	// When app.build is set, build the image remotely instead of pulling the app image.
 	if cfg.App.Build != "" {
 		fmt.Fprintln(out, "Building and starting services on remote...")
 		upCmd := fmt.Sprintf("cd %s && docker compose up -d --build --force-recreate", remoteDir)
@@ -203,9 +219,7 @@ func (s *Service) Deploy(ctx context.Context, cfg *config.Config, opts RunOption
 	}
 	healthURL := fmt.Sprintf("%s://%s:%d/_vibewarden/health", scheme, host, port)
 	fmt.Fprintf(out, "Waiting for sidecar health check at %s...\n", healthURL)
-	if err := s.waitHealthy(ctx, healthURL, out); err != nil {
-		return fmt.Errorf("health check failed: %w", err)
-	}
+	s.waitHealthy(ctx, healthURL, out)
 
 	fmt.Fprintln(out, "Deploy complete.")
 	return nil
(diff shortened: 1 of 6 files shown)


internal/app/deploy/service.go
```diff
@@ -304,38 +318,43 @@ func (s *Service) checkRemotePrerequisites(ctx context.Context) error {
...
 		} else {
 			fmt.Fprintf(out, "  attempt %d: not yet healthy\n", attempt)
 		}
 
 		if time.Now().After(deadline) {
-			return fmt.Errorf("sidecar did not become healthy within %s", healthCheckTimeout)
+			if lastErr != nil {
+				fmt.Fprintf(out, "Warning: health check timed out (last error: %v).\n", lastErr)
+			}
+			fmt.Fprintln(out, "Warning: health check timed out. Services may still be starting. Run: vibew deploy status --target ... to check.")
```
tibtof: **Bug: double Warning line on deadline expiry when last error exists**

When the deadline is exceeded and `lastErr != nil`, two consecutive `Warning:` lines are written:

```
Warning: health check timed out (last error: connection refused).
Warning: health check timed out. Services may still be starting. Run: vibew deploy status --target ... to check.
```

The first line is redundant because the second already covers the general timeout message. An operator reading the output sees two `Warning:` prefixes in a row with different phrasing, which is confusing. The error detail from `lastErr` is useful, but it belongs in the same line as the actionable message. Replace the two-line block with a single message:

```go
if lastErr != nil {
    fmt.Fprintf(out, "Warning: health check timed out (last error: %v). Services may still be starting. Run: vibew deploy status --target ... to check.\n", lastErr)
} else {
    fmt.Fprintln(out, "Warning: health check timed out. Services may still be starting. Run: vibew deploy status --target ... to check.")
}
```

internal/app/deploy/service_test.go
```diff
@@ -934,6 +953,123 @@ func TestProjectNameFromConfig_RelativeDoesNotReturnDot(t *testing.T) {
...
+	err := svc.Deploy(context.Background(), cfg, deployapp.RunOptions{
+		ConfigPath: "/tmp/proj/vibewarden.yaml",
+	})
+	if err != nil {
+		t.Fatalf("Deploy() unexpected error: %v", err)
+	}
+
+	// Both the targeted sidecar pull and the full pull must be present.
+	assertRunCalledContains(t, executor.runCalls, "docker compose pull vibewarden")
+	assertRunCalledContains(t, executor.runCalls, "docker compose pull")
```
tibtof, line 1037: **Test defect: `assertRunCalledContains(t, ..., "docker compose pull")` is a false-positive assertion**

`assertRunCalledContains` does a substring match. The string `"docker compose pull"` is a substring of `"docker compose pull vibewarden"`, so this assertion passes trivially even if the full `docker compose pull` (all-services pull) is never executed.

A correct assertion needs to distinguish the two calls. Use a helper that checks for the exact substring exclusive of the more specific variant, for example:

```go
// assert full pull (not vibewarden-specific)
found := false
for _, c := range executor.runCalls {
    if strings.Contains(c, "docker compose pull") && !strings.Contains(c, "pull vibewarden") {
        found = true
        break
    }
}
if !found {
    t.Error("expected a full 'docker compose pull' call in image mode")
}
```

Additionally, this test never asserts that `pull vibewarden` comes *before* the full pull, unlike `TestService_Deploy_PullsSidecarBeforeBuild`. Add an ordering assertion consistent with the build-mode test.

internal/app/deploy/service_test.go
```diff
@@ -203,11 +206,22 @@ func TestService_Deploy_HealthCheckTimeout(t *testing.T) {
 		cancel()
```
tibtof, line 206: **Test race: context may cancel before waitHealthy is reached**

Both TestService_Deploy_HealthCheckTimeout and TestService_Deploy_HealthCheckWarnOnTimeout cancel the context in a goroutine with no synchronisation. cancel() can fire before Deploy has even entered the health-check loop. If the context is cancelled while fakeExecutor.Run is executing an earlier step (e.g. 'which docker', 'docker compose pull vibewarden'), that step returns a context error and Deploy returns a non-nil error before waitHealthy is ever reached. The test then fails at t.Fatalf("Deploy() should succeed...") with a confusing error from a completely different step.

Fix by using a context with a very short deadline instead of a free-running goroutine cancel. Since fakeExecutor.Run is instantaneous, a 1 ms deadline fires after all synchronous executor calls complete but before the 3-second health-check interval elapses:

  ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
  defer cancel()

This change applies to both TestService_Deploy_HealthCheckTimeout (this function) and TestService_Deploy_HealthCheckWarnOnTimeout.

internal/config/config.go
```diff
@@ -2322,6 +2322,19 @@ func Load(configPath string) (*Config, error) {
...
+	// These cannot be expressed via viper.SetDefault because they depend on
+	// the final resolved value of another key.
+	//
+	// When the TLS provider is letsencrypt and storage_path is not set by the
+	// user, default to Caddy's standard storage path when running as root
+	// inside Docker (/root/.local/share/caddy). This suppresses the
+	// "storage_path is required for certificate monitoring" warning for the
+	// common deployment case.
+	if cfg.TLS.Provider == "letsencrypt" && cfg.TLS.StoragePath == "" {
+		cfg.TLS.StoragePath = "/root/.local/share/caddy"
```
tibtof, line 2335: **Issue #761 spec says /data/caddy — PR uses /root/.local/share/caddy — toggler.go generates ./data/caddy: three different paths with no documented resolution**

Issue #761 specifies the default should be '/data/caddy (or wherever the Docker volume maps)'. The PR's default is '/root/.local/share/caddy'. The existing toggler.go (used by 'vibew init') already writes './data/caddy' into generated configs. These three values are inconsistent:

- Issue spec: /data/caddy
- toggler.go (vibew init output): ./data/caddy
- This PR's config.Load default: /root/.local/share/caddy

The Caddy adapter test in internal/adapters/caddy/config_test.go also uses '/data/caddy' as its expected storage root.

If a user runs 'vibew init' they get './data/caddy' in their config. If they omit storage_path entirely, config.Load now injects '/root/.local/share/caddy'. These are different directories on the container filesystem, so the cert-monitor will look in one place while Caddy actually writes certs to another.

Before merging, one of these must happen:
1. Align toggler.go to write the same path that config.Load defaults to, OR
2. Document in an ADR why /root/.local/share/caddy is the canonical default and update the issue, toggler.go, and the caddy adapter test to match.

The PR description does not address this inconsistency.

internal/config/config_test.go
```diff
@@ -2768,3 +2768,65 @@ func TestIsProdProfile(t *testing.T) {
...
+			wantStoragePath: "/data/caddy",
+		},
+		{
+			name: "self-signed without storage_path stays empty",
+			yaml: `
+tls:
+  enabled: true
+  provider: self-signed
+`,
+			wantStoragePath: "",
```
tibtof, line 2810: **Test missing: no 'external' provider case in TestLoad_LetsEncryptDefaultStoragePath**

The table covers letsencrypt and self-signed, but not 'external'. Although the current code only applies the default when provider == 'letsencrypt', the absence of an external-provider case means future regressions (e.g. someone accidentally broadening the condition to 'provider != self-signed') would go undetected. Add a fourth table entry:

  {
      name: "external provider without storage_path stays empty",
      yaml: "tls:\n  enabled: true\n  provider: external\n  cert_path: /etc/certs/server.crt\n  key_path: /etc/certs/server.key\n",
      wantStoragePath: "",
  }

tibtof (commented):
## VibeWarden Automated Review — CHANGES REQUESTED

Four issues require changes before this can be approved. Inline comments have been posted on each.

### 1. Bug — double Warning line on health-check deadline expiry (service.go:347-350)
When the deadline expires and lastErr is set, two consecutive Warning: lines are printed. The first is a fragment ('Warning: health check timed out (last error: ...)'); the second is the actionable message. They should be combined into a single fmt.Fprintf call that conditionally includes the last error detail. See inline comment at service.go:350.

### 2. Test defect — false-positive assertion in TestService_Deploy_PullsSidecarInImageMode (service_test.go:1037)
assertRunCalledContains(t, ..., "docker compose pull") matches the "docker compose pull vibewarden" call because it is a substring match. The all-services pull is never actually verified: the assertion passes even if that step is deleted from production code. Fix with an explicit negative-substring check, and add an ordering assertion (pull vibewarden before full pull) to match the build-mode test. See inline comment at service_test.go:1037.

### 3. Test race — goroutine cancel in both timeout tests (service_test.go:206)
Both TestService_Deploy_HealthCheckTimeout and TestService_Deploy_HealthCheckWarnOnTimeout spawn a goroutine that calls cancel() with no synchronisation. cancel() can fire before Deploy reaches waitHealthy, causing the deploy to fail at an earlier executor step and the test to report a confusing context-cancelled error from the wrong location. Replace with context.WithTimeout(context.Background(), 1*time.Millisecond), which is race-free since fakeExecutor.Run is instantaneous. See inline comment at service_test.go:206.

### 4. Path inconsistency — three divergent storage paths (config.go:2335)
Issue #761 specifies /data/caddy. toggler.go (vibew init) generates ./data/caddy. The caddy adapter test uses /data/caddy. This PR introduces /root/.local/share/caddy without updating the other references or adding an ADR entry. A user who runs vibew init and then removes storage_path from their config will get a different path injected by config.Load than what their volume mount expects. Align all three to the same canonical value and document the decision. See inline comment at config.go:2335.

---
What is correct and does not need to change:
- Sidecar pull placed before compose up in both modes is correct; the build-mode ordering assertion is well-written.
- The waitHealthy void-return signature and config.Load placement (after unmarshal, before Validate) are sound.
- NewService insecure-client consolidation correctly fixes the pre-existing test bypass bug.
- TestLoad_LetsEncryptDefaultStoragePath table structure is good; it needs only an external-provider case added (see inline comment at config_test.go:2810).

PR #791, opened 2026-04-10
fix(#784): default WAF to enabled=true, mode=detect (log-only)

### LINKED ISSUE #784: fix: WAF should default to detect mode (log-only)

tibtof:
Closes #784

## Summary

- `waf.enabled` default changed from `false` to `true`
- `waf.mode` default changed from `"block"` to `"detect"` (log-only, no blocking)
- Updated godoc comments on `WAFConfig.Enabled` and `WAFConfig.Mode` to reflect new defaults
- Added a `waf` section to the `vibewarden wrap` template so generated configs explicitly show the new defaults and the escape-hatch comment

## Test plan

- `TestLoad_WAFDefaults` — asserts all six WAF default fields against their expected values after loading with no config file
- `TestLoad_WAFModeOverride` — three table-driven cases: explicit `mode: block`, `enabled: false` opt-out, and no WAF section at all (uses defaults)
- `make check` passes: lint, build, and full test suite green

### DIFF
--- a/internal/cli/templates/vibewarden.yaml.tmpl
+++ b/internal/cli/templates/vibewarden.yaml.tmpl
@@ -84,6 +84,16 @@ rate_limit:
     - "/ready"
 {{- end }}
 
+
+waf:
+  enabled: true
+  mode: detect  # "detect" logs attacks without blocking; change to "block" to enforce
+  rules:
+    sqli: true
+    xss: true
+    path_traversal: true
+    command_injection: true
+
 # Escape hatches — uncomment to supply hand-crafted config files instead of
 # relying on VibeWarden's runtime generation. Paths are relative to the
 # .vibewarden/generated/ directory.


tibtof:
## Code Review — CHANGES_REQUESTED

Two issues must be fixed before this can merge. Both are in files that were not touched by this PR but are directly affected by the default change.

---

### Issue 1 — Stale godoc comment in `internal/middleware/waf.go` line 27

The `WAFConfig.Mode` field carries this comment:

```go
// Mode controls whether detections block or only log. Default: WAFModeBlock.
```

This PR changed the config default from `"block"` to `"detect"`, but this comment was not updated. Any caller reading the middleware package in isolation will see a false claim about the default.

**Fix:** change the comment to:

```go
// Mode controls whether detections block or only log. Default: WAFModeDetect.
```

---

### Issue 2 — Inconsistent empty-string fallback in `internal/adapters/caddy/waf_engine_handler.go` lines 93–94

```go
mode := middleware.WAFMode(h.Config.Mode)
if mode == "" {
    mode = middleware.WAFModeBlock
}
```

The fallback here is `WAFModeBlock`. With the new project default of `"detect"`, this path now contradicts the intended behaviour: if `Mode` arrives as an empty string for any reason (e.g. a plugin constructed without going through `config.Load`), the handler silently enforces blocking — the opposite of what the PR intends.

**Fix:** change the fallback to `middleware.WAFModeDetect` so that the adapter is consistent with the config layer.

---

### What is correct in this PR

- `config.go` `Load()` defaults are correct (`waf.enabled=true`, `waf.mode="detect"`).
- Godoc comments on `WAFConfig` in `config.go` are updated.
- The wrap template emits an explicit `waf:` section with `mode: detect` and a helpful inline comment. The unconditional emission (no `{{- if }}` guard) is intentional and appropriate since WAF is now on by default.
- `TestLoad_WAFDefaults` is table-driven, covers all six default fields, and uses descriptive names.
- `TestLoad_WAFModeOverride` covers the three meaningful override cases with temp-file fixtures.
- No new dependencies; no license concerns.

Fix the two stale references and this is ready to merge.

tibtof:
Status: READY_FOR_REVIEW

PR #792, opened 2026-04-10
fix(#781): broaden default CSP to allow unsafe-inline for styles and scripts

### LINKED ISSUE #781: fix: default CSP should allow unsafe-inline for styles
default-src 'self' blocks inline styles. Every app looks broken out of the box. Default to: default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'

tibtof:
Closes #781

## Summary

- Updated `internal/cli/templates/vibewarden.yaml.tmpl` (generated by `vibew wrap`) to use the new default CSP: `default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'`
- Updated `internal/cli/templates/init-vibewarden.yaml.tmpl` (generated by `vibew init`) with the same new default CSP
- Updated `test/e2e/e2e_test.go` to match the new CSP value in both the inline config fixture and the header assertion

The runtime default in `internal/config/config.go` is already empty (CSP opt-in), so no change was needed there. The templates are the source of the broken default that users see immediately after `vibew init` or `vibew wrap`.

## Test plan

- [ ] `go build ./...` — passes
- [ ] `go vet ./...` — passes
- [ ] `golangci-lint run ./...` — 0 issues
- [ ] `go test -race ./...` — all green (confirmed via pre-push hook)
- [ ] Run `vibew init` on a fresh project and verify the generated `vibewarden.yaml` contains the new CSP string
- [ ] Run `vibew wrap` on an existing project and verify the same

### DIFF
--- a/internal/cli/templates/init-vibewarden.yaml.tmpl
+++ b/internal/cli/templates/init-vibewarden.yaml.tmpl
@@ -29,7 +29,7 @@ security_headers:
   hsts_preload: false
   content_type_nosniff: true
   frame_option: "SAMEORIGIN"
-  content_security_policy: "default-src 'self'"
+  content_security_policy: "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'"
   referrer_policy: "strict-origin-when-cross-origin"
   permissions_policy: "geolocation=(), microphone=(), camera=()"
 
--- a/internal/cli/templates/vibewarden.yaml.tmpl
+++ b/internal/cli/templates/vibewarden.yaml.tmpl
@@ -29,7 +29,7 @@ security_headers:
   hsts_preload: false
   content_type_nosniff: true
   frame_option: "SAMEORIGIN"
-  content_security_policy: "default-src 'self'"
+  content_security_policy: "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'"
   referrer_policy: "strict-origin-when-cross-origin"
   permissions_policy: "geolocation=(), microphone=(), camera=()"
 {{- if .TLSEnabled }}
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -81,7 +81,7 @@ security_headers:
   hsts_max_age: 0
   content_type_nosniff: true
   frame_option: "DENY"
-  content_security_policy: "default-src 'self'"
+  content_security_policy: "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'"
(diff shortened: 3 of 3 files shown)


internal/cli/templates/vibewarden.yaml.tmpl
```diff
@@ -29,7 +29,7 @@ security_headers:
   hsts_preload: false
   content_type_nosniff: true
   frame_option: "SAMEORIGIN"
-  content_security_policy: "default-src 'self'"
+  content_security_policy: "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'"
```
tibtof, line 32: **Security regression — `unsafe-inline` on `script-src` defeats XSS protection**: Adding `'unsafe-inline'` to `script-src` allows any inline `<script>` tag to execute, which nullifies the XSS-barrier benefit of CSP for browsers that support it. Issue #781 mentions that inline *styles* break the default look; it does not mention a functional need for inline *scripts*. If the root problem is that the Caddy admin UI or the VibeWarden status page uses inline styles, only `style-src 'unsafe-inline'` is needed.

Suggested safer default:
```
default-src 'self'; style-src 'self' 'unsafe-inline'
```

If there is a concrete reason to allow inline scripts in freshly generated projects, that reason should be documented in an ADR (DECISIONS.md) and referenced in the PR.

internal/cli/templates/init-vibewarden.yaml.tmpl
```diff
@@ -29,7 +29,7 @@ security_headers:
   hsts_preload: false
   content_type_nosniff: true
   frame_option: "SAMEORIGIN"
-  content_security_policy: "default-src 'self'"
+  content_security_policy: "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'"
```
tibtof, line 32: **Stale documentation outside the diff**: Three files that were not changed in this PR now contain documentation that contradicts the new template default:

1. `internal/plugins/securityheaders/meta.go` line 17: ConfigSchema says "default: empty — header not sent; set explicitly to opt in". Line 30: Example() comment says "content_security_policy is empty by default (header not sent)".
2. `internal/plugins/catalog.go` lines 148–149: same stale wording.
3. `vibewarden.reference.yaml` lines 606–610: comment explicitly reads "CSP is disabled by default because 'default-src self' breaks most real apps" and sets `content_security_policy: ""`.

A user running `vibew init` will get an active CSP in their generated file, but the reference config and the plugin introspection output will say CSP is disabled. These three files must be updated in this PR.

tibtof:
Status: READY_FOR_REVIEW

Issue #873, opened 2026-04-17
feat: hot-reload directory watcher for sites.d/


tibtof:
## Parent epic

#869 — Multi-app deployment (Phase 1, Story 4 of 5)

## Problem statement

The existing file watcher (ADR-062) monitors a single `vibewarden.yaml` file. Multi-app requires watching a directory (`sites.d/`) for added, modified, or removed app fragments and triggering a full config merge + Caddy reload when any fragment changes.

## User story

As a vibe coder, I want VibeWarden to automatically detect when I deploy a new app (which places a config fragment in `sites.d/`) and start serving the new subdomain without requiring a manual sidecar restart.

## Acceptance criteria

- [ ] The file watcher watches the `sites.d/` directory recursively for file creation, modification, and deletion events.
- [ ] When a fragment is added or modified, the watcher triggers: (1) re-read all fragments, (2) merge with global config, (3) validate the merged config, (4) rebuild Caddy JSON, (5) call `caddy.Load()` to apply without dropping connections.
- [ ] When a fragment is deleted, the corresponding site's routes are removed from Caddy on the next reload.
- [ ] Validation failure on one fragment does not break other sites — the watcher logs the error and keeps the previous valid config for the failing fragment (or excludes it).
- [ ] Debounce logic applies: rapid successive changes (e.g., rsync writing multiple files) are coalesced into a single reload. Reuse the existing debounce mechanism from ADR-062.
- [ ] A `config.reloaded` domain event (from ADR-062) is emitted on successful reload, with the trigger source set to `"file_watcher"`.
- [ ] A `config.reload_failed` domain event is emitted on validation failure, including the fragment path and error details.
- [ ] The existing single-file watch mode continues to work for single-app deployments. The directory watcher activates only when `sites.d/` exists.
- [ ] Integration test: place a new fragment file in `sites.d/`, assert that within the debounce window the Caddy config is updated to include the new host.

## Out of scope

- Domain model / config loading (Story 1)
- Caddy route generation (Story 2)
- Deploy-time mechanics (Story 3)
- CLI changes (Story 5)

## Dependencies

- Story 1 (multi-site config loader) — provides the merge logic.
- Story 2 (Caddy multi-host routes) — provides the multi-site Caddy config builder.

tibtof:
Status: READY_FOR_ARCH (depends on #870, #871)

tibtof:
Status: READY_FOR_DEV

Design: ADR-071 in DECISIONS.md

[...]

PR #875, opened 2026-04-17
feat(#870): add Site domain model + multi-config loader (ADR-068)

### LINKED ISSUE #870: feat: domain model — Site aggregate and multi-site config loader
## Parent epic

#869 — Multi-app deployment (Phase 1, Story 1 of 5)

## Problem statement

VibeWarden has no domain concept for "a site" — the entire config assumes a single upstream app. Multi-app deployment requires a first-class Site entity and a config loader that can merge multiple per-app fragments with a global config.

## User story

As a vibe coder, I want each app I deploy to be represented as an independent site in VibeWarden's domain model so that each site has its own upstream, domain, auth, and rate-limit configuration without interfering with other sites.

## Acceptance criteria

- [ ] A `Site` entity (or value object) exists in `internal/domain/` with at minimum: name, domain, upstream address, and a reference to its per-site plugin config (auth mode, rate limit, security headers, etc.).
- [ ] A `SiteCollection` or equivalent aggregate can hold multiple Sites and enforce invariants (no duplicate domains, no duplicate site names).
- [ ] A config loader function can read a directory of YAML fragments (`sites.d/*.yaml`) and produce a `SiteCollection`.
- [ ] A global config file (`vibewarden-global.yaml`) provides defaults that per-site fragments inherit from but can override (except for global-only settings like admin token).
- [ ] Validation: the loader rejects fragments with missing required fields (domain, upstream), duplicate domains across fragments, or invalid YAML.
- [ ] The existing single-site config path (`vibewarden.yaml` with no `sites.d/`) continues to work unchanged — backward compatibility is mandatory.
- [ ] Unit tests cover: single-site fallback, multi-site loading, merge/override behavior, duplicate domain rejection, missing field rejection.

## Out of scope

- Caddy route generation (Story 2)
- Deploy-time file placement (Story 3)
- File-watching / hot-reload (Story 4)
- CLI changes (Story 5)

## Dependencies

None — this is the foundation story. Stories 2-5 depend on this.

## Open questions for architect

- What is the exact YAML schema for a site fragment? Should it be the full `vibewarden.yaml` shape (so existing configs work as fragments) or a trimmed-down subset?
- Which config fields are global-only (cannot be overridden per-site)?

tibtof:
Closes #870

## Summary

- **Domain types** (`internal/domain/site/`): `Status` enum (Healthy/Error/Degraded), `GlobalConfig` value object with validation (5 fields), `Site` entity wrapping `*config.Config` with DNS-safe name validation, `Registry` aggregate (thread-safe via `sync.RWMutex`) with duplicate-domain validation
- **Config loaders** (`internal/config/sites/`): `LoadGlobal(path)` reads `global.yaml` with defaults fallback, `LoadSites(basePath)` walks `sites/*/vibewarden.yaml` with partial-success semantics (broken configs produce error sites, healthy configs load normally)
- **Backward compat**: absent `sites/` directory returns `(nil, nil)` for single-app mode; absent `global.yaml` returns defaults
- **`.gitignore` fix**: narrowed `site/` to `/site/` so it only matches the top-level mkdocs build dir, not `internal/domain/site/`

### ADR deviation

The loaders live in `internal/config/sites/` (a sub-package) instead of `internal/config/` as the architect spec'd. Reason: `domain/site` imports `config` for `*config.Config` (sanctioned boundary violation), so `config` cannot import `domain/site` without creating an import cycle. The separate sub-package `config/sites` breaks the cycle while keeping loaders co-located with the config layer. This is the minimal structural change that respects Go's import rules.

## Test plan

- [ ] `make check` passes (lint, build, all tests)
- [ ] `go test -race -v ./internal/domain/site/...` -- 57 test cases: name validation edge cases, Site/ErrorSite construction, Registry CRUD, concurrent access (50 goroutines x 100 ops), duplicate domain detection
- [ ] `go test -race -v ./internal/config/sites/...` -- 37 test cases: global.yaml loading/defaults/validation, multi-site loading, partial success (2 valid + 1 broken), backward compat (no sites dir), integration tests with realistic directory tree


internal/domain/site/site.go
```diff
@@ -0,0 +1,116 @@
...
+type Site struct {
+	name   string
+	cfg    *config.Config
+	status Status
+	err    error
+}
+
+// NewSite creates a healthy Site with the given name and configuration.
+// The name must match ValidNamePattern (DNS-safe: lowercase alphanumeric
+// and hyphens, 1-63 chars, no leading/trailing hyphen). The config must
```
tibtof, line 33: **ADR-068 deviation -- missing `configPath` field**: ADR-068 specifies that `Site` has a `configPath string` field, and both constructors accept it (`NewSite(name, cfg, configPath)`, `NewErrorSite(name, configPath, err)`). This field is needed by #873 (directory watcher) to know which file to watch for changes, and by hot-reload to re-read the correct file.

Omitting it now means a breaking constructor signature change later. Please add the `configPath` field and the corresponding `ConfigPath()` accessor as specified in the ADR.

internal/domain/site/registry.go
```diff
@@ -0,0 +1,151 @@
...
+		sites: make(map[string]Site),
+	}
+}
+
+// SetGlobal stores the global configuration in the registry.
+func (r *Registry) SetGlobal(g GlobalConfig) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.global = &g
+}
```
tibtof, line 30: **ADR-068 deviation -- `All()` must return alphabetically-sorted sites**: The ADR specifies: "`All` returns a snapshot of all sites, ordered alphabetically by name." This implementation iterates over a `map[string]Site`, which yields non-deterministic order in Go.

Add `sort.Slice(result, func(i, j int) bool { return result[i].Name() < result[j].Name() })` before returning. Downstream consumers (admin API, status output) will rely on stable ordering.

internal/domain/site/registry.go
```diff
@@ -0,0 +1,151 @@
+package site
+
+import (
+	"fmt"
+	"sync"
+)
+
```
tibtof, line 8: **Design issue -- value-type `Site` in map vs pointer-receiver mutators**: The registry stores `map[string]Site` (by value), but `SetStatus` and `SetErr` are pointer receivers on `*Site`. This means calling `s, _ := r.Get("app1"); s.SetErr(err)` mutates a **copy**, leaving the registry's entry unchanged.

ADR-068 specifies `map[string]*Site` and `[]*Site` return types throughout. Either:
1. Store `*Site` in the map (matching the ADR), so mutations through pointer receivers propagate, or
2. Remove `SetStatus`/`SetErr` entirely if `Site` is meant to be immutable after construction (replace-via-`Add` for state transitions).

Option 1 aligns with the ADR and is simpler. Option 2 is more functional but deviates from the ADR's constructor design.

tibtof:
## Code Review -- REQUEST CHANGES (3 issues: 1 critical, 2 major)

**CRITICAL: Missing `configPath` field on `Site` entity.** ADR-068 specifies `configPath string` as a field and both constructors accept it (`NewSite(name, cfg, configPath)`, `NewErrorSite(name, configPath, err)`). Story #873 (directory watcher) will need this to know which file to watch. Adding it later breaks the constructor signature for all call sites. Add it now per the ADR spec.

**MAJOR: `All()` returns unsorted results.** ADR-068 specifies: "`All` returns a snapshot of all sites, ordered alphabetically by name." The current implementation iterates a `map[string]Site`, which yields non-deterministic order in Go. Add `sort.Slice` before returning.

**MAJOR: Value-type `Site` in `map[string]Site` vs pointer-receiver mutators.** `SetStatus`/`SetErr` are pointer receivers on `*Site`, but `Registry.Get()` returns a `Site` by value (a copy). Mutations through the returned value do not propagate to the registry. ADR-068 specifies `map[string]*Site` and `[]*Site` return types throughout. Either switch to `*Site` storage (matching the ADR) or remove the mutator methods entirely.

---

**What looks good:** The `internal/config/sites/` sub-package to break the import cycle is well-justified. The 50-goroutine concurrent test covers Add/Remove/Get/All/Len simultaneously. Partial-success loading is correctly tested (empty dir, all valid, all invalid, mixed, missing dir, permission errors via invalid dir names). Backward compat (no `sites/` dir = nil, nil) is verified. Name validation covers all ADR-068 edge cases. GlobalConfig defaults, validation, and YAML loading are thorough. The `.gitignore` fix (`site/` to `/site/`) prevents the domain package directory from being ignored.

See inline comments on `internal/domain/site/site.go` and `internal/domain/site/registry.go` for details.

tibtof:
## Code Review -- APPROVED

All three previously-flagged issues confirmed fixed in 8c2f2b1:

1. **configPath field** -- `Site` struct has `configPath string` field. Both `NewSite(name, configPath, cfg)` and `NewErrorSite(name, configPath, err)` accept it. `ConfigPath()` accessor is present.
2. **Sorted returns** -- `All()`, `HealthySites()`, and `ErrorSites()` all use `sort.Slice` to sort by name before returning.
3. **Pointer-type in map** -- `Registry.sites` is `map[string]*Site`. All return types use `[]*Site`. Both constructors return `(*Site, error)`. `LoadSites` returns `[]*site.Site`.

Full checklist re-verified: architecture, code quality, testing, Go idioms, security, and licenses all pass. Ready for human review.

PR #904, opened 2026-04-18
feat(#899): add built-in AES-256-GCM encrypted secret store

### LINKED ISSUE #899: feat: built-in encrypted secret store — eliminate OpenBao dependency for common cases
## Context

VibeWarden requires a separate OpenBao container for secret management. This undermines the "single binary, zero-to-secure" promise:

- OpenBao is MPL-2.0 — not in our approved license list, cannot be embedded as a library
- Even if licensed differently, OpenBao is a full server — not an embeddable Go library
- Most users need 3-5 secrets (DB password, admin token, API key), not a full Vault
- The OpenBao container + bootstrap script adds operational complexity to every deployment

### Tier 1 — Built-in encrypted KV (default, 90% of users)

A simple encrypted key-value store embedded in the VibeWarden binary:

- **Storage**: encrypted file at \`.vibewarden/secrets.db\` (AES-256-GCM, master key from env var or keyfile)
- **CLI**: same surface as today — \`vibew secret set\`, \`vibew secret get\`, \`vibew secret list\`
- **Zero containers** — no OpenBao, no bootstrap script
- **Locked at rest** — decrypted on sidecar startup only
- **License-clean**: \`crypto/aes\` + \`crypto/cipher\` (stdlib), optionally BadgerDB (Apache 2.0) or plain encrypted JSON

### Tier 2 — OpenBao (opt-in, power users)

Keep OpenBao for users who need dynamic credentials, rotation policies, audit trails, cloud KMS integration.

\`\`\`yaml
secrets:
  store: builtin          # default — encrypted file, no container
  # store: openbao        # opt-in — requires separate OpenBao container
\`\`\`

## Architecture implications

- \`ports.SecretStore\` interface stays the same — both backends implement it
- New adapter: \`internal/adapters/secrets/builtin.go\`
- Docker Compose generation skips OpenBao when \`store: builtin\`
- \`.credentials\` file fallback replaced by the built-in store (cleaner)
- The \`tryOpenBao\` → \`.credentials\` fallback chain (#812, #832) simplifies

## Acceptance criteria (Phase 1)

- [ ] \`vibew secret set mykey myvalue\` stores encrypted secret without OpenBao
- [ ] \`vibew secret get mykey\` retrieves it
- [ ] Sidecar starts and injects secrets without any external container
- [ ] \`vibew init\` default config uses \`store: builtin\`
- [ ] Existing \`store: openbao\` configs still work unchanged
- [ ] No MPL-2.0 or unapproved-license dependencies added

tibtof:
Closes #899

## Summary

- **New adapter** `internal/adapters/builtin/` — AES-256-GCM encrypted JSON file store implementing `ports.SecretStore` (Get, Put, Delete, List, Health)
- **Config** — added `secrets.store` field (`"builtin"` | `"openbao"`, default `"builtin"`), `secrets.builtin.path` and `secrets.builtin.key_file` to `SecretsConfig`
- **Plugin wiring** — updated `internal/plugins/secrets/plugin.go` to select builtin vs openbao based on `secrets.store`, refactored `Plugin.store` from `*openbao.Adapter` to `ports.SecretStore` interface
- **CLI** — added `vibew secret set <path> <key=value>...` command; updated `buildSecretService` to support both store backends
- **Domain** — added `SourceBuiltin` constant to `SecretSource`

## Key design decisions

- Master key: hex-encoded 32-byte key from `VIBEWARDEN_SECRETS_MASTER_KEY` env var or `secrets.builtin.key_file`
- Atomic writes via temp file + `os.Rename`
- Thread-safe via `sync.RWMutex`
- Zero external dependencies — stdlib `crypto/aes`, `crypto/cipher`, `crypto/rand` only
- Builtin is the new default; existing OpenBao configs continue to work unchanged via `secrets.store: openbao`

## Test plan

- [x] 20 unit tests for the builtin adapter (encrypt/decrypt round-trip, concurrent access, atomic writes, wrong key detection, persistence, copy semantics, empty file, nested dirs)
- [x] Existing secrets plugin tests updated and passing (all use `Store: "openbao"` explicitly)
- [x] Domain tests updated for `SourceBuiltin`
- [x] CLI tests passing
- [x] `make check` green (golangci-lint 0 issues, go build, go test -race)

### DIFF
--- a/internal/cli/cmd/secret_get.go
+++ b/internal/cli/cmd/secret_get.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -160,31 +159,10 @@ func buildBuiltinStore(cfg *config.Config) (ports.SecretStore, error) {
 
 // resolveBuiltinMasterKey reads the master key from the key file or the
 // VIBEWARDEN_SECRETS_MASTER_KEY environment variable.
+// resolveBuiltinMasterKey delegates to the shared builtin.ResolveMasterKey
+// to avoid duplicating key-resolution logic between CLI and plugin.
 func resolveBuiltinMasterKey(cfg *config.Config) ([]byte, error) {
-	var hexKey string
-
-	if cfg.Secrets.Builtin.KeyFile != "" {
-		raw, err := os.ReadFile(cfg.Secrets.Builtin.KeyFile)
-		if err != nil {
-			return nil, fmt.Errorf("reading key file %q: %w", cfg.Secrets.Builtin.KeyFile, err)
-		}
-		hexKey = strings.TrimSpace(string(raw))
-	} else {
-		hexKey = os.Getenv("VIBEWARDEN_SECRETS_MASTER_KEY")
-	}
-
-	if hexKey == "" {
-		return nil, fmt.Errorf("master key not set")
-	}
-
-	key, err := hex.DecodeString(hexKey)
-	if err != nil {
-		return nil, fmt.Errorf("decoding hex master key: %w", err)
-	}
-	if len(key) != 32 {
-		return nil, fmt.Errorf("master key must be 32 bytes (64 hex chars), got %d bytes", len(key))
-	}
-	return key, nil
+	return builtinstore.ResolveMasterKey(cfg.Secrets.Builtin.KeyFile)
 }
 
 // buildOpenBaoStore creates an OpenBao adapter from config.
--- a/internal/cli/cmd/secret_get.go
+++ b/internal/cli/cmd/secret_get.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	builtinstore "github.com/vibewarden/vibewarden/internal/adapters/builtin"
 	credentialsadapter "github.com/vibewarden/vibewarden/internal/adapters/credentials"
 	openbao "github.com/vibewarden/vibewarden/internal/adapters/openbao"
 	appsecret "github.com/vibewarden/vibewarden/internal/app/secret"
@@ -92,9 +94,11 @@ Examples:
 // defaultOutputDir is the standard generated files directory.
 const defaultOutputDir = ".vibewarden/generated"
 
-// buildSecretService constructs the SecretService, wiring OpenBao and credentials adapters.
-// If OpenBao is not configured in vibewarden.yaml, a nil SecretStore is passed
-// to the service (falling back to .credentials for all lookups).
+// buildSecretService constructs the SecretService, wiring the appropriate store
+// adapter and the credentials file adapter. The store is selected based on
+// config: "builtin" (default) uses an AES-256-GCM encrypted file, "openbao"
+// uses the OpenBao HTTP adapter. When no config exists, credentials-file-only
+// mode is used.
 func buildSecretService(configPath, outputDir string) (*appsecret.Service, error) {
 	if outputDir == "" {
 		outputDir = defaultOutputDir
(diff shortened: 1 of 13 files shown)


internal/cli/cmd/secret_set.go
```diff
@@ -0,0 +1,93 @@
...
+
+// parseKeyValueArgs parses a list of "key=value" strings into a map.
+// Returns an error if any argument is not in "key=value" format.
+func parseKeyValueArgs(args []string) (map[string]string, error) {
+	data := make(map[string]string, len(args))
+	for _, arg := range args {
+		idx := strings.Index(arg, "=")
+		if idx < 1 {
+			return nil, fmt.Errorf("invalid key=value format: %q (expected key=value)", arg)
+		}
```
tibtof, line 77: **Swallowed error**: `loadConfigForCLI` returns an empty config on *any* `config.Load` error, not just file-not-found. A malformed YAML file (syntax error, invalid field types) would be silently ignored, causing the builtin store to quietly fall back to env-var-only mode. The user gets no indication their config is broken.

Fix: Only suppress `os.ErrNotExist`. Propagate all other errors:

```go
func loadConfigForCLI(configPath string) (*config.Config, error) {
    cfg, err := config.Load(configPath)
    if err != nil {
        if errors.Is(err, os.ErrNotExist) {
            return &config.Config{}, nil
        }
        return nil, fmt.Errorf("loading config: %w", err)
    }
    return cfg, nil
}
```

internal/cli/cmd/secret_set.go
```diff
@@ -0,0 +1,93 @@
+package cmd
```
tibtof, line 1: **Missing tests**: CLAUDE.md requires every new `.go` file to have a corresponding `_test.go`. `secret_set.go` has no test file. At minimum, `parseKeyValueArgs` needs table-driven tests (valid pairs, missing `=`, empty key, empty value, multiple `=` in value) and `loadConfigForCLI` needs tests for the file-not-found vs malformed-config distinction.

Add `internal/cli/cmd/secret_set_test.go` with unit tests for `parseKeyValueArgs` and `loadConfigForCLI`.

internal/cli/cmd/secret_get.go
```diff
@@ -106,24 +110,102 @@ func buildSecretService(configPath, outputDir string) (*appsecret.Service, error
...
+	store := cfg.Secrets.Store
+	if store == "" {
+		store = "builtin"
+	}
+
+	switch store {
+	case "builtin":
+		return buildBuiltinStore(cfg)
+	case "openbao":
+		return buildOpenBaoStore(cfg), nil
```
tibtof, line 132: **DRY violation**: `resolveBuiltinMasterKey` here is nearly identical to `resolveMasterKey` in `internal/plugins/secrets/plugin.go` (lines 170-196 in the diff). Both read from key file or env var, hex-decode, and validate length. If the resolution logic changes (e.g., adding KMS support), both sites must be updated in lockstep.

Extract a shared function into the `builtin` package (or a new `internal/adapters/builtin/keyresolver.go`), e.g.:

```go
// ResolveKeyFromHex decodes a hex-encoded master key, validating it is 32 bytes.
func ResolveKeyFromHex(keyFile, envVar string) ([]byte, error) { ... }
```

Then call it from both the plugin and the CLI.

tibtof (commented):
REQUEST CHANGES -- 3 issues found (1 critical, 2 major).

**Crypto, atomic writes, thread safety, port compliance, config wiring, master key security: all good.** The AES-256-GCM implementation is correct (random nonce via crypto/rand, nonce prepended to ciphertext, proper length validation). Atomic writes use temp+rename. RWMutex discipline is correct (RLock for Get/List, Lock for Put/Delete). All 5 SecretStore methods implemented. No secrets logged. No new external deps.

**Issues to fix:**

1. CRITICAL -- `loadConfigForCLI` swallows all config errors (secret_set.go:75). A malformed YAML is silently ignored, falling back to defaults with no warning. Only suppress `os.ErrNotExist`.

2. MAJOR -- `secret_set.go` has no corresponding `_test.go` (CLAUDE.md requirement). `parseKeyValueArgs` and `loadConfigForCLI` need table-driven tests.

3. MAJOR -- `resolveMasterKey` logic is duplicated between `plugin.go` and `secret_get.go`. Extract to a shared function in the `builtin` package.

The adapter code itself (store.go + store_test.go with 20 tests) is solid. The 3 issues above are all in the CLI/plugin wiring layer.

PR #915, opened 2026-04-18
fix(#906): include WAF handlers in vibew eject output

### LINKED ISSUE #906: CRITICAL: WAF is a Potemkin feature — config accepted but never enforced in Caddy route chain
## From v0.11.0 retrospective

WAF is in \`vibew plugins\`, the schema, the docs, \`AGENTS-VIBEWARDEN.md\` boasts about it — but \`vibew eject\` confirms **no \`vibewarden_waf\` handler is ever inserted into the Caddy route chain** regardless of config shape.

\`vibew validate\` accepts WAF config. The agent confidently tells the user "WAF is enabled" based on the exit code. The runtime silently does nothing.

For a product whose value proposition is "security sidecar," shipping a non-functional WAF while documenting it as functional is the single worst bug.

## Fix

Either make WAF work in the route chain, or remove it from the schema/docs/plugins/AGENTS file until it does. Zero tolerance for features that accept config and no-op.

## Acceptance

- [ ] \`vibew eject\` shows a WAF handler in the Caddy route chain when WAF is configured
- [ ] OR: WAF removed from schema/docs/plugins until implemented

tibtof:
Closes #906

## Summary
- **Eject fix**: `buildProxyConfig()` in `internal/app/eject/eject.go` now calls `buildWAFHandlers()` which constructs WAF content-type and engine handler JSON and populates `ExtraHandlers`. This ensures `vibew eject` includes both `vibewarden_waf_content_type` and `vibewarden_waf_engine` handlers in the generated Caddy config when WAF is enabled.
- **Exported builders**: `BuildContentTypeHandlerJSON` and `BuildEngineHandlerJSON` in `internal/plugins/waf/plugin.go` are now exported so the eject code path can reuse them without initialising the full plugin registry.
- **Catalog fix**: The WAF entry in `internal/plugins/catalog.go` now documents all engine config fields (`enabled`, `mode`, `rules.sqli`, `rules.xss`, `rules.path_traversal`, `rules.command_injection`, `exempt_paths`) alongside the existing content-type validation fields. The example YAML shows the full WAF config.

## Test plan
- [x] `TestService_Eject_WAFHandlersIncluded` — table-driven test covering: both features enabled, engine-only, content-type-only, both disabled
- [x] `TestService_Eject_WAFEngineHandlerConfig` — verifies engine handler has correct name and config payload
- [x] `TestCatalog_WAFDescriptorIncludesEngineFields` — verifies all 9 expected config schema fields are present
- [x] All existing tests continue to pass
- [x] `make check` passes (golangci-lint 0 issues, all tests green)

### DIFF
--- a/internal/app/eject/eject.go
+++ b/internal/app/eject/eject.go
@@ -5,9 +5,11 @@ package eject
 
 import (
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/vibewarden/vibewarden/internal/config"
+	wafplugin "github.com/vibewarden/vibewarden/internal/plugins/waf"
 	"github.com/vibewarden/vibewarden/internal/ports"
 )
 
@@ -132,7 +134,8 @@ func buildProxyConfig(cfg *config.Config) *ports.ProxyConfig {
 			Addresses:         cfg.IPFilter.Addresses,
 			TrustProxyHeaders: cfg.IPFilter.TrustProxyHeaders,
 		},
-		Resilience: buildResilienceConfig(cfg),
+		Resilience:    buildResilienceConfig(cfg),
+		ExtraHandlers: buildWAFHandlers(cfg),
 		Compression: ports.CompressionConfig{
 			Enabled:    cfg.Compression.Enabled,
 			Algorithms: cfg.Compression.Algorithms,
@@ -149,6 +152,57 @@ func buildProxyConfig(cfg *config.Config) *ports.ProxyConfig {
 		// these internal HTTP servers are managed by VibeWarden and have no
 		// equivalent in a standalone Caddy deployment.
 	}
+}
+
+// buildWAFHandlers constructs the WAF Caddy handler fragments from the app
+// config, replicating the same handler JSON that the WAF plugin produces at
+// runtime. This allows "vibew eject" to include WAF handlers without
+// initialising the full plugin registry.
+//
+// Returns nil when both WAF features are disabled.
+func buildWAFHandlers(cfg *config.Config) []ports.CaddyHandler {
+	var handlers []ports.CaddyHandler
+
+	if cfg.WAF.ContentTypeValidation.Enabled {
+		handler, err := wafplugin.BuildContentTypeHandlerJSON(wafplugin.ContentTypeValidationConfig{
+			Enabled: cfg.WAF.ContentTypeValidation.Enabled,
+			Allowed: cfg.WAF.ContentTypeValidation.Allowed,
+		})
+		if err == nil {
+			handlers = append(handlers, ports.CaddyHandler{
+				Handler:  handler,
+				Priority: 25,
+			})
+		}
+	}
+
+	if cfg.WAF.Enabled {
+		handler, err := wafplugin.BuildEngineHandlerJSON(wafplugin.WAFEngineConfig{
+			Enabled: cfg.WAF.Enabled,
+			Mode:    wafplugin.Mode(cfg.WAF.Mode),
+			Rules: wafplugin.RulesConfig{
+				SQLInjection:     cfg.WAF.Rules.SQLInjection,
+				XSS:              cfg.WAF.Rules.XSS,
+				PathTraversal:    cfg.WAF.Rules.PathTraversal,
+				CommandInjection: cfg.WAF.Rules.CommandInjection,
+			},
+			ExemptPaths: cfg.WAF.ExemptPaths,
+		})
+		if err == nil {
+			handlers = append(handlers, ports.CaddyHandler{
+				Handler:  handler,
+				Priority: 25,
+			})
+		}
+	}
+
+	// Sort by ascending priority for deterministic ordering, matching the
+	// behaviour of the serve code path in wiring_serve_helpers.go.
+	sort.Slice(handlers, func(i, j int) bool {
+		return handlers[i].Priority < handlers[j].Priority
+	})
+
+	return handlers
 }
 
 // buildBodySizeConfig converts the app config body size settings into a
--- a/internal/app/eject/eject.go
+++ b/internal/app/eject/eject.go
@@ -5,11 +5,9 @@ package eject
 
 import (
 	"fmt"
-	"sort"
 	"time"
 
 	"github.com/vibewarden/vibewarden/internal/config"
-	wafplugin "github.com/vibewarden/vibewarden/internal/plugins/waf"
 	"github.com/vibewarden/vibewarden/internal/ports"
 )
 
@@ -53,15 +51,19 @@ func NewService(builder ConfigBuilder) *Service {
 // configuration map. The returned map can be serialised to JSON and fed
 // directly to the target proxy (e.g. Caddy's /load API or a config file).
 //
+// extraHandlers are Caddy handler fragments contributed by plugins (e.g. the
+// WAF plugin). The caller is responsible for building these via the plugin
+// registry, keeping the application layer free of plugin-specific imports.
+//
 // Note: internal-only addresses (metrics, admin, readiness) are omitted from
 // the generated config because those internal HTTP servers are managed by
 // VibeWarden itself and are not meaningful outside of it.
-func (s *Service) Eject(cfg *config.Config) (map[string]any, error) {
+func (s *Service) Eject(cfg *config.Config, extraHandlers []ports.CaddyHandler) (map[string]any, error) {
 	if cfg == nil {
 		return nil, fmt.Errorf("config is required")
 	}
 
-	proxyCfg := buildProxyConfig(cfg)
+	proxyCfg := buildProxyConfig(cfg, extraHandlers)
 	result, err := s.builder.Build(proxyCfg)
 	if err != nil {
 		return nil, fmt.Errorf("building proxy config: %w", err)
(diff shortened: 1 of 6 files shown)


internal/app/eject/eject.go
```diff
@@ -5,9 +5,11 @@ package eject
 
 import (
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/vibewarden/vibewarden/internal/config"
```
tibtof, line 10: **Architecture violation**: `internal/app/eject/eject.go` imports `internal/plugins/waf` — this is the first time an `internal/app/` package depends on a specific plugin implementation. The app layer should orchestrate through ports/interfaces, not call plugin internals directly.

In the `serve` path, `ExtraHandlers` are collected generically via `registry.CaddyContributors()` and `ContributeCaddyHandlers()`. The eject path should follow the same pattern.

**Why it matters**: This coupling means that every future plugin that contributes Caddy handlers will need a hand-coded case in the eject service. It defeats the purpose of the `CaddyContributor` interface and hexagonal architecture.

**Suggested fix**: The eject service should accept (or construct) a `plugins.Registry` (or a subset interface) and call `CaddyContributors()` on it, just like the serve path does. This way eject automatically picks up handlers from *all* current and future plugins without per-plugin code. The registry can be initialised with `Init()` but without `Start()` since eject only needs the handler JSON, not a running plugin.

Alternatively, if bringing in the full registry is too heavy for eject, define a smaller interface in `internal/ports/` like:

```go
// HandlerContributor builds Caddy handler JSON fragments.
type HandlerContributor interface {
    ContributeCaddyHandlers() []CaddyHandler
}
```

and have the eject service accept a `[]HandlerContributor` that is assembled in the composition root (`cmd/vibewarden/`), not in the app layer.

internal/app/eject/eject.go
```diff
@@ -151,6 +154,57 @@ func buildProxyConfig(cfg *config.Config) *ports.ProxyConfig {
...
+//
+// Returns nil when both WAF features are disabled.
+func buildWAFHandlers(cfg *config.Config) []ports.CaddyHandler {
+	var handlers []ports.CaddyHandler
+
+	if cfg.WAF.ContentTypeValidation.Enabled {
+		handler, err := wafplugin.BuildContentTypeHandlerJSON(wafplugin.ContentTypeValidationConfig{
+			Enabled: cfg.WAF.ContentTypeValidation.Enabled,
+			Allowed: cfg.WAF.ContentTypeValidation.Allowed,
+		})
```
tibtof: **Swallowed error**: When `BuildContentTypeHandlerJSON` returns an error, it is silently ignored — the handler is simply not added and no error is propagated or logged. This violates the project rule "never swallow errors" from CLAUDE.md.

The same issue exists for `BuildEngineHandlerJSON` below (line 183).

**Why it matters**: If the JSON marshaling fails (unlikely but possible with a malformed config), `vibew eject` will silently produce a Caddy config without WAF handlers and the operator will have no idea why. This is especially bad given that issue #906 was specifically about WAF silently doing nothing.

**Suggested fix**: Return the error from `buildWAFHandlers`, propagate it up through `buildProxyConfig`, and ultimately surface it to the user via the `Eject` method:

```go
func buildWAFHandlers(cfg *config.Config) ([]ports.CaddyHandler, error) {
    // ...
    handler, err := wafplugin.BuildContentTypeHandlerJSON(...)
    if err != nil {
        return nil, fmt.Errorf("building WAF content-type handler: %w", err)
    }
    // ...
}
```

Note: this matches the serve path where the WAF plugin logs the error via `p.logger.Error(...)`. In the eject path there is no logger, so returning the error is the correct approach.

internal/app/eject/eject.go
```diff
@@ -151,6 +154,57 @@ func buildProxyConfig(cfg *config.Config) *ports.ProxyConfig {
...
+			ExemptPaths: cfg.WAF.ExemptPaths,
+		})
+		if err == nil {
+			handlers = append(handlers, ports.CaddyHandler{
+				Handler:  handler,
+				Priority: 25,
+			})
+		}
+	}
+
```
tibtof: **Redundant sort on a fixed-priority slice**: Both handlers are assigned `Priority: 25` unconditionally. Sorting a slice where all elements have the same priority is a no-op. The added `"sort"` import exists solely for this dead code path.

**Why it matters**: This adds unnecessary complexity and a misleading comment ("matching the behaviour of the serve code path"). In the serve path, sorting is necessary because *multiple plugins with different priorities* contribute handlers. Here, only WAF handlers are present and they all share priority 25.

**Suggested fix**: If the architecture fix from the earlier comment is applied (using the registry/contributor interface), sorting will happen naturally in the caller. If this per-plugin approach is kept, remove the sort and the `"sort"` import.

internal/plugins/waf/plugin.go
```diff
@@ -163,9 +163,11 @@ type contentTypeHandlerConfig struct {
 	Allowed []string `json:"allowed"`
 }
 
-// buildContentTypeHandlerJSON serialises cfg into the Caddy handler JSON map
-// expected by the vibewarden_waf_content_type module.
-func buildContentTypeHandlerJSON(cfg ContentTypeValidationConfig) (map[string]any, error) {
+// BuildContentTypeHandlerJSON serialises cfg into the Caddy handler JSON map
+// expected by the vibewarden_waf_content_type module. It is exported so that
+// the eject use case can replicate the same handler without initialising the
```
tibtof: **Unnecessary export if architecture is fixed**: These functions were made public solely to support the direct-call approach in the eject service. If the eject service is refactored to use the `CaddyContributor` interface (see comment on `eject.go`), the WAF plugin already implements `ContributeCaddyHandlers()` and these functions can remain unexported.

**Why it matters**: Exporting internal builder functions increases the plugin's public API surface and creates a maintenance burden — any change to the JSON schema now needs to consider two call sites instead of one.

**Suggested fix**: If the architecture is fixed, revert these to lowercase (`buildContentTypeHandlerJSON`, `buildEngineHandlerJSON`). The eject path would call `ContributeCaddyHandlers()` on the plugin instance, which internally calls these unexported functions.

tibtof (commented):
## Review summary — PR #915 (fix #906: WAF handlers in vibew eject)

**Verdict: CHANGES REQUESTED** (cannot use request-changes on own PR, but these must be addressed before merge)

### What was reviewed

- `internal/app/eject/eject.go` — new `buildWAFHandlers()` function
- `internal/app/eject/eject_test.go` — 3 new test functions (table-driven + standalone)
- `internal/plugins/waf/plugin.go` — exported `BuildContentTypeHandlerJSON` and `BuildEngineHandlerJSON`
- `internal/plugins/catalog.go` — WAF descriptor updated with engine fields
- `internal/plugins/catalog_test.go` — new test for WAF descriptor completeness

### Issues found (3 issues, 1 blocking)

**1. BLOCKING — Architecture violation: app layer imports plugin implementation**

`internal/app/eject/eject.go` imports `internal/plugins/waf`. No other `internal/app/` package imports a specific plugin package — this is the first such coupling. The serve path collects ExtraHandlers generically via `registry.CaddyContributors()`. The eject path should follow the same pattern: accept a `[]ports.CaddyContributor` (or `[]ports.CaddyHandler` pre-computed) and let the composition root (`cmd/vibewarden/`) wire the plugins. This also makes the fix future-proof — any plugin that contributes handlers will be picked up automatically without modifying the eject service.

**2. Swallowed errors in `buildWAFHandlers`**

When `BuildContentTypeHandlerJSON` or `BuildEngineHandlerJSON` return errors, the errors are silently ignored (the handler is simply not appended). This violates the "never swallow errors" rule and is ironic given that issue #906 was about WAF *silently* doing nothing. The errors must be propagated to the caller.

**3. Minor — Unnecessary sort on same-priority slice**

Both handlers have `Priority: 25`. Sorting is a no-op. If the architecture fix is applied, sorting happens in the composition root.

### What looks good

- Catalog update correctly documents all WAF engine fields and matches the `config.WAFConfig` struct
- Tests are table-driven, descriptive, and cover all four combinations (both/engine-only/content-type-only/neither)
- Priority 25 is correct (matches `Plugin.Priority()`)
- Config field mapping between `config.WAFConfig` and `wafplugin.WAFEngineConfig` / `wafplugin.ContentTypeValidationConfig` is correct
- No new dependencies, no license concerns
- No security issues

tibtof (commented):
## Reviewer verdict: APPROVED (cannot formally approve own PR)

All four issues from the first review are resolved:

1. **Architecture violation fixed**: `internal/app/eject/eject.go` no longer imports `internal/plugins/waf`. The `Eject()` method now accepts `[]ports.CaddyHandler` as an injected parameter. Handler construction moved to `buildEjectHandlers()` in `internal/cli/cmd/eject.go`, which uses the plugin registry pattern (`NewRegistry` / `RegisterBuiltinPlugins` / `CaddyContributors()`) — consistent with how the serve path works.

2. **Swallowed errors fixed**: The problematic `buildWAFHandlers()` in the app layer no longer exists. The replacement `buildEjectHandlers()` in the CLI layer propagates errors from `registry.InitAll()`. The `ContributeCaddyHandlers()` interface's internal error logging is pre-existing behavior, not introduced by this PR.

3. **Unnecessary exports reverted**: `BuildContentTypeHandlerJSON` and `BuildEngineHandlerJSON` are back to unexported (`buildContentTypeHandlerJSON`, `buildEngineHandlerJSON`) with cleaned-up godoc.

4. **Unnecessary sort removed**: No sort in the diff.

Additionally verified: all CI checks pass (Build, Test, Integration Tests, Lint, Coverage, Semgrep, Trivy, Go Vulnerability Check), no new dependencies added, catalog descriptor update is complete, table-driven test for handler passthrough covers nil/empty/single/multiple cases.

Ready for human review.

PR #919, opened 2026-04-18
fix(#907): use HTTPS for deploy health check when TLS is enabled

### LINKED ISSUE #907: bug: deploy health check polls plain HTTP on port 443 — always times out on HTTPS-only stacks
## From v0.11.0 retrospective

\`vibew deploy\` polls \`http://localhost:443/_vibewarden/health\` — plaintext HTTP on a port that's HTTPS-only with HTTP→HTTPS redirect on port 80. Every deploy times out "waiting for health" even when everything is fine, then prints "Site deployed."

The deploy output trains users to ignore its own errors.

## Fix

Use \`https://<domain>/_vibewarden/health\` with TLS (accept self-signed for the check), or hit the admin endpoint over the Docker network.

tibtof:
Closes #907

## Summary
- When TLS is enabled (`cfg.TLS.Enabled`), deploy health checks now use `https://localhost:<port>` with `curl -sfk` (the `-k` flag skips TLS cert verification since we are hitting localhost, not the public domain)
- When TLS is disabled, behavior is unchanged: `http://localhost:<port>` with `curl -sf`
- Extracted `healthCheckURL()` and `healthCheckCmd()` helpers to eliminate the duplicated URL construction across `service.go` and `multiapp.go`
- Updated all three callers: `Deploy()`, `BootstrapSidecar()`, and `DeployMultiApp()`
- Fixed the existing test `TestService_Deploy_HealthCheckAlwaysUsesLocalhost` to expect the correct HTTPS URLs when TLS is enabled

## Test plan
- [x] `make check` passes (formatting, golangci-lint, build, all tests)
- [x] `TestService_Deploy_HealthCheckAlwaysUsesLocalhost` covers TLS disabled (HTTP), TLS enabled without domain (HTTPS -k), and TLS enabled with domain (HTTPS -k)
- [x] `TestService_Deploy_HealthCheckTimeout` and `TestService_Deploy_HealthCheckWarnOnTimeout` continue to pass (default config has TLS disabled, so they still match the `http://` curl command)
- [x] All multiapp tests pass unchanged (fakeExecutor returns success for unknown commands, so the new `https` curl is accepted)

### DIFF
--- a/internal/app/deploy/multiapp.go
+++ b/internal/app/deploy/multiapp.go
@@ -134,9 +134,9 @@ func (s *Service) BootstrapSidecar(ctx context.Context, cfg *config.Config, opts
 	if port == 0 {
 		port = defaultHealthPort
 	}
-	healthURL := fmt.Sprintf("http://localhost:%d/_vibewarden/health", port)
+	healthURL := healthCheckURL(port, cfg.TLS.Enabled)
 	fmt.Fprintf(out, "Waiting for sidecar health check at %s (via SSH)...\n", healthURL)
-	s.waitHealthy(ctx, port, out)
+	s.waitHealthy(ctx, port, cfg.TLS.Enabled, out)
 
 	fmt.Fprintln(out, "Bootstrap complete.")
 	return nil
@@ -177,9 +177,9 @@ func (s *Service) DeployMultiApp(ctx context.Context, cfg *config.Config, opts R
 	if port == 0 {
 		port = defaultHealthPort
 	}
-	healthURL := fmt.Sprintf("http://localhost:%d/_vibewarden/health", port)
+	healthURL := healthCheckURL(port, cfg.TLS.Enabled)
 	fmt.Fprintf(out, "Waiting for sidecar health check at %s (via SSH)...\n", healthURL)
-	s.waitHealthy(ctx, port, out)
+	s.waitHealthy(ctx, port, cfg.TLS.Enabled, out)
 
 	fmt.Fprintln(out, "Site deployed.")
 	return nil
--- a/internal/app/deploy/multiapp_test.go
+++ b/internal/app/deploy/multiapp_test.go
@@ -20,6 +20,16 @@ func multiappConfig() *config.Config {
 	}
 }
 
+// multiappTLSConfig returns a Config with TLS enabled for multi-app deploy tests.
+func multiappTLSConfig() *config.Config {
+	return &config.Config{
+		Server:   config.ServerConfig{Port: 443},
+		Upstream: config.UpstreamConfig{Port: 3000},
+		App:      config.AppConfig{Image: "myapp:latest"},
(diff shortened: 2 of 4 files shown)


internal/app/deploy/multiapp.go
```diff
@@ -134,9 +134,9 @@ func (s *Service) BootstrapSidecar(ctx context.Context, cfg *config.Config, opts
 	if port == 0 {
 		port = defaultHealthPort
 	}
-	healthURL := fmt.Sprintf("http://localhost:%d/_vibewarden/health", port)
+	healthURL := healthCheckURL(port, cfg.TLS.Enabled)
```
tibtof, line 137: **Missing test coverage**: The multi-app deploy paths (`BootstrapSidecar` and `DeployMultiApp`) now pass `cfg.TLS.Enabled` to `waitHealthy` and `healthCheckURL`, but there are no test cases in `multiapp_test.go` that exercise the TLS-enabled code path. The existing `multiappConfig()` fixture has TLS disabled, so only the `http://` + `curl -sf` path is tested for multi-app deploys.

Please add at least one table-driven test case (or a new test function) in `multiapp_test.go` that creates a config with `TLS: config.TLSConfig{Enabled: true}` and asserts that:
1. The health check URL logged to output contains `https://`
2. The curl command executed via `fakeExecutor` is `curl -sfk https://localhost:<port>/_vibewarden/health`

This mirrors what `TestService_Deploy_HealthCheckAlwaysUsesLocalhost` already does for the single-app `Deploy` path. Without this, a regression in the multi-app TLS health check path would go undetected.

tibtof (commented):
## Review: CHANGES REQUESTED

The fix correctly addresses issue #907 — the deploy health check now uses `https://` with `curl -sfk` when `cfg.TLS.Enabled` is true, and keeps `http://` with `curl -sf` when TLS is disabled. The implementation is clean:

- Two well-documented helper functions (`healthCheckURL` and `healthCheckCmd`) centralize the protocol/flag logic
- All three deploy paths (single-app `Deploy`, multi-app `BootstrapSidecar`, multi-app `DeployMultiApp`) are updated consistently
- The `waitHealthy` signature change is propagated correctly to all call sites
- Godoc comments are present on all new functions
- The non-TLS path is unchanged (no regression)
- No new dependencies, no architecture violations

### Issue requiring changes

**Missing multi-app TLS test coverage**: The `multiapp_test.go` file has no test cases exercising the TLS-enabled health check path. All existing multi-app tests use `multiappConfig()` which has TLS disabled. The single-app path has good TLS test coverage via `TestService_Deploy_HealthCheckAlwaysUsesLocalhost`, but the multi-app path does not. Please add equivalent coverage for at least one of `BootstrapSidecar` or `DeployMultiApp` with TLS enabled, asserting the correct `curl -sfk https://` command and `https://` URL in output.

PR #923, opened 2026-04-18
feat(#913): add local runtime and production checks to vibew doctor

### LINKED ISSUE #913: feat: vibew doctor should diagnose runtime state, not just config syntax
## From v0.11.0 retrospective

\`vibew doctor\` validates config but doesn't check runtime state. After a deploy, it should tell you:

- \"WAF is configured but not active in the route chain\"
- \"TLS cert issued but proxy.started log reports tls_enabled: false\"
- \"Health check endpoint unreachable on the configured port\"

Walking the ejected config and comparing to the declared config is mechanical. One line of output would save an hour of debugging.

tibtof:
Closes #913

## Summary

- Added two new check layers to `vibew doctor` beyond config/Docker:
  - **Local Runtime**: upstream reachability via HTTP health check, TLS certificate validity (expiry check for self-signed certs)
  - **Production** (requires `--target` flag): SSH connectivity, remote container health via `docker compose ps`, domain DNS resolution against target IP, remote TLS certificate expiry
- Added `--target` and `--ssh-key` flags to `vibew doctor` CLI command
- Added `WithRemoteExecutor()` method to `DoctorService` for optional SSH executor injection
- Output is grouped with section headers ("Config & Docker", "Local Runtime", "Production") in both human and JSON output
- Production checks only run when `--target` is provided; graceful handling when SSH fails

## Test plan

- [x] All 11 original doctor tests pass unchanged
- [x] 15 new tests added covering:
  - Upstream reachable / unreachable
  - TLS cert valid / expired / expiring soon / missing / non-self-signed skip
  - SSH connectivity success / failure
  - Remote container health (healthy / unhealthy)
  - Auto-detection: no target = no remote checks
  - Section headers in human and JSON output
- [x] `make check` passes (golangci-lint 0 issues, all tests with race detector)

### DIFF
--- a/internal/app/ops/doctor.go
+++ b/internal/app/ops/doctor.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -167,8 +168,8 @@ func (s *DoctorService) runChecks(ctx context.Context, cfg *config.Config, opts 
 		results = append(results, withSection(s.checkSSHConnectivity(ctx), sectionProduction))
 		results = append(results, withSection(s.checkRemoteContainerHealth(ctx), sectionProduction))
 		if cfg.TLS.Domain != "" {
-			results = append(results, withSection(checkDomainDNS(cfg.TLS.Domain, opts.Target), sectionProduction))
-			results = append(results, withSection(checkRemoteTLSCert(cfg.TLS.Domain), sectionProduction))
+			results = append(results, withSection(checkDomainDNS(ctx, cfg.TLS.Domain, opts.Target), sectionProduction))
+			results = append(results, withSection(checkRemoteTLSCert(ctx, cfg.TLS.Domain), sectionProduction))
 		}
 	}
 
@@ -517,9 +518,13 @@ func (s *DoctorService) checkRemoteContainerHealth(ctx context.Context) CheckRes
 }
 
 // checkDomainDNS resolves the configured TLS domain and verifies it points to
-// the target host IP. Uses net.LookupHost from stdlib.
-func checkDomainDNS(domain, target string) CheckResult {
-	addrs, err := net.LookupHost(domain)
+// the target host IP. Uses net.DefaultResolver.LookupHost for context-aware
+// DNS resolution.
+func checkDomainDNS(ctx context.Context, domain, target string) CheckResult {
+	dnsCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	addrs, err := net.DefaultResolver.LookupHost(dnsCtx, domain)
 	if err != nil {
 		return CheckResult{
 			Name:     "Domain DNS",
@@ -540,7 +545,7 @@ func checkDomainDNS(domain, target string) CheckResult {
 	targetIP := extractHostFromTarget(target)
 
 	// Resolve the target host in case it is a hostname rather than an IP.
-	targetAddrs, err := net.LookupHost(targetIP)
+	targetAddrs, err := net.DefaultResolver.LookupHost(dnsCtx, targetIP)
 	if err != nil {
 		targetAddrs = []string{targetIP}
 	}
@@ -568,38 +573,32 @@ func checkDomainDNS(domain, target string) CheckResult {
 }
 
 // extractHostFromTarget parses the host component from an ssh://user@host[:port]
-// URL. If parsing fails it returns the raw input.
+// URL. It uses net/url.Parse which correctly handles IPv6 addresses and edge
+// cases. If parsing fails it returns the raw input.
 func extractHostFromTarget(raw string) string {
-	// Try standard URL parsing: ssh://user@host:port
-	if strings.Contains(raw, "://") {
-		parts := strings.SplitN(raw, "://", 2)
-		if len(parts) == 2 {
-			after := parts[1]
-			// Remove user@ prefix
-			if idx := strings.Index(after, "@"); idx >= 0 {
-				after = after[idx+1:]
-			}
-			// Remove :port suffix
-			if host, _, err := net.SplitHostPort(after); err == nil {
-				return host
-			}
-			return after
-		}
+	u, err := url.Parse(raw)
+	if err != nil || u.Host == "" {
+		return raw
 	}
-	return raw
+	host := u.Hostname()
+	if host == "" {
+		return raw
+	}
+	return host
 }
 
 // checkRemoteTLSCert connects to domain:443 and checks the certificate expiry.
-func checkRemoteTLSCert(domain string) CheckResult {
-	conn, err := tls.DialWithDialer(
-		&net.Dialer{Timeout: 10 * time.Second},
-		"tcp",
-		domain+":443",
-		&tls.Config{
+func checkRemoteTLSCert(ctx context.Context, domain string) CheckResult {
+	tlsCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	dialer := tls.Dialer{
+		Config: &tls.Config{
 			// We accept any cert — the purpose is to inspect expiry, not validate trust.
 			InsecureSkipVerify: true, //nolint:gosec
 		},
-	)
+	}
+	conn, err := dialer.DialContext(tlsCtx, "tcp", domain+":443")
 	if err != nil {
 		return CheckResult{
 			Name:     "Remote TLS cert",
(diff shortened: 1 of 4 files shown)


internal/app/ops/doctor.go
```diff
@@ -275,17 +325,345 @@ func (s *DoctorService) checkContainerHealth(ctx context.Context, composePath st
...
+		Name:     "Remote containers",
+		Severity: SeverityOK,
+		Detail:   fmt.Sprintf("%d container(s) running", total),
+	}
+}
+
+// checkDomainDNS resolves the configured TLS domain and verifies it points to
+// the target host IP. Uses net.LookupHost from stdlib.
+func checkDomainDNS(domain, target string) CheckResult {
+	addrs, err := net.LookupHost(domain)
```
tibtof: **Missing context.Context**: `checkDomainDNS` performs DNS lookups via `net.LookupHost` (I/O) but does not accept a `context.Context` parameter. Per project rules, `context.Context` must be the first argument on all I/O functions. This also means there is no way for the caller to cancel the DNS lookup if it hangs.

**Suggestion**: Change the signature to `checkDomainDNS(ctx context.Context, domain, target string) CheckResult` and use `net.Resolver{}.LookupHost(ctx, ...)` instead of the package-level `net.LookupHost`.

internal/app/ops/doctor.go
```diff
@@ -275,17 +325,345 @@ func (s *DoctorService) checkContainerHealth(ctx context.Context, composePath st
...
+				return host
+			}
+			return after
+		}
+	}
+	return raw
+}
+
+// checkRemoteTLSCert connects to domain:443 and checks the certificate expiry.
+func checkRemoteTLSCert(domain string) CheckResult {
```
tibtof: **Missing context.Context**: `checkRemoteTLSCert` performs a TLS dial (I/O) but does not accept a `context.Context` parameter. Per project rules, `context.Context` must be the first argument on all I/O functions. The 10-second timeout is hardcoded and cannot be cancelled by the parent context.

**Suggestion**: Change the signature to `checkRemoteTLSCert(ctx context.Context, domain string) CheckResult` and use a `tls.Dialer` with `DialContext` (Go 1.15+) or use `net.Dialer{Timeout: ..., Deadline: ctx.Deadline()}` to respect the context.

internal/app/ops/doctor.go
```diff
@@ -275,17 +325,345 @@ func (s *DoctorService) checkContainerHealth(ctx context.Context, composePath st
...
+	return CheckResult{
+		Name:     "Remote containers",
+		Severity: SeverityOK,
+		Detail:   fmt.Sprintf("%d container(s) running", total),
+	}
+}
+
+// checkDomainDNS resolves the configured TLS domain and verifies it points to
+// the target host IP. Uses net.LookupHost from stdlib.
+func checkDomainDNS(domain, target string) CheckResult {
```
tibtof: **Missing unit tests**: `checkDomainDNS` and `checkRemoteTLSCert` are the only new check functions without any unit tests. These functions make real network calls (DNS lookup, TLS dial to port 443), which makes them untestable in the current form.

Because they bypass the port/adapter pattern and call `net.LookupHost` and `tls.DialWithDialer` directly, consider one of:
1. Extract the DNS resolution and TLS dialing behind small interfaces (or function parameters) that can be faked in tests.
2. At minimum, add a unit test for `extractHostFromTarget` (which is pure logic and easy to test with a table-driven test covering `ssh://user@host`, `ssh://user@host:22`, bare host, etc.).

internal/app/ops/doctor.go
```diff
@@ -275,17 +325,345 @@ func (s *DoctorService) checkContainerHealth(ctx context.Context, composePath st
...
+	return CheckResult{
+		Name:     "Domain DNS",
+		Severity: SeverityWarn,
+		Detail:   fmt.Sprintf("%s resolves to %s but target is %s", domain, strings.Join(addrs, ", "), targetIP),
+	}
+}
+
+// extractHostFromTarget parses the host component from an ssh://user@host[:port]
+// URL. If parsing fails it returns the raw input.
+func extractHostFromTarget(raw string) string {
```
tibtof, line 578: **Prefer `net/url.Parse` over manual string splitting**: `extractHostFromTarget` manually splits on `://`, then on `@`, then uses `net.SplitHostPort`. The stdlib `net/url.Parse` handles all of these cases correctly and is more robust against edge cases (e.g., IPv6 addresses in brackets).

**Suggestion**:
```go
func extractHostFromTarget(raw string) string {
    u, err := url.Parse(raw)
    if err != nil || u.Host == "" {
        return raw
    }
    host := u.Hostname() // strips port and brackets for IPv6
    if host == "" {
        return raw
    }
    return host
}
```

This also needs a unit test -- at minimum a table-driven test covering: `ssh://user@host`, `ssh://user@host:22`, `ssh://host`, bare hostname, and IPv6 (`ssh://user@[::1]:22`).

internal/app/ops/doctor_test.go
```diff
@@ -27,6 +36,47 @@ func (f *fakePortChecker) IsPortAvailable(_ context.Context, _ string, port int)
...
+func (f *fakeRemoteExecutor) Transfer(_ context.Context, _, _ string, _ bool) error {
+	return nil
+}
+
+func (f *fakeRemoteExecutor) TransferFile(_ context.Context, _, _ string) error {
+	return nil
+}
+
+func (f *fakeRemoteExecutor) DryRunTransfer(_ context.Context, _, _ string) ([]string, error) {
+	return nil, nil
```
tibtof: **Nil slice return**: Per project convention, slices should never be returned as nil when empty -- use `[]string{}` instead. This is a test fake, so the impact is low, but it sets a bad example.

**Suggestion**: `return []string{}, nil`

tibtof (commented):
## Review Summary -- CHANGES REQUESTED

Overall this is a well-structured PR that adds valuable diagnostic checks to `vibew doctor`. The three-layer approach (Config & Docker, Local Runtime, Production) is clean, the existing tests all pass (30/30), and the new tests are thorough for the checks that go through ports.

### Issues found (4 actionable, 1 minor)

**1. Missing `context.Context` on I/O functions (code quality violation)**
Both `checkDomainDNS` and `checkRemoteTLSCert` perform network I/O (DNS lookups, TLS dials) but do not accept `context.Context` as their first parameter. Per project rules and Go idioms, all I/O functions must accept context for cancellation. The hardcoded timeouts cannot be overridden by the parent context.

**2. Missing unit tests for `checkDomainDNS` and `checkRemoteTLSCert`**
These are the only new check functions without any test coverage. Because they make real network calls directly (bypassing ports), they cannot be tested with fakes. Either inject the DNS resolver and TLS dialer as dependencies, or at minimum extract and test the pure logic (`extractHostFromTarget`).

**3. `extractHostFromTarget` has no unit test and uses manual URL parsing**
This function should use `net/url.Parse` from stdlib (handles IPv6, edge cases). It also needs a table-driven test covering `ssh://user@host`, `ssh://user@host:22`, bare hostname, and IPv6 forms.

**4. Minor: `fakeRemoteExecutor.DryRunTransfer` returns nil slice**
Per project convention: `return []string{}, nil` instead of `return nil, nil`.

### What looks good

- Correct reuse of existing `HealthChecker` and `RemoteExecutor` ports -- no unnecessary new interfaces
- `WithRemoteExecutor` builder pattern is clean and avoids changing the constructor signature
- Graceful error handling throughout -- SSH failures produce FAIL, not panics
- Section headers in both human and JSON output work correctly
- Production checks properly gated behind `opts.Target != "" && s.remoteExecutor != nil`
- All 11 existing tests continue to pass
- 19 new tests cover local runtime and production check scenarios well
- No new dependencies added
- CLI flags (`--target`, `--ssh-key`) properly integrated

tibtof (commented):
LGTM (self-authored PR -- cannot formally approve via API). Second review pass -- all 4 previously requested changes have been correctly addressed:

1. **context.Context added to I/O functions**: `checkDomainDNS` and `checkRemoteTLSCert` now accept `ctx context.Context` as their first parameter. DNS resolution uses `net.DefaultResolver.LookupHost(ctx, ...)` and TLS inspection uses `tls.Dialer{}.DialContext(ctx, ...)`, both properly respecting cancellation via derived timeout contexts.

2. **Tests for extractHostFromTarget**: New `doctor_internal_test.go` with a table-driven `TestExtractHostFromTarget` covering 9 cases: SSH URLs with/without user, with/without port, hostnames, IPv6 with brackets and port, bare hostname, and empty string.

3. **url.Parse replaces manual string splitting**: `extractHostFromTarget` now uses `url.Parse` + `u.Hostname()` from `net/url`, correctly handling IPv6 addresses and edge cases.

4. **fakeRemoteExecutor.DryRunTransfer returns empty slice**: Returns `[]string{}, nil` instead of `nil, nil`, matching project conventions.

All tests pass (30+ tests in `internal/app/ops`, including 20 new ones) with race detector enabled. No remaining issues found.

PR #934, opened 2026-04-18
fix(#925): wire plugin handlers into multi-site Caddy config

### LINKED ISSUE #925: CRITICAL: WAF and rate-limiting silently not enforcing in multi-site runtime
## From v0.12.0 agent retrospective

In multi-site (deployed) mode, WAF and rate-limiting are declared in both \`sites/<name>/vibewarden.yaml\` and \`global.yaml\`, no warning/error is emitted, but:
- SQLi payloads in query strings pass through (should be 403 in block mode)
- 30 rapid-fire requests all return 200 (rate limiter not enforcing)

Single-site \`vibew dev\` correctly blocks SQLi with 403. The same config deployed in multi-site mode silently drops the security layer.

**This is different from #906** (eject output). #906 fixed the eject/display path. This bug is about the **runtime Caddy config** in multi-site mode not including plugin-contributed handlers.

## Root cause hypothesis

\`cmd/vibewarden/wiring_serve_multisite.go\` builds the multi-site Caddy config via \`BuildMultiSiteConfig()\` in \`internal/adapters/caddy/multisite_config.go\`. This likely does not call \`CaddyContributors()\` to collect plugin handlers (WAF, rate-limit) for each site's route — the same class of omission that affected eject.

## Acceptance

- SQLi payload against a multi-site deployed endpoint returns 403 when \`waf.mode: block\`
- Rate limiter enforces per-IP limits in multi-site mode
- \`vibew eject\` in multi-site mode shows WAF + rate-limit handlers in each site's route

tibtof:
Closes #925

## Summary

- Multi-site mode was not running the plugin lifecycle, so plugins like WAF, auth, CORS, IP filter, input validation, maintenance mode, body size, and webhook signature verification never contributed their Caddy handlers to the per-site middleware chain. Only config-driven handlers (security headers, rate limiting, compression) were active.
- Added `buildPerSitePluginHandlers` in `cmd/vibewarden/wiring_serve_multisite.go` that creates a per-site `plugins.Registry` for each healthy site, calls `RegisterBuiltinPlugins` and `InitAll`, then collects `CaddyContributor` handlers into a `map[string][]ports.CaddyHandler`.
- Updated `NewMultiSiteAdapter` and `BuildMultiSiteConfig` to accept and propagate per-site handlers to the Caddy config builder.
- Updated `buildSiteRoutes` to insert plugin-contributed handlers after security headers but before rate limiting, sorted by ascending priority, matching the single-site handler chain order.

## Test plan

- [x] Existing tests updated to pass the new `perSiteHandlers` parameter (nil for backward compat)
- [x] `TestBuildMultiSiteConfig_PerSitePluginHandlers` — verifies plugin handlers are inserted only for the correct site and sorted by priority
- [x] `TestBuildMultiSiteConfig_PerSitePluginHandlers_NilMap` — verifies nil map produces no plugin handlers (backward compat)
- [x] `TestBuildSiteRoutes_WithExtraHandlers` — verifies extra handlers are inserted and sorted by priority
- [x] `TestBuildSiteRoutes_ExtraHandlersBeforeRateLimit` — verifies plugin handlers appear before config-driven rate limiting
- [x] `make check` passes (gofmt, golangci-lint, build, all tests)

### DIFF
--- a/cmd/vibewarden/wiring_serve_multisite.go
+++ b/cmd/vibewarden/wiring_serve_multisite.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
 	"syscall"
 
 	caddyadapter "github.com/vibewarden/vibewarden/internal/adapters/caddy"
@@ -16,6 +17,7 @@ import (
 	reloadsvc "github.com/vibewarden/vibewarden/internal/app/reload"
 	"github.com/vibewarden/vibewarden/internal/config/sites"
 	"github.com/vibewarden/vibewarden/internal/domain/site"
+	"github.com/vibewarden/vibewarden/internal/plugins"
 	"github.com/vibewarden/vibewarden/internal/ports"
 )
 
@@ -95,15 +97,27 @@ func runServeMultiSite(ctx context.Context, baseDir string, ver string) error {
 	// Step 5: Create the event logger.
 	eventLogger := logadapter.NewSlogEventLogger(os.Stdout)
 
-	// Step 6: Create the multi-site Caddy adapter.
+	// Step 6: Build per-site plugin registries and collect Caddy handlers.
+	//
+	// Each healthy site gets its own plugin registry so that plugins like WAF,
+	// rate limiting, auth, CORS, and IP filter contribute their handlers to the
+	// correct site's middleware chain. Only Init is called (no Start) because
+	// handler contribution requires initialised state but not running background
+	// work. See wiring_serve_helpers.go for the single-site equivalent.
+	perSiteHandlers, err := buildPerSitePluginHandlers(ctx, registry, eventLogger, logger)
+	if err != nil {
+		return fmt.Errorf("building per-site plugin handlers: %w", err)
+	}
+
+	// Step 7: Create the multi-site Caddy adapter.
 	// A minimal ProxyConfig is needed for backward-compatible fields (version).
 	minimalProxyCfg := &ports.ProxyConfig{
 		ListenAddr: fmt.Sprintf("%s:%d", globalCfg.ListenHost, globalCfg.ListenPort),
 		Version:    ver,
 	}
-	adapter := caddyadapter.NewMultiSiteAdapter(minimalProxyCfg, registry, logger, eventLogger)
+	adapter := caddyadapter.NewMultiSiteAdapter(minimalProxyCfg, registry, perSiteHandlers, logger, eventLogger)
 
-	// Step 7: Set up signal handling.
+	// Step 8: Set up signal handling.
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
@@ -120,7 +134,7 @@ func runServeMultiSite(ctx context.Context, baseDir string, ver string) error {
 		}
 	}()
 
-	// Step 8: Start the site watcher for hot-reload.
+	// Step 9: Start the site watcher for hot-reload.
 	siteWatcher := fsnotifyadapter.NewSiteWatcher(logger)
 	watchCh, watchErr := siteWatcher.Watch(ctx, sitesDir)
 	if watchErr != nil {
@@ -129,7 +143,7 @@ func runServeMultiSite(ctx context.Context, baseDir string, ver string) error {
 			slog.String("error", watchErr.Error()),
 		)
 	} else {
-		// Step 9: Create the MultiSiteService event loop.
+		// Step 10: Create the MultiSiteService event loop.
 		reloadFn := func(reloadCtx context.Context) error {
 			return adapter.Reload(reloadCtx)
 		}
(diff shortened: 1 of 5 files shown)


cmd/vibewarden/wiring_serve_multisite.go
```diff
@@ -95,15 +97,27 @@ func runServeMultiSite(ctx context.Context, baseDir string, ver string) error {
...
+	}
+
+	// Step 7: Create the multi-site Caddy adapter.
 	// A minimal ProxyConfig is needed for backward-compatible fields (version).
 	minimalProxyCfg := &ports.ProxyConfig{
 		ListenAddr: fmt.Sprintf("%s:%d", globalCfg.ListenHost, globalCfg.ListenPort),
 		Version:    ver,
 	}
-	adapter := caddyadapter.NewMultiSiteAdapter(minimalProxyCfg, registry, logger, eventLogger)
+	adapter := caddyadapter.NewMultiSiteAdapter(minimalProxyCfg, registry, perSiteHandlers, logger, eventLogger)
```
tibtof, line 118: **Bug: hot-reload serves stale plugin handlers.** `perSiteHandlers` is computed once in `buildPerSitePluginHandlers()` at startup (step 6) and stored in the `Adapter` struct. When a site's config changes at runtime, `MultiSiteService.handleModified` updates the site registry and calls `adapter.Reload()`, which calls `buildConfigJSON()` -> `BuildMultiSiteConfig()` with the _original_ `a.perSiteHandlers`. The plugin handlers are never recomputed, so config changes to WAF mode, CORS origins, rate-limit rules, etc. are silently ignored until the sidecar is restarted.

**Why this matters:** This is the exact class of bug this PR is fixing -- handlers not taking effect. If an operator changes WAF mode from `detect` to `block` via hot-reload, the stale handlers will keep running in detect mode.

**Suggested fix:** Either:
1. Re-run `buildPerSitePluginHandlers()` inside the reload path (store the builder deps in the adapter), or
2. Make `Adapter` re-derive `perSiteHandlers` in `buildConfigJSON()` using a callback/factory, or
3. At minimum, document this as a known limitation with a TODO and a code comment noting that plugin handler changes require a sidecar restart.

Option 3 is acceptable for this PR if the scope is intentionally limited to the initial startup path, but please add an explicit comment and a follow-up issue.

cmd/vibewarden/wiring_serve_multisite.go
```diff
@@ -146,6 +160,76 @@ func runServeMultiSite(ctx context.Context, baseDir string, ver string) error {
...
+		}
+
+		siteLogger := logger.With(slog.String("site", s.Name()))
+
+		// Create a per-site plugin registry with a site-scoped logger so that
+		// plugin lifecycle log lines include the site name.
+		reg := plugins.NewRegistry(siteLogger)
+		plugins.RegisterBuiltinPlugins(reg, cfg, eventLogger, siteLogger)
+
+		if err := reg.InitAll(ctx); err != nil {
```
tibtof, line 210: **Resource leak: per-site plugin registries are initialised but never stopped.** Each iteration creates a `plugins.NewRegistry`, calls `RegisterBuiltinPlugins` (which registers ~15 plugins including metrics with OTel SDK, rate-limiting with memory stores, etc.), and then calls `reg.InitAll(ctx)`. After handler collection, the registry goes out of scope without `reg.StopAll()` ever being called.

While most plugins are lightweight during Init (allocating config, validating), the metrics plugin creates an OTel MeterProvider (line visible in test output: "metrics plugin initialised with OTel SDK"), and future plugins may allocate goroutines or connections during Init.

**Suggested fix:** Call `defer reg.StopAll(ctx)` after `InitAll` succeeds, or immediately after collecting handlers if you want deterministic cleanup within the loop iteration. Since you only need the handler JSON fragments (plain `map[string]any` values), the plugin registries can be safely torn down after collection.

cmd/vibewarden/wiring_serve_multisite.go
```diff
@@ -146,6 +160,76 @@ func runServeMultiSite(ctx context.Context, baseDir string, ver string) error {
...
+
+		// Collect handlers from CaddyContributor plugins.
+		var handlers []ports.CaddyHandler
+		for _, contrib := range reg.CaddyContributors() {
+			handlers = append(handlers, contrib.ContributeCaddyHandlers()...)
+		}
+
+		if len(handlers) > 0 {
+			// Sort by ascending priority for deterministic ordering.
+			sort.Slice(handlers, func(i, j int) bool {
```
tibtof: **Nit: double-sorting of handlers.** The handlers are sorted here by priority, and then sorted again inside `buildSiteRoutes()` (which makes a defensive copy and re-sorts). The sorting in `buildSiteRoutes` is the authoritative one since it operates on a copy, making this sort redundant.

This is not a bug -- it is harmless defensive programming -- but it creates a false impression that the order in this map matters. Consider removing this sort and letting `buildSiteRoutes` be the single point of ordering, or removing the sort in `buildSiteRoutes` and documenting that callers must pre-sort. Having both is confusing.

tibtof (commented):
## Review Summary -- PR #934 (fix #925: multi-site plugin handler wiring)

**Verdict: CHANGES REQUESTED** (submitting as comment because GH does not allow request-changes on own PR)

### What was reviewed

- `cmd/vibewarden/wiring_serve_multisite.go` -- new `buildPerSitePluginHandlers()` function and its integration into the startup sequence
- `internal/adapters/caddy/adapter.go` -- `NewMultiSiteAdapter` signature change and `perSiteHandlers` field
- `internal/adapters/caddy/multisite_config.go` -- `BuildMultiSiteConfig` and `buildSiteRoutes` updated to accept and insert extra handlers
- `internal/adapters/caddy/multisite_config_test.go` -- 4 new tests covering plugin handlers, nil map, ordering, and rate-limit interaction
- `internal/adapters/caddy/adapter_test.go` -- existing tests updated for new signature

### What works well

- **Plugin lifecycle is correct**: Only `Init` is called (not `Start`), matching the single-site CaddyContributor lifecycle documented in `wiring_serve_helpers.go`.
- **Handler ordering is correct**: Extra handlers are inserted after security headers but before rate limiting, matching the single-site middleware chain order.
- **Priority sorting is correct**: Lower priority numbers run first, matching the `ports.CaddyHandler` contract.
- **Test coverage is thorough**: Tests cover per-site handler injection, nil map (backward compat), priority ordering, handler-before-rate-limit ordering, and site isolation (site B does not get site A's handlers).
- **No architecture violations**: All imports respect the hexagonal boundaries. Domain layer is untouched. Interfaces remain in `ports/`. No new dependencies added.
- **No single-site regression**: All existing tests pass with the updated signatures (nil perSiteHandlers).
- **Error handling is solid**: Critical plugin failures are logged and the site continues without plugin handlers.

### Issues requiring changes

1. **Bug: hot-reload serves stale plugin handlers** (blocking). See inline comment at line 119 of `wiring_serve_multisite.go`. The `perSiteHandlers` map is computed once at startup and baked into the adapter. When `MultiSiteService.handleModified` triggers a reload, the adapter reuses the stale handlers. This must be either fixed or explicitly documented as a known limitation with a follow-up issue.

2. **Resource leak: per-site plugin registries never stopped** (blocking). See inline comment at line 202 of `wiring_serve_multisite.go`. `InitAll` is called but `StopAll` is never called. The metrics plugin allocates an OTel MeterProvider during Init which is never cleaned up.

3. **Nit: double-sorting** (non-blocking). See inline comment at line 219. Handlers are sorted twice -- once in `buildPerSitePluginHandlers` and again in `buildSiteRoutes`. Pick one location.

### Checklist

- [x] Architecture: no layer violations
- [x] Godoc comments on all new exported symbols
- [x] Errors wrapped with context
- [x] No swallowed errors
- [x] No panic outside main
- [x] context.Context as first argument on I/O functions
- [x] Tests for new code (4 new test functions)
- [x] Table-driven tests where appropriate
- [x] No new dependencies added
- [x] No secrets or credentials
- [ ] Resource cleanup (StopAll not called)
- [ ] Hot-reload correctness (stale handlers)

tibtof:
## Review Summary -- PR #934 (second pass)

**Verdict: APPROVED** (submitting as comment because GH does not allow approval on own PR)

All three issues from the first review are resolved:

1. **Stale handlers on hot-reload**: Fixed. The `reloadFn` closure now calls `buildPerSitePluginHandlers()` to compute fresh handlers and `adapter.UpdatePerSiteHandlers()` before calling `Reload()`. Two new tests (`TestAdapter_UpdatePerSiteHandlers`, `TestAdapter_UpdatePerSiteHandlers_OverwritesExisting`) verify the method works correctly.

2. **Resource leak on per-site plugin registries**: Fixed. `reg.StopAll(ctx)` is called after collecting handlers from `reg.CaddyContributors()`, with the error logged at warn level.

3. **Double-sorting**: Fixed. Sorting removed from `buildPerSitePluginHandlers` (with explicit comment explaining deferral). Only `buildSiteRoutes` sorts, using a defensive copy to avoid mutating input.

### Checklist verified

- [x] Architecture: no layer violations -- `plugins` import in `cmd/` composition root is correct
- [x] `perSiteHandlers` field in adapter follows existing pattern (no mutex, same as `UpdateConfig`)
- [x] Godoc comments on all new exported symbols (`UpdatePerSiteHandlers`, `buildPerSitePluginHandlers`)
- [x] Errors wrapped with context (`fmt.Errorf("building per-site plugin handlers: %w", err)`)
- [x] No swallowed errors -- `StopAll` error logged at warn level
- [x] Tests: 6 new test functions covering handler injection, nil map, priority ordering, overwrite, and rate-limit interaction
- [x] No new dependencies added
- [x] No secrets or credentials

PR #949, opened 2026-04-19
feat(#937): transfer local Docker image to remote via docker save/load

### LINKED ISSUE #937: feat: vibew deploy should support local image transfer (docker save/load)
## Problem

When using \`app.build: .\` locally (\`vibew build\`), the resulting Docker image lives only in the local daemon. \`vibew deploy\` then either:
- Builds from source on the remote (slow, requires rsyncing source)
- Or pulls from a registry (\`app.image\`) which requires pushing first

There's no path for "I built locally, send that image to the server" without a registry.

## Proposed behavior

Auto-detect the deploy strategy based on config:

1. **\`app.image\` with registry prefix** (e.g., \`ghcr.io/...\`, \`docker.io/...\`) → pull from registry on remote (current behavior)
2. **\`app.image\` without registry prefix** (e.g., \`myapp:latest\`) → transfer local image via \`docker save | ssh docker load\`
3. **\`app.build: .\`** → rsync source + build on remote (current behavior)

Option 2 is the new path. The flow:
\`\`\`
vibew build                    # builds myapp:latest locally
vibew deploy --target ssh://...  # detects local-only image, transfers it
\`\`\`

Under the hood:
\`\`\`bash
docker save myapp:latest | ssh user@host docker load
\`\`\`

## Why opt-in by detection

- If the image name has a registry prefix → assume registry workflow (don't waste bandwidth transferring)
- If the image name is bare (no \`/\` before the tag) → assume local build, transfer it
- \`app.build\` → always build on remote (source transfer)

This covers the common vibe coder scenario: no registry, just \`vibew build\` + \`vibew deploy\`.

## Acceptance

- \`vibew deploy\` with \`app.image: myapp:latest\` (no registry prefix) transfers the local image via \`docker save | ssh docker load\`
- \`vibew deploy\` with \`app.image: ghcr.io/org/myapp:latest\` pulls from registry (unchanged)
- \`vibew deploy\` with \`app.build: .\` builds on remote (unchanged)
- Progress output shows transfer size/progress

tibtof:
Closes #937

## Summary
- Auto-detect local vs registry images based on image name: bare names (no `/` before `:`) are treated as local images
- When a local image is detected, transfer it to the remote via `docker save` -> rsync -> `docker load`
- Add `ImageExporter` port interface in `internal/ports/ops.go` and adapter in `internal/adapters/ops/image_export.go`
- Wire the exporter into the deploy `Service` via `WithImageExporter()` method (backward compatible -- when nil, falls back to normal `docker compose pull`)
- Works in both single-app (`Deploy()`) and multi-app (`deploySite()`) flows

## Test plan
- [x] `TestIsLocalImage` -- table-driven test covering bare names, registry prefixes, empty strings, edge cases
- [x] `TestService_Deploy_LocalImage` -- verifies docker save, rsync, docker load are called in order
- [x] `TestService_Deploy_LocalImage_SaveFails` -- error from docker save is propagated
- [x] `TestService_Deploy_LocalImage_TransferFails` -- rsync failure for image archive is propagated
- [x] `TestService_Deploy_LocalImage_LoadFails` -- docker load failure on remote is propagated
- [x] `TestService_Deploy_RegistryImage_NoTransfer` -- registry images skip local transfer, use normal pull
- [x] `TestService_Deploy_LocalImage_NoExporter` -- nil exporter falls back to pull (backward compat)
- [x] `TestDeployMultiApp_LocalImage` -- multi-app flow transfers local image
- [x] `TestBootstrapSidecar_LocalImage` -- bootstrap flow transfers local image
- [x] `TestDeployMultiApp_RegistryImage_NoLocalTransfer` -- registry images skip transfer in multi-app
- [x] `TestDeployMultiApp_LocalImage_SaveFails` -- multi-app save failure propagated
- [x] `make check` passes (golangci-lint 0 issues, all tests pass with -race)

### DIFF
--- a/internal/adapters/ops/image_export.go
+++ b/internal/adapters/ops/image_export.go
@@ -1,0 +1,28 @@
+package ops
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+)
+
+// ImageExportAdapter implements ports.ImageExporter by shelling out to the
+// docker CLI.
+type ImageExportAdapter struct{}
+
+// NewImageExportAdapter creates a new ImageExportAdapter.
+func NewImageExportAdapter() *ImageExportAdapter {
+	return &ImageExportAdapter{}
+}
+
+// Save exports the named Docker image to destPath using "docker save -o".
+func (a *ImageExportAdapter) Save(ctx context.Context, imageName, destPath string) error {
+	//nolint:gosec // "docker" is a hardcoded binary; args are operator-controlled
+	cmd := exec.CommandContext(ctx, "docker", "save", "-o", destPath, imageName)
+
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("docker save %s: %w\noutput: %s", imageName, err, string(output))
+	}
+	return nil
+}
--- a/internal/app/deploy/image_transfer.go
+++ b/internal/app/deploy/image_transfer.go
@@ -1,0 +1,81 @@
+package deploy
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+)
+
+// isLocalImage returns true when imageName is a bare Docker image name without
+// a registry prefix. A registry prefix is identified by the presence of a "/"
+// before the optional ":" tag separator.
+//
+// Examples:
+//
+//	"myapp:latest"                  -> true  (local)
+//	"myapp"                         -> true  (local, no tag)
+//	"ghcr.io/org/myapp:latest"     -> false (registry)
+//	"docker.io/library/nginx:latest" -> false (registry)
+//	"localhost:5000/myapp:latest"   -> false (registry — contains "/")
+func isLocalImage(imageName string) bool {
+	if imageName == "" {
+		return false
+	}
+
+	// Strip the tag (":latest", ":v1.0", etc.) to isolate the repository part.
+	repo := imageName
+	if idx := strings.LastIndex(imageName, ":"); idx != -1 {
+		repo = imageName[:idx]
+	}
+
+	// A registry prefix always introduces at least one "/" in the repo part
+	// (e.g. "ghcr.io/org/app", "docker.io/library/nginx", "localhost:5000/app").
+	// A bare name like "myapp" has no "/".
+	return !strings.Contains(repo, "/")
+}
+
+// transferLocalImage saves the Docker image from the local daemon, rsyncs the
+// tar archive to the remote host, loads it into the remote Docker daemon, and
+// cleans up the temporary files on both sides.
+func (s *Service) transferLocalImage(ctx context.Context, imageName, remoteDir string, out io.Writer) error {
+	if s.imageExporter == nil {
+		return fmt.Errorf("local image %q requires an image exporter; this is a bug — please report it", imageName)
+	}
+
+	fmt.Fprintf(out, "Transferring local image %s to remote...\n", imageName)
+
+	// Step 1: save the image to a local temp file.
+	tmpFile, err := os.CreateTemp("", "vibewarden-image-*.tar")
+	if err != nil {
+		return fmt.Errorf("creating temp file for image export: %w", err)
+	}
+	tmpPath := tmpFile.Name()
+	tmpFile.Close() //nolint:errcheck // we only need the path; docker save writes to it
+
+	// Ensure local cleanup.
+	defer os.Remove(tmpPath) //nolint:errcheck
+
+	fmt.Fprintf(out, "  Saving image %s to local temp file...\n", imageName)
+	if err := s.imageExporter.Save(ctx, imageName, tmpPath); err != nil {
+		return fmt.Errorf("saving image %s locally: %w", imageName, err)
+	}
+
+	// Step 2: rsync the tar to the remote host.
+	remoteTmpPath := remoteDir + "image-transfer.tar"
+	fmt.Fprintln(out, "  Uploading image archive to remote...")
+	if err := s.executor.TransferFile(ctx, tmpPath, remoteTmpPath); err != nil {
+		return fmt.Errorf("transferring image archive to remote: %w", err)
+	}
+
+	// Step 3: load the image on the remote and clean up the remote file.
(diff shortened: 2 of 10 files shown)


internal/adapters/ops/image_export.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"os/exec"
+)
+
+// ImageExportAdapter implements ports.ImageExporter by shelling out to the
+// docker CLI.
+type ImageExportAdapter struct{}
+
+// NewImageExportAdapter creates a new ImageExportAdapter.
+func NewImageExportAdapter() *ImageExportAdapter {
```
tibtof, line 14: **Missing test file**: `internal/adapters/ops/image_export.go` has no corresponding `image_export_test.go`. The project rules require every new `.go` file to have a companion test file.

While the adapter shells out to `docker save` (making full integration testing impractical in CI), you should still add a unit test that:
1. Verifies `NewImageExportAdapter()` returns a non-nil value.
2. Verifies `Save()` returns an error when given a non-existent image name (the `docker` binary will fail, which is testable if docker is available, or you can at least verify the error wrapping by testing with an invalid binary path via `exec.LookPath`).

At minimum, add a constructor test similar to the pattern in `health_test.go` and `port_test.go` in the same package.

internal/app/deploy/service.go
```diff
@@ -80,6 +81,15 @@ func NewService(
 	}
 }
 
+// WithImageExporter returns a copy of the Service with the given ImageExporter
+// set. The exporter is used to save Docker images from the local daemon so they
+// can be transferred to the remote host when the image name has no registry
```
tibtof: **Misleading godoc**: The comment says "returns a copy of the Service" but the implementation mutates the receiver in place (`s.imageExporter = exporter`) and returns the same pointer. This is not a copy -- it modifies the original `*Service`.

Either:
- Fix the doc to say "sets the ImageExporter on the Service and returns it for chaining", or
- Actually return a copy: `copy := *s; copy.imageExporter = exporter; return &copy`

The current pattern (mutate + return for chaining) is fine and idiomatic, so updating the doc is the simpler fix.

internal/app/deploy/image_transfer.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Step 2: rsync the tar to the remote host.
+	remoteTmpPath := remoteDir + "image-transfer.tar"
+	fmt.Fprintln(out, "  Uploading image archive to remote...")
+	if err := s.executor.TransferFile(ctx, tmpPath, remoteTmpPath); err != nil {
+		return fmt.Errorf("transferring image archive to remote: %w", err)
+	}
+
+	// Step 3: load the image on the remote and clean up the remote file.
+	fmt.Fprintln(out, "  Loading image on remote...")
```
tibtof, line 75: **Remote temp file leaked on docker load failure**: The `&&` operator means `rm -f` only runs when `docker load` succeeds. If `docker load` fails (corrupt archive, disk full, etc.), `image-transfer.tar` -- potentially hundreds of MB -- is orphaned on the remote host.

Use `;` or a separate cleanup step so the remote file is always removed:

```go
loadCmd := fmt.Sprintf("docker load < %s; loadExit=\$?; rm -f %s; exit \$loadExit", remoteTmpPath, remoteTmpPath)
```

Or split into two executor calls:
```go
_, loadErr := s.executor.Run(ctx, fmt.Sprintf("docker load < %s", remoteTmpPath))
// Always clean up the remote file, even on load failure.
_, _ = s.executor.Run(ctx, fmt.Sprintf("rm -f %s", remoteTmpPath))
if loadErr != nil {
    return fmt.Errorf("loading image on remote: %w", loadErr)
}
```

The second approach is cleaner because it separates concerns and makes the cleanup explicit.

tibtof (commented):
**Reviewer: CHANGES_REQUESTED**

Three issues found:

1. **Missing test file** (`internal/adapters/ops/image_export.go`): No `image_export_test.go` exists. Project rules require every new `.go` file to have a corresponding test file.

2. **Misleading godoc on `WithImageExporter`**: Doc says "returns a copy" but the implementation mutates the receiver in place. Fix the doc to match the (correct) behavior.

3. **Remote temp file leak on docker load failure** (`image_transfer.go:73`): The `&&` in `docker load < file && rm -f file` means `rm -f` is skipped when `docker load` fails, orphaning potentially large tar archives on the remote host. Split into two executor calls or use a shell pattern that always cleans up.

Everything else looks good:
- `isLocalImage()` detection logic is correct and well-tested with 11 table-driven cases.
- Transfer flow (docker save -> rsync -> docker load -> cleanup) is well-structured.
- `ImageExporter` port interface is clean and minimal.
- Wired correctly into both `Deploy()` (single-site) and `deploySite()` (multi-app).
- Registry-prefixed images and `app.build` paths are unaffected.
- Local temp file cleanup via `defer os.Remove` is correct.
- Backward compatibility preserved when `imageExporter` is nil (falls back to pull).
- Test coverage is thorough: happy path, save failure, transfer failure, load failure, registry-no-transfer, no-exporter fallback, and multi-app variants.
- No new dependencies added.
- Architecture rules followed (interface in ports, adapter in adapters, logic in app layer).

tibtof (commented):
**Reviewer: APPROVED**

All three issues from the previous review are resolved:

1. **Adapter test added**: `internal/adapters/ops/image_export_test.go` now exists with two tests -- cancelled-context error propagation and missing-docker error. Both properly use the existing `dockerAvailable()` helper and skip when conditions are not met.

2. **WithImageExporter godoc fixed**: Now correctly states "sets the ImageExporter on the Service and returns it for chaining" instead of the misleading "returns a copy" language.

3. **Remote temp file leak fixed**: `docker load` and `rm -f` are now split into separate executor calls. The cleanup runs unconditionally regardless of `docker load` outcome, with a clear `//nolint:errcheck // best-effort cleanup` annotation.

Full checklist verified: architecture rules followed (interface in ports, adapter in adapters, orchestration in app layer), all exported symbols have godoc, errors wrapped with context, table-driven tests where appropriate, no new dependencies, no security concerns.

tibtof:
Superseded by #950 (rebased on deploy redesign)

PR #967, opened 2026-04-19
fix(#953): generator overwrites merged config, health check uses wrong port

### LINKED ISSUE #953: bug: vibewarden.production.yaml not merged during deploy
The documented deep-merge of \`vibewarden.production.yaml\` on top of \`vibewarden.yaml\` during \`vibew deploy\` doesn't happen. The remote gets the base dev config (self-signed, port 8443). Agent had to manually merge configs.

Either the merge isn't wired up or the \`--env production\` default isn't triggering it.

tibtof:
## Summary
- Generator.Generate() was copying the unmerged vibewarden.yaml on top of the merged version. Swapped execution order.
- overlayProdConfig merges production overrides into the Config struct for compose template rendering.
- Health check now uses merged config's port/TLS instead of base config.

## Verified
- Deployed to demo.vibewarden.dev — Let's Encrypt cert, WAF 403 on SQLi, HSTS headers, port 443

Closes #953

### DIFF
--- a/internal/app/deploy/bundle.go
+++ b/internal/app/deploy/bundle.go
@@ -93,27 +93,40 @@ func (s *Service) BundleSidecar(_ context.Context, cfg *config.Config, outputDir
 
 // bundleSingleSite generates the complete single-site deploy bundle.
 func (s *Service) bundleSingleSite(ctx context.Context, cfg *config.Config, configPath, prodConfigPath, projectName string, outputDir string) error {
-	// Resolve upstream.host on the typed Config for template rendering.
-	resolved := ResolveProdConfig(cfg, projectName, false)
+	// Load and overlay the production override into the Config struct so that
+	// the compose template uses production values (port 443, letsencrypt, etc.)
+	// for port bindings, TLS config, and other template-driven settings.
+	mergedCfg := cfg
+	if prodConfigPath != "" {
+		if prodCfg, err := config.Load(prodConfigPath); err == nil {
+			mergedCfg = overlayProdConfig(cfg, prodCfg)
+		}
+	}
+
+	// Resolve upstream.host on the merged Config for template rendering.
+	resolved := ResolveProdConfig(mergedCfg, projectName, false)
 
 	// Set deploy mode so the template uses build context paths relative to
 	// the deploy bundle directory (e.g. "." instead of "../../.").
 	resolved.DeployMode = true
 
+	// Use the generator to produce docker-compose.yml, kratos/, credentials,
+	// etc. The generator writes directly into outputDir. This must run BEFORE
+	// the merged config write because the generator copies the original
+	// vibewarden.yaml into outputDir — the merged write overwrites it.
+	if err := s.generator.Generate(ctx, resolved.ToGeneratorInput(), outputDir); err != nil {
+		return fmt.Errorf("generating config files: %w", err)
+	}
+
 	// Build the merged YAML map for writing vibewarden.yaml.
 	// This preserves original field names (rate_limit, security_headers, etc.).
+	// Written AFTER generator.Generate() to overwrite the unmerged copy.
 	configData, err := buildMergedConfigYAML(configPath, prodConfigPath, projectName, false, cfg)
 	if err != nil {
 		return fmt.Errorf("building merged config YAML: %w", err)
 	}
 	if err := writeFile(filepath.Join(outputDir, "vibewarden.yaml"), configData); err != nil {
 		return fmt.Errorf("writing vibewarden.yaml to bundle: %w", err)
-	}
-
-	// Use the generator to produce docker-compose.yml, kratos/, credentials,
-	// etc. The generator writes directly into outputDir.
-	if err := s.generator.Generate(ctx, resolved.ToGeneratorInput(), outputDir); err != nil {
-		return fmt.Errorf("generating config files: %w", err)
 	}
 
 	return nil
--- a/internal/app/deploy/bundle.go
+++ b/internal/app/deploy/bundle.go
@@ -96,11 +96,9 @@ func (s *Service) bundleSingleSite(ctx context.Context, cfg *config.Config, conf
 	// Load and overlay the production override into the Config struct so that
 	// the compose template uses production values (port 443, letsencrypt, etc.)
 	// for port bindings, TLS config, and other template-driven settings.
-	mergedCfg := cfg
-	if prodConfigPath != "" {
-		if prodCfg, err := config.Load(prodConfigPath); err == nil {
-			mergedCfg = overlayProdConfig(cfg, prodCfg)
-		}
+	mergedCfg, err := loadMergedConfig(cfg, prodConfigPath)
+	if err != nil {
+		return err
 	}
 
 	// Resolve upstream.host on the merged Config for template rendering.
--- a/internal/app/deploy/resolve.go
+++ b/internal/app/deploy/resolve.go
@@ -160,3 +160,29 @@ func deepMerge(dst, src map[string]any) {
 		}
 	}
 }
+
+// overlayProdConfig creates a copy of base with non-zero values from prod
+// overlaid on top. Only fields relevant to deploy are checked: Server.Port,
+// TLS (Enabled, Provider, Domain), and Log.Level.
+func overlayProdConfig(base, prod *config.Config) *config.Config {
+	merged := *base
+	if prod.Server.Port != 0 {
+		merged.Server.Port = prod.Server.Port
+	}
+	if prod.TLS.Enabled {
+		merged.TLS.Enabled = prod.TLS.Enabled
+	}
+	if prod.TLS.Provider != "" {
+		merged.TLS.Provider = prod.TLS.Provider
+	}
+	if prod.TLS.Domain != "" {
+		merged.TLS.Domain = prod.TLS.Domain
+	}
+	if prod.Log.Level != "" {
+		merged.Log.Level = prod.Log.Level
+	}
+	if prod.WAF.Mode != "" {
+		merged.WAF.Mode = prod.WAF.Mode
+	}
+	return &merged
+}
--- a/internal/app/deploy/resolve.go
+++ b/internal/app/deploy/resolve.go
@@ -161,6 +161,22 @@ func deepMerge(dst, src map[string]any) {
 	}
 }
 
+// loadMergedConfig loads the production override file (if prodConfigPath is
+// non-empty) and overlays its values on top of base. When prodConfigPath is
+// empty the base config is returned unchanged. Errors from config.Load are
+// propagated so that syntax errors in the production YAML are never silently
+// ignored.
+func loadMergedConfig(base *config.Config, prodConfigPath string) (*config.Config, error) {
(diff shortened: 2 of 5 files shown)


internal/app/deploy/resolve.go
```diff
@@ -160,3 +160,29 @@ func deepMerge(dst, src map[string]any) {
...
+		merged.TLS.Domain = prod.TLS.Domain
+	}
+	if prod.Log.Level != "" {
+		merged.Log.Level = prod.Log.Level
+	}
+	if prod.WAF.Mode != "" {
+		merged.WAF.Mode = prod.WAF.Mode
+	}
+	return &merged
+}
```
tibtof, line 204: **Missing tests**: `overlayProdConfig` is a new function with conditional logic for 6 fields. It has no unit test. Per CLAUDE.md: "Every new `.go` file has a corresponding `_test.go`" -- and new functions in existing files need coverage too. This function is in `internal/app/`, which has an 80% coverage target.

**Why it matters**: If a future change adds a field to `Config` that needs overlaying, or if the zero-value guard logic is wrong (see related comment about `TLS.Enabled`), there is no safety net.

**Suggestion**: Add a table-driven test `TestOverlayProdConfig` in `resolve_test.go` covering:
- All fields overlaid when non-zero in prod
- All fields preserved from base when prod has zero values
- Original base config is not mutated (pointer safety check)
- `TLS.Enabled = false` in prod does not override base (documenting the current behavior)

internal/app/deploy/bundle.go
```diff
@@ -93,15 +93,34 @@ func (s *Service) BundleSidecar(_ context.Context, cfg *config.Config, outputDir
...
 func (s *Service) bundleSingleSite(ctx context.Context, cfg *config.Config, configPath, prodConfigPath, projectName string, outputDir string) error {
-	// Resolve upstream.host on the typed Config for template rendering.
-	resolved := ResolveProdConfig(cfg, projectName, false)
+	// Load and overlay the production override into the Config struct so that
+	// the compose template uses production values (port 443, letsencrypt, etc.)
+	// for port bindings, TLS config, and other template-driven settings.
+	mergedCfg := cfg
+	if prodConfigPath != "" {
+		if prodCfg, err := config.Load(prodConfigPath); err == nil {
+			mergedCfg = overlayProdConfig(cfg, prodCfg)
```
tibtof: **Swallowed error**: When `config.Load(prodConfigPath)` fails, the error is silently discarded and execution continues with the unmerged config. This violates the project rule: "never swallow errors."

**Why it matters**: If the production override file has a syntax error or references an environment variable that is not set, the deploy will silently use the base config. Port bindings, TLS, and health checks will all be wrong, and the user will have no indication of why.

**Suggestion**: Return the error with context:
```go
if prodConfigPath != "" {
    prodCfg, err := config.Load(prodConfigPath)
    if err != nil {
        return fmt.Errorf("loading production config overlay %s: %w", prodConfigPath, err)
    }
    mergedCfg = overlayProdConfig(cfg, prodCfg)
}
```

internal/app/deploy/service.go
```diff
@@ -281,13 +281,20 @@ func (s *Service) Deploy(ctx context.Context, cfg *config.Config, opts RunOption
 
 	// Step 6: health check -- run curl on the remote so the probe is independent
 	// of DNS propagation, external port availability, and TLS certificate issuance.
-	port := cfg.Server.Port
+	// Use the merged config (base + prod overlay) for the correct port and TLS state.
+	healthCfg := cfg
+	if opts.ProdConfigPath != "" {
+		if prodCfg, loadErr := config.Load(opts.ProdConfigPath); loadErr == nil {
+			healthCfg = overlayProdConfig(cfg, prodCfg)
```
tibtof: **Swallowed error (same pattern)**: Identical issue as in `bundleSingleSite` -- if the production config file is malformed, the error is silently discarded and the health check uses the wrong port/TLS settings.

**Why it matters**: A health check probing port 8443 over HTTP when the sidecar is listening on port 443 over HTTPS will always fail, producing a misleading "health check failed" error instead of the real cause.

**Suggestion**: Same fix -- return the error:
```go
if opts.ProdConfigPath != "" {
    prodCfg, loadErr := config.Load(opts.ProdConfigPath)
    if loadErr != nil {
        return fmt.Errorf("loading production config for health check: %w", loadErr)
    }
    healthCfg = overlayProdConfig(cfg, prodCfg)
}
```

internal/app/deploy/resolve.go
```diff
@@ -160,3 +160,29 @@ func deepMerge(dst, src map[string]any) {
...
+// overlaid on top. Only fields relevant to deploy are checked: Server.Port,
+// TLS (Enabled, Provider, Domain), and Log.Level.
+func overlayProdConfig(base, prod *config.Config) *config.Config {
+	merged := *base
+	if prod.Server.Port != 0 {
+		merged.Server.Port = prod.Server.Port
+	}
+	if prod.TLS.Enabled {
+		merged.TLS.Enabled = prod.TLS.Enabled
+	}
```
tibtof, line 190: **One-way TLS toggle**: The guard `if prod.TLS.Enabled` means the production overlay can turn TLS *on* but never turn it *off*. Since `false` is the zero value for `bool`, the condition is never satisfied when the prod config has `tls.enabled: false`.

**Why it matters**: This is likely fine for the current use case (prod enables TLS on top of a dev config), but it is an asymmetry that should at least be documented.  If a user later adds a staging overlay that disables TLS, it will silently fail.

**Suggestion (low priority)**: Either:
1. Add a comment documenting this is intentional ("prod overlay can only enable TLS, not disable it"), or
2. Change the overlay to always apply the prod value regardless of zero-ness for boolean fields, and document that omitted fields in prod config mean "keep base value" (which requires distinguishing "field present and false" from "field absent" -- may need a different approach like the YAML map merge that already exists in `buildMergedConfigYAML`).

internal/app/deploy/service.go
```diff
@@ -281,13 +281,20 @@ func (s *Service) Deploy(ctx context.Context, cfg *config.Config, opts RunOption
...
+		}
```
tibtof: **Duplicated overlay logic**: The pattern `config.Load(prodConfigPath) + overlayProdConfig(cfg, prodCfg)` is repeated identically in `bundleSingleSite` (bundle.go:100-103) and `Deploy` (service.go:286-289). If a third caller appears, or if error handling is fixed in one but not the other, they will drift.

**Why it matters**: DRY violation increases maintenance risk. Both callers also share the same silent-error bug, which is a sign they should be a single function.

**Suggestion**: Extract a helper:
```go
func (s *Service) loadMergedConfig(cfg *config.Config, prodConfigPath string) (*config.Config, error) {
    if prodConfigPath == "" {
        return cfg, nil
    }
    prodCfg, err := config.Load(prodConfigPath)
    if err != nil {
        return nil, fmt.Errorf("loading production config overlay: %w", err)
    }
    return overlayProdConfig(cfg, prodCfg), nil
}
```
Then call `s.loadMergedConfig(cfg, prodConfigPath)` in both places.

internal/app/deploy/bundle.go
```diff
@@ -93,15 +93,34 @@ func (s *Service) BundleSidecar(_ context.Context, cfg *config.Config, outputDir
...
 	// Set deploy mode so the template uses build context paths relative to
 	// the deploy bundle directory (e.g. "." instead of "../../.").
 	resolved.DeployMode = true
 
+	// Use the generator to produce docker-compose.yml, kratos/, credentials,
+	// etc. The generator writes directly into outputDir. This must run BEFORE
+	// the merged config write because the generator copies the original
+	// vibewarden.yaml into outputDir — the merged write overwrites it.
+	if err := s.generator.Generate(ctx, resolved.ToGeneratorInput(), outputDir); err != nil {
+		return fmt.Errorf("generating config files: %w", err)
```
tibtof, line 116: **Missing regression test for execution order fix**: The root cause of #953 was that `generator.Generate()` ran after the merged config write, overwriting it. The fix (moving Generate before the write) is correct, but there is no regression test proving the order invariant holds.

**Why it matters**: A future refactor could reorder these calls again without any test catching the regression.

**Suggestion**: Add a test like `TestBundleSingleSite_MergedConfigOverwritesGenerator` that:
1. Uses a `fakeGenerator` that writes a known sentinel value to `vibewarden.yaml` in the output dir (simulating the generator's copy behavior)
2. Calls `Bundle()` with a prod override
3. Reads the final `vibewarden.yaml` and asserts it contains the production override values (not the sentinel), proving the merged write happened last

tibtof (commented):
## Review Summary -- PR #967

**Core fix is correct.** The execution order swap in `bundleSingleSite` (Generate first, then merged config write) correctly resolves the root cause of #953. The `overlayProdConfig` function and health check merge are also sound in principle.

**Changes requested (3 blocking, 1 recommended, 1 low priority):**

### Blocking

1. **Swallowed errors (2 instances)** -- `config.Load(prodConfigPath)` errors are silently ignored in both `bundleSingleSite` (bundle.go:101-102) and `Deploy` (service.go:287-289). This violates the project's "never swallow errors" rule and will cause silent misconfiguration when the production override file is malformed.

2. **Missing tests for `overlayProdConfig`** -- New function with 6 conditional branches and no unit test. The `internal/app/` layer has an 80% coverage target.

3. **Missing regression test for the order fix** -- The root cause fix has no test proving the invariant (merged config write must follow generator). Without one, a future refactor can silently reintroduce #953.

### Recommended

4. **Duplicated overlay logic** -- The `config.Load + overlayProdConfig` pattern appears in two places. Extract a `loadMergedConfig` helper to prevent future drift and centralize error handling.

### Low Priority

5. **One-way TLS toggle** -- `overlayProdConfig` can enable TLS via prod overlay but cannot disable it (bool zero value issue). Document the limitation or fix.

Existing tests pass (`go test ./internal/app/deploy/...` clean). No architecture violations, no license issues, no security concerns. The diff is well-commented and follows project conventions.

tibtof (commented):
**Review: APPROVED** (submitted as comment because GitHub blocks self-approval)

All three previously requested fixes are resolved:

1. **Swallowed errors fixed.** `loadMergedConfig` propagates `config.Load` errors with context wrapping (`fmt.Errorf("loading production config %s: %w", ...)`). Both call sites in `bundleSingleSite` (bundle.go:99) and `Deploy` (service.go:285) return the error to the caller.

2. **`overlayProdConfig` tests added.** `TestOverlayProdConfig` is a thorough table-driven test with 6 cases covering: all-fields override, zero-value preservation for port and provider, WAF mode, all-zero prod passthrough, and base non-mutation. Covers every branch in the function.

3. **Regression test for write order added.** `TestArtifact_Bundle_GeneratorWriteOverwrittenByMerge` uses a `sentinelGenerator` fake to prove the invariant: merged config write must follow `generator.Generate()`. Sentinel is overwritten; production values (letsencrypt, port 443, domain) survive.

**Bonus:** The recommended `loadMergedConfig` helper was extracted, eliminating the duplicated `config.Load + overlayProdConfig` pattern between `bundleSingleSite` and `Deploy`.

`make check` passes clean. No architecture violations, no missing error handling, no license concerns.

Issue #976, opened 2026-04-19
feat: TLS provider alternatives (ZeroSSL, staging) + localhost cert trust automation


tibtof:
## Problems

1. Let's Encrypt rate limits (5 certs per domain per 168h) block test deploys
2. Self-signed localhost cert requires manual browser trust
3. No way to switch ACME CA without editing Caddy config

### ACME CA alternatives
\`\`\`yaml
tls:
  provider: letsencrypt    # default
  # provider: zerossl      # alternative free CA, different rate limits
  # provider: buypass      # another free ACME CA
  # acme_ca: https://acme-staging-v02.api.letsencrypt.org/directory  # staging (unlimited, not browser-trusted)
\`\`\`

### Localhost cert trust automation
\`vibew cert trust\` — exports the Caddy internal CA cert and installs it in the OS trust store:
- macOS: \`security add-trusted-cert\`
- Linux: copy to \`/usr/local/share/ca-certificates/\` + \`update-ca-certificates\`

Already partially exists via \`vibew cert export\` but the trust step is manual.

### Staging for test deploys
Consider defaulting to LE staging for repeated deploys to the same domain, or when the domain has hit rate limits. Caddy can fall back automatically.

## Documentation
- Document all TLS provider options in vibewarden.reference.yaml
- Add ACME CA override to llms-full.txt
- Document \`vibew cert export\` + trust workflow in getting-started

tibtof:
Status: READY_FOR_REVIEW
PR: https://github.com/VibeWarden/vibewarden/pull/980

tibtof:
Status: APPROVED — ready for human review

tibtof:
Fixed in v0.13.1 PRs.

Issue #974, opened 2026-04-19
feat: persist Let's Encrypt certs across deploys — don't delete Caddy data volume


tibtof:
## Problem

\`vibew deploy\` recreates containers with \`--force-recreate\`. If the previous deploy used \`docker compose down -v\`, the Caddy data volume (with Let's Encrypt certs) is deleted. The next deploy hits rate limits re-issuing certs.

## Fix

1. Never use \`docker compose down -v\` in the deploy path — only \`down\` (preserves volumes)
2. Name the Caddy cert volume explicitly and consider marking it \`external: true\`
3. Optionally: backup/restore cert data during deploy

tibtof:
Status: READY_FOR_REVIEW
PR: https://github.com/VibeWarden/vibewarden/pull/980

tibtof:
Status: APPROVED — ready for human review

tibtof:
Fixed in v0.13.1 PRs.

PR #980, opened 2026-04-19
fix(#972,#974,#976): build --platform, persist certs, TLS alternatives + cert trust

### LINKED ISSUE #972: bug: vibew build should detect target platform for deploy (arm64 → amd64 mismatch)
## Problem

\`vibew build\` builds for the local platform (e.g., arm64 on Apple Silicon). \`vibew deploy\` transfers the local image to the remote server which may be amd64. The platform mismatch causes the app container to fail with:

\`\`\`
The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64/v3)
\`\`\`

## Fix

\`vibew build\` should accept \`--platform\` flag (default: detect from \`--target\` if set, or build multi-arch). For deploy, always build for linux/amd64 unless the target is known to be arm64.

Alternative: \`vibew build --for-deploy\` that builds for linux/amd64 by default.

tibtof:
Closes #972, #974, #976

## Summary

- **#972 -- build --platform**: Added `--platform` flag to `vibew build` (e.g. `vibew build --platform linux/amd64`) for cross-platform Docker builds. The flag propagates through the `DockerBuilder` port, adapter, and application service.
- **#974 -- persist certs**: Confirmed the deploy code already preserves volumes (uses `docker compose up -d --force-recreate`, never `docker compose down -v`). Added explicit warning comments to the named `vibewarden-data` volume in both compose templates to prevent accidental cert deletion.
- **#976 -- TLS alternatives + cert trust**: Added `tls.acme_ca` config field to override the ACME directory URL (enables Let's Encrypt staging, ZeroSSL, Buypass, etc.). Added `vibew cert trust` command that installs Caddy's local CA root certificate into the OS trust store (macOS System keychain / Linux update-ca-certificates).

## Test plan

- [ ] `vibew build --platform linux/amd64` passes `--platform` to `docker build`
- [ ] `vibew build` without `--platform` omits the flag (backward compatible)
- [ ] `vibew cert trust --cert-path <path>` attempts OS trust store installation
- [ ] `vibew cert trust --help` shows macOS and Linux instructions
- [ ] `tls.acme_ca` set in config produces `"ca"` field in Caddy ACME issuer JSON
- [ ] `tls.acme_ca` empty produces no `"ca"` field (default Let's Encrypt production)
- [ ] `make check` passes (golangci-lint, build, tests)

### DIFF
--- a/internal/cli/cmd/cert.go
+++ b/internal/cli/cmd/cert.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 
@@ -25,7 +26,8 @@ func NewCertCmd() *cobra.Command {
 Examples:
   vibew cert export
   vibew cert export > vibewarden-ca.pem
-  vibew cert export --path`,
+  vibew cert export --path
+  vibew cert trust`,
 		// Default: print help when no subcommand is given.
 		Run: func(cmd *cobra.Command, _ []string) {
 			cmd.Help() //nolint:errcheck
@@ -33,6 +35,7 @@ Examples:
 	}
 
 	cmd.AddCommand(newCertExportCmd())
+	cmd.AddCommand(newCertTrustCmd())
 
 	return cmd
 }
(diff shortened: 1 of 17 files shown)


internal/cli/cmd/cert.go
```diff
@@ -170,3 +173,120 @@ func caddyUserDataDir() string {
...
+		return fmt.Errorf("copying certificate to %s: %w", destPath, err)
+	}
+
+	fmt.Fprintln(cmd.OutOrStdout(), "Updating CA certificates...")
+	//nolint:gosec // "sudo" + "update-ca-certificates" are standard system binaries
+	updateCmd := exec.CommandContext(cmd.Context(), "sudo", "update-ca-certificates")
+	updateCmd.Stdout = cmd.OutOrStdout()
+	updateCmd.Stderr = cmd.ErrOrStderr()
+	updateCmd.Stdin = os.Stdin
+	return updateCmd.Run()
```
tibtof, line 291: **Missing error wrapping**: `updateCmd.Run()` is returned bare, while the `cp` command error three lines above is properly wrapped with `fmt.Errorf("copying certificate to %s: %w", ...)`. This inconsistency means a failure in `update-ca-certificates` will surface as an opaque exec error with no VibeWarden context.

Wrap it the same way:
```go
if err := updateCmd.Run(); err != nil {
    return fmt.Errorf("running update-ca-certificates: %w", err)
}
return nil
```

internal/cli/cmd/cert.go
```diff
@@ -170,3 +173,120 @@ func caddyUserDataDir() string {
...
+	//nolint:gosec // certPath is validated above; "security" is a macOS system binary
+	c := exec.CommandContext(cmd.Context(), "sudo", "security", "add-trusted-cert",
+		"-d", "-r", "trustRoot",
+		"-k", "/Library/Keychains/System.keychain",
+		certPath,
+	)
+	c.Stdout = cmd.OutOrStdout()
+	c.Stderr = cmd.ErrOrStderr()
+	c.Stdin = os.Stdin // allow sudo password prompt
+	return c.Run()
```
tibtof, line 265: **Missing error wrapping**: Same issue as `installCertLinux` -- `c.Run()` is returned bare. When the `security add-trusted-cert` command fails, the user gets a raw exec error with no VibeWarden context explaining what was being attempted.

Wrap consistently:
```go
if err := c.Run(); err != nil {
    return fmt.Errorf("adding certificate to macOS keychain: %w", err)
}
return nil
```

tibtof (commented):
LGTM. Approving with two minor nits posted as inline comments (unwrapped errors in `installCertDarwin` and `installCertLinux`).

**What was reviewed:**

- **--platform flag (#972)**: Clean propagation from CLI flag through `BuildOptions` -> `BuildService` -> `DockerBuilder` port -> `BuildAdapter`. Table-driven tests at adapter, app, and CLI levels. Flag defaults to empty (backward compatible). Port interface updated correctly.
- **Persist certs (#974)**: Confirmed no `down -v` in any Go deploy code path. The deploy adapter uses `docker compose up -d --force-recreate` which preserves volumes. Warning comments added to both compose templates (`docker-compose.yml.tmpl` and `sidecar-compose.yml.tmpl`).
- **ACME CA override (#976)**: `tls.acme_ca` flows through `config.Config` -> wiring -> `ports.TLSConfig.ACMECA` -> Caddy adapter `buildLetsEncryptTLSApp`. Conditionally sets `"ca"` field only when non-empty. Table-driven test with 3 cases (empty, staging, ZeroSSL). Reference YAML documented.
- **cert trust command (#976)**: `vibew cert trust` with `--cert-path` override. Handles macOS (`security add-trusted-cert`) and Linux (`cp` + `update-ca-certificates`). Unsupported platforms get actionable error. Tests cover registration, flag presence, missing cert error, and help content.

**Architecture**: No violations. Domain layer untouched. Interfaces in `ports/`. Adapters in `adapters/`. App layer orchestrates only. No global state. DI used throughout. No new external dependencies (only `os/exec` from stdlib).

**Nits** (non-blocking): Two bare `return cmd.Run()` calls in `installCertDarwin` (line 265) and `installCertLinux` (line 291) should wrap errors with context for consistency with the rest of the file and the project convention.

Issue #979, opened 2026-04-19
bug: vibew dev build context ../../. builds wrong Dockerfile when run from subdirectory


tibtof:
## Problem

\`vibew dev\` runs \`docker compose up --build\` from \`.vibewarden/generated/\`. The compose template has \`build: context: ../../.\` which resolves relative to the compose file location, not the project root. When the project is inside a monorepo or subdirectory, this points to the wrong Dockerfile.

\`vibew build\` correctly builds from the project directory. But \`vibew dev\` then overwrites the correctly-built image with the wrong build context.

This is the fourth time this bug has surfaced — every agent test and manual test hits it.

## Fix

\`vibew dev\` should use the image built by \`vibew build\` instead of rebuilding via compose. Or the compose template's build context must resolve correctly regardless of where the compose file lives.

The deploy mode fix (DeployMode → \`context: .\`) doesn't apply to dev mode.

tibtof:
Status: READY_FOR_REVIEW
PR: https://github.com/VibeWarden/vibewarden/pull/981

tibtof:
Status: CHANGES_REQUESTED

PR #981 reviewed. The move to absolute build context via ProjectRoot is the right approach, but the current template drops the App.Build subdirectory value (e.g. `./src`) in non-deploy mode. See inline comments on PR #981.

tibtof:
Status: APPROVED -- ready for human review

tibtof:
Fixed in v0.13.1 PRs.

Issue #978, opened 2026-04-19
bug: Kratos auth redirect loop — public_paths not exempting /auth/login from auth middleware


tibtof:
## Problem

With \`auth.mode: kratos\`, the sidecar's auth middleware intercepts \`/auth/login\` even though it's listed in \`public_paths\`. This creates a redirect loop:

1. Browser hits \`/auth/login\` → sidecar checks auth → not authenticated → redirects to Kratos
2. Kratos creates flow → redirects to \`/auth/login?flow=...\`
3. Sidecar intercepts again → not authenticated → redirects to Kratos
4. Loop forever

The app correctly serves the login HTML when hit directly (bypassing the sidecar). The problem is in the Caddy auth handler's path exemption logic.

## Steps to reproduce

1. \`cd examples/demo-app\`
2. \`vibew generate && docker compose -f .vibewarden/generated/docker-compose.yml up -d\`
3. Open https://localhost:8443/auth/login
4. ERR_TOO_MANY_REDIRECTS

## Fix

The Caddy auth handler module must check \`public_paths\` BEFORE redirecting unauthenticated requests. Investigate \`internal/adapters/caddy/auth_handler.go\` for the path exemption logic.

tibtof:
Status: READY_FOR_REVIEW
PR: https://github.com/VibeWarden/vibewarden/pull/981

tibtof:
Status: CHANGES_REQUESTED

PR #981 reviewed. The Kratos public paths auto-append logic is correct, but a duplicate `kratosFlowPaths` list in `internal/plugins/auth/plugin.go` was not updated to match the changes in `internal/adapters/caddy/config_routes.go`. See inline comments on PR #981.

tibtof:
Status: APPROVED -- ready for human review

tibtof:
Fixed in v0.13.1 PRs.

Issue #977, opened 2026-04-19
fix: auto-add Kratos flow paths to public_paths when auth.mode: kratos


tibtof:
## Problem

When \`auth.mode: kratos\`, the auth middleware requires \`/login\`, \`/registration\`, \`/recovery\`, \`/verification\`, \`/error\`, and \`/self-service/*\` to be public. If they're not in \`public_paths\`, the sidecar creates a redirect loop — the auth middleware intercepts the login page itself.

Users have to manually list all Kratos flow paths. This is error-prone and undocumented.

## Fix

When \`auth.mode: kratos\`, automatically append these paths to the effective \`public_paths\` (in the Caddy config, not in the user's YAML):
- \`/login\`
- \`/registration\`
- \`/recovery\`
- \`/verification\`
- \`/error\`
- \`/self-service/*\`

The user's \`public_paths\` list is additive on top of these defaults.

tibtof:
Status: READY_FOR_REVIEW
PR: https://github.com/VibeWarden/vibewarden/pull/981

tibtof:
Status: CHANGES_REQUESTED

PR #981 reviewed. Kratos base_url, ui_url /auth/ prefix, and flow proxy path changes look correct. Blocked on the two issues found in the PR (build context subdirectory bug and kratosFlowPaths inconsistency). See inline comments on PR #981.

tibtof:
Status: APPROVED -- ready for human review

tibtof:
Fixed in v0.13.1 PRs.

PR #981, opened 2026-04-19
fix(#973,#977-979): absolute build context, Kratos public paths, image naming, auth flow fixes

### LINKED ISSUE #973: bug: vibew build image name doesn't match vibew deploy expected name
## Problem

\`vibew build\` names the image based on directory name (e.g., \`vibew-kratos-demo:latest\`).
\`vibew deploy\` expects \`ComposeProjectName()-app:latest\` (e.g., \`kratos-demo-app:latest\`).

These don't match when \`--name\` is used or when the directory name differs from the project name.

## Fix

\`vibew build\` should use the same naming convention as deploy: \`ComposeProjectName()-app:latest\`. Or both should derive from the same source.

tibtof:
Closes #973, #977, #978, #979

## Summary

- **#979 compose build context**: Replaced fragile relative `../../` path in docker-compose.yml.tmpl with absolute `ProjectRoot` resolved at generation time. Also fixed seed-users volume mount.
- **#978 + #977 Kratos auth public paths**: Auth handler now auto-appends default Kratos public paths (`/auth/*`, `/self-service/*`, `/login`, `/registration`, `/recovery`, `/verification`, `/error`, `/sessions/whoami`) in `Provision()` to prevent redirect loops.
- **#973 vibew build image name**: `vibew build` now names images as `<ComposeProjectName>-app:latest` to match Docker Compose's naming convention, ensuring the built image is found by `vibew dev` and `vibew deploy`.
- **Kratos base_url**: `serve.public.base_url` in kratos.yml now uses the sidecar's public URL (`http[s]://localhost:<port>`) instead of the internal Docker hostname.
- **Kratos ui_url paths**: All flow `ui_url` entries now use `/auth/` prefix to match the app's route convention.
- **Kratos flow proxy paths**: Added exact paths (without wildcard) for POST support and `/sessions/whoami` endpoint.
- **Demo-app login/register forms**: Submit buttons now include `name` and `value` attributes required by Kratos for method selection.

## Test plan

- [x] `TestGenerate_Integration_ComposeBuildContextIsAbsolutePath` -- verifies compose build context is absolute, not relative `../../`
- [x] `TestAuthHandler_Provision_AppendKratosPublicPaths` -- verifies auto-append of Kratos default public paths, no duplicates
- [x] `TestAuthHandler_Provision_KratosPathsBypass` -- verifies auth/self-service/login paths bypass authentication
- [x] `TestBuildService_Run_ImageNameMatchesComposeProjectName` -- verifies `vibew build` produces `<name>-app:latest` matching compose
- [x] `TestGenerate_Integration_KratosBaseURL` -- verifies base_url uses sidecar public URL for both HTTP and TLS modes
- [x] `TestGenerate_Integration_KratosUIUrlsUseAuthPrefix` -- verifies all ui_url entries use `/auth/` prefix
- [x] `make check` passes (golangci-lint, go build, go test -race, demo-app checks)

Generated with Claude Code

### DIFF
--- a/internal/adapters/caddy/config_routes.go
+++ b/internal/adapters/caddy/config_routes.go
@@ -9,13 +9,20 @@ import (
 // the Kratos public API instead of the upstream application.
 // These paths are the Kratos self-service browser flows and the Ory canonical prefix.
 var kratosFlowPaths = []string{
+	"/self-service/login",
 	"/self-service/login/*",
+	"/self-service/registration",
 	"/self-service/registration/*",
+	"/self-service/logout",
 	"/self-service/logout/*",
+	"/self-service/settings",
 	"/self-service/settings/*",
+	"/self-service/recovery",
 	"/self-service/recovery/*",
+	"/self-service/verification",
 	"/self-service/verification/*",
 	"/.ory/kratos/public/*",
+	"/sessions/whoami",
 }
 
 // buildKratosFlowRoute constructs a Caddy route that transparently proxies all
--- a/internal/app/generate/service_test.go
+++ b/internal/app/generate/service_test.go
@@ -599,13 +599,17 @@ func TestGenerate_AppService_BothSet_BuildTakesPrecedence(t *testing.T) {
 	if !bytes.Contains(compose, []byte("build:")) {
 		t.Error("expected 'build:' directive when both build and image are set")
 	}
-	// The build context should be an absolute path (ProjectRoot), not a
+	// The build context should be an absolute path (ProjectRoot/src), not a
 	// relative "../.././src" prefix.
 	if bytes.Contains(compose, []byte("context: ../../")) {
 		t.Error("build context must be an absolute path, not relative '../../'")
 	}
 	if !bytes.Contains(compose, []byte("context: /")) {
 		t.Error("expected build context to be an absolute path starting with '/'")
+	}
+	// The subdirectory "src" must be preserved in the build context path.
+	if !bytes.Contains(compose, []byte("/src")) {
+		t.Error("build context must include the 'src' subdirectory from App.Build")
 	}
 	// image: for the app service must not appear when build takes precedence.
 	// Note: "image:" still appears in vibewarden service itself so we check
--- a/internal/app/generate/service_test.go
+++ b/internal/app/generate/service_test.go
@@ -563,8 +563,14 @@ func TestGenerate_AppService_BuildMode(t *testing.T) {
 	if !bytes.Contains(compose, []byte("build:")) {
 		t.Error("expected 'build:' directive")
 	}
-	if !bytes.Contains(compose, []byte("context: ../../.")) {
-		t.Error("expected 'context: ../../.' (build path prefixed with ../../ for compose-file-relative resolution)")
+	// The build context should be an absolute path (ProjectRoot), not a
+	// relative "../../." prefix. The absolute path is resolved from cwd at
+	// generation time.
+	if bytes.Contains(compose, []byte("context: ../../.")) {
+		t.Error("build context must be an absolute path, not relative '../../.'")
+	}
+	if !bytes.Contains(compose, []byte("context: /")) {
+		t.Error("expected build context to be an absolute path starting with '/'")
 	}
 	if bytes.Contains(compose, []byte("image: ${VIBEWARDEN_APP_IMAGE")) {
 		t.Error("image: directive must not appear in build mode")
@@ -593,8 +599,13 @@ func TestGenerate_AppService_BothSet_BuildTakesPrecedence(t *testing.T) {
 	if !bytes.Contains(compose, []byte("build:")) {
 		t.Error("expected 'build:' directive when both build and image are set")
 	}
-	if !bytes.Contains(compose, []byte("context: ../.././src")) {
-		t.Error("expected 'context: ../.././src' (build path prefixed with ../../ for compose-file-relative resolution)")
+	// The build context should be an absolute path (ProjectRoot), not a
+	// relative "../.././src" prefix.
+	if bytes.Contains(compose, []byte("context: ../../")) {
+		t.Error("build context must be an absolute path, not relative '../../'")
+	}
+	if !bytes.Contains(compose, []byte("context: /")) {
+		t.Error("expected build context to be an absolute path starting with '/'")
 	}
 	// image: for the app service must not appear when build takes precedence.
 	// Note: "image:" still appears in vibewarden service itself so we check
--- a/internal/config/templates/docker-compose.yml.tmpl
+++ b/internal/config/templates/docker-compose.yml.tmpl
@@ -41,7 +41,7 @@ services:
     restart: unless-stopped
 {{- if .App.Build }}
     build:
-      context: {{ if .DeployMode }}{{ .App.Build }}{{ else }}{{ .ProjectRoot }}{{ end }}
+      context: {{ .ResolvedBuildContext }}
 {{- else if .App.Image }}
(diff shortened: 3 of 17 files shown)


internal/config/templates/docker-compose.yml.tmpl
```diff
@@ -41,7 +41,7 @@ services:
     restart: unless-stopped
 {{- if .App.Build }}
     build:
-      context: {{ if .DeployMode }}{{ .App.Build }}{{ else }}../../{{ .App.Build }}{{ end }}
+      context: {{ if .DeployMode }}{{ .App.Build }}{{ else }}{{ .ProjectRoot }}{{ end }}
```
tibtof: **Bug -- build context drops subdirectory**: When `App.Build` is a subdirectory like `"./src"`, this template renders `context: <ProjectRoot>` and silently drops the subdirectory. The old template was `context: ../../{{ .App.Build }}` which preserved the subdirectory (e.g. `context: ../.././src`).

With the current code, a user who sets `app.build: ./src` would get their build context pointed at the project root instead of the `src` subdirectory, causing Docker to build from the wrong directory.

**Suggested fix**: Include `App.Build` in the non-deploy path:
```
context: {{ if .DeployMode }}{{ .App.Build }}{{ else }}{{ joinPath .ProjectRoot .App.Build }}{{ end }}
```
Or resolve this in Go code before template rendering, e.g. `cfg.ProjectRoot = filepath.Join(projectRoot, cfg.App.Build)` and clear `cfg.App.Build` for the dev-mode template path. The existing test for `BothSet_BuildTakesPrecedence` only checks `context: /` which passes but does not verify the subdirectory is preserved.

internal/adapters/caddy/config_routes.go
```diff
@@ -9,13 +9,20 @@ import (
...
 	"/self-service/logout/*",
+	"/self-service/settings",
 	"/self-service/settings/*",
+	"/self-service/recovery",
 	"/self-service/recovery/*",
+	"/self-service/verification",
 	"/self-service/verification/*",
 	"/.ory/kratos/public/*",
+	"/sessions/whoami",
 }
```
tibtof, line 26: **Consistency issue -- duplicate kratosFlowPaths not updated**: `internal/plugins/auth/plugin.go` (line 25) declares its own `kratosFlowPaths` variable with the old set of paths (wildcard-only, no exact matches, no `/sessions/whoami`). This PR added exact paths and `/sessions/whoami` to `internal/adapters/caddy/config_routes.go` but did not update the copy in `internal/plugins/auth/plugin.go`.

The two lists now disagree, which means the auth plugin's flow-path routing may reject POST requests to exact self-service endpoints that the Caddy adapter would correctly proxy.

**Suggested fix**: Either update `internal/plugins/auth/plugin.go:kratosFlowPaths` to match, or better yet, extract a single shared list (e.g. in `internal/domain/` or `internal/ports/`) and import it from both locations. Having the same list duplicated in two packages is fragile.

internal/app/generate/service_test.go
```diff
@@ -593,8 +599,13 @@ func TestGenerate_AppService_BothSet_BuildTakesPrecedence(t *testing.T) {
...
 	}
-	if !bytes.Contains(compose, []byte("context: ../.././src")) {
-		t.Error("expected 'context: ../.././src' (build path prefixed with ../../ for compose-file-relative resolution)")
+	// The build context should be an absolute path (ProjectRoot), not a
+	// relative "../.././src" prefix.
+	if bytes.Contains(compose, []byte("context: ../../")) {
+		t.Error("build context must be an absolute path, not relative '../../'")
+	}
+	if !bytes.Contains(compose, []byte("context: /")) {
+		t.Error("expected build context to be an absolute path starting with '/'")
```
tibtof, line 608: **Test gap -- subdirectory not verified**: This test checks that the build context is an absolute path (`context: /`), but it does not verify that the `./src` subdirectory is preserved in the final path. With the current template logic (`{{ .ProjectRoot }}` without `{{ .App.Build }}`), this test passes even though the subdirectory is silently dropped.

**Suggested fix**: Add an assertion that verifies the rendered context includes the subdirectory, e.g.:
```go
// When App.Build is "./src", the rendered context must end with "/src".
if !bytes.Contains(compose, []byte("context: "+cfg.ProjectRoot+"/src")) {
    t.Errorf("build context should include subdirectory /src, got:\n%s", compose)
}
```

tibtof (commented):
## Review Summary -- CHANGES REQUESTED

This PR addresses four issues (#979, #978, #977, #973) covering absolute build context, Kratos public path auto-append, build image naming, and auth flow fixes. The overall direction is correct and the test coverage for the new functionality is solid. However, two issues need to be fixed before merging:

### Must fix

1. **Build context drops subdirectory (bug)**: `docker-compose.yml.tmpl` line 44 renders `context: {{ .ProjectRoot }}` in non-deploy mode, which silently drops the `App.Build` subdirectory. When a user sets `app.build: ./src`, the build context would point to the project root instead of the `src` subdirectory. The old `../../{{ .App.Build }}` preserved the subdirectory. The test for `BothSet_BuildTakesPrecedence` only checks `context: /` and does not catch this regression.

2. **Duplicate `kratosFlowPaths` not updated**: `internal/plugins/auth/plugin.go` declares its own `kratosFlowPaths` (line 25) with the old wildcard-only entries. This PR updated `internal/adapters/caddy/config_routes.go` with exact paths and `/sessions/whoami` but did not update the copy in the auth plugin. The two lists are now inconsistent. Consider extracting a shared list to avoid future divergence.

### What looks good

- `kratosDefaultPublicPaths` and `appendKratosPublicPaths` logic is clean, with proper deduplication.
- `resolveImageTag` correctly derives `<ComposeProjectName>-app:latest` to match Docker Compose naming.
- Kratos `base_url` correctly uses TLS-aware scheme.
- Kratos `ui_url` entries correctly use `/auth/` prefix.
- Login/register form submit buttons now include name/value attributes for Kratos method selection.
- Tests are table-driven, well-named, and cover the key scenarios.
- Error wrapping follows project conventions.
- No license, security, or architecture violations found.

tibtof (commented):
LGTM. Second review -- both previously requested fixes are resolved:

1. **Build context subdirectory fix**: `ResolvedBuildContext()` method correctly preserves the `App.Build` subdirectory by joining `ProjectRoot` with the cleaned subdirectory path. The template now uses `{{ .ResolvedBuildContext }}` instead of inline logic. Tests cover ".", empty, "./src", "src", "./apps/web", and deploy-mode cases.

2. **kratosFlowPaths sync fix**: Both `internal/adapters/caddy/config_routes.go` and `internal/plugins/auth/plugin.go` now have identical 14-entry `kratosFlowPaths` lists (6 exact paths + 7 wildcards + `/sessions/whoami`). The test in `plugin_test.go` was also updated to assert all 14 entries.

**Full checklist verified on re-review:**
- Architecture: no violations. Domain layer has zero external deps. Interfaces in ports. Adapters in adapters.
- Code quality: all new exported symbols have godoc comments. Errors wrapped with context. No swallowed errors.
- Testing: all new code has corresponding tests. Table-driven tests used throughout. Descriptive test names.
- Go idioms: `ResolvedBuildContext()` is a pure method. `appendKratosPublicPaths` returns a new slice (no mutation). `resolveProjectRoot` validates inputs and returns errors.
- Security: no hardcoded secrets. HTML escapes applied to user-supplied `attrs.name` and `attrs.value` in login/register templates.
- Licenses: no new dependencies added.

**Note (not blocking)**: The `ui_url` entries in `kratos.yml.tmpl` are still hardcoded to `http://` while `base_url` is now TLS-aware. This is a pre-existing inconsistency not introduced by this PR -- consider addressing it in a follow-up issue.

Status: APPROVED -- ready for human review.

PR #1056, opened 2026-04-20
fix(#1053): route prod-override through YAML merge, strict validate

### LINKED ISSUE #1053: fix: production config override silently drops unknown keys (tls.acme_ca, tls.email)
## Problem

When \`vibewarden.production.yaml\` adds keys under existing sections (e.g. \`tls.acme_ca\`, \`tls.email\`), they are silently dropped during the merge. \`vibew validate\` accepts the prod-override file; \`vibew deploy --dry-run\` regenerates the bundle; the merged \`vibewarden.yaml\` in the bundle **does not contain** the added keys. No warning, no error.

## Impact

Deployed sidecar uses Let's Encrypt with no email set, despite the prod override declaring \`tls.email: ops@example.com\` and \`tls.acme_ca: https://acme.zerossl.com/v2/DV90\`. The user's only signal that the merge failed is \`curl\`ing the prod URL, seeing TLS provisioning errors, SSHing in, and reading the live \`vibewarden.yaml\` on the remote.

## Reproduction

1. \`vibew init my-app && cd my-app\`
2. Add to \`vibewarden.production.yaml\`:
   \`\`\`yaml
   tls:
     acme_ca: https://acme.zerossl.com/v2/DV90
     email: ops@example.com
   \`\`\`
3. \`vibew validate\` — passes
4. \`vibew deploy --dry-run\` — regenerates \`.vibewarden/deploy/\`
5. \`cat .vibewarden/deploy/vibewarden.yaml\` — neither key is present

## Root cause hypothesis

The merge logic for production overrides is probably either:
- Using a static allow-list of keys that lists \`domain\`, \`provider\`, \`auto_https\`, but not \`acme_ca\` / \`email\`, or
- Failing silently on schema mismatch (e.g. the prod-override schema omits those fields)

## Acceptance criteria

- [ ] Any valid key under \`tls\` (or any other plugin section) added in \`vibewarden.production.yaml\` is preserved in the merged output
- [ ] If a key is schema-unknown, loading fails loudly with file+line reference (no silent drop)
- [ ] Test: table-driven unit test that writes base + prod-override, asserts merged struct has all keys
- [ ] Test: \`vibew validate\` rejects unknown keys in either file with actionable error
- [ ] No regression in existing override tests

## Found in

v0.15.0 fresh-agent retrospective, qr-andy-warhol demo deploy (2026-04-20)

tibtof:
Fixes #1053

## Summary

Replaces the hand-written allow-list in `overlayProdConfig` that silently dropped schema-valid fields (`tls.email`, `tls.acme_ca`, `tls.cert_monitoring.*`, `server.host`, `rate_limit.enabled`, and any future plugin key) whenever they were set only in `vibewarden.production.yaml`.

`LoadMergedConfig` now reads both files, routes through `MergeConfigYAML` — the same deep-merge that feeds the on-disk bundle — marshals to a tempfile, and calls `config.Load`. One canonical merge path feeds both the bundle YAML and the runtime `*config.Config` used for template rendering and the deploy health check.

Adds `config.LoadStrict` + `*UnknownKeyError` for `vibew validate` and `vibew deploy`. Unknown keys in either file (e.g. typo `tls.dmain`) now fail loudly with an error naming the file and the offending dotted keys. The runtime loader (`vibewarden serve`) is unchanged for forward-compat per ADR-065.

See [ADR-082](decisions/adr-082-strict-config-merge-unknown-keys-fail-loudly.md) for the full design. Regression guard for ADR-078 (`tls.email` wired to single-site ACME) and ADR-079 (`tls.acme_ca` fallback chain).

### Breaking

- `vibew validate` / `vibew deploy` now reject unknown keys with an actionable error. Typos that previously silently loaded (e.g. `tls.dmain: example.com`) now fail. Migration: move scratch annotations to YAML comments (`# …`). Documented in CHANGELOG under Unreleased.

## Files changed

- `internal/config/strict.go` (new) — `LoadStrict`, `UnknownKeyError`, schema reflection walker
- `internal/config/strict_test.go` (new) — table-driven tests for strict mode
- `internal/app/deploy/resolve.go` — `LoadMergedConfig` rewritten, `overlayProdConfig` deleted
- `internal/app/deploy/resolve_test.go` — replaces `TestOverlayProdConfig` with `TestLoadMergedConfig_PreservesAllProdFields` + env-precedence + no-override guards
- `internal/app/deploy/bundle.go`, `service.go` — pass `configPath` to `LoadMergedConfig`; fall back to supplied cfg when the file is not on disk
- `internal/app/deploy/bundle_test.go` — extend `TestBundle_WithProdOverride_MergesCorrectly` with `tls.email` + `tls.acme_ca` assertions on both on-disk YAML and the returned `*Config`
- `internal/cli/cmd/validate.go` — switches to `LoadStrict`, auto-discovers `vibewarden.production.yaml` next to the base
- `internal/cli/cmd/deploy.go` — calls `LoadStrict` before writing any bundle files
- `CHANGELOG.md` — Unreleased Breaking + Fixes entries

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run ./...` (0 issues)
- [x] `go test -race ./...` (all packages pass)
- [x] `make check` (full pipeline green)
- [x] New tests exercise: happy path for every field class named in the spec (TLS, cert monitoring, non-TLS section), env-var precedence over prod override, unknown key in base file, unknown key in prod file, multiple unknowns sorted, nested unknown under valid parent, error message contains file + key
- [x] Existing `TestBundle_WithProdOverride_MergesCorrectly` extended to assert `tls.email` + `tls.acme_ca` land in both the on-disk bundle YAML and the merged `*Config` (runtime parity)

### DIFF
--- a/internal/app/deploy/service.go
+++ b/internal/app/deploy/service.go
@@ -345,9 +345,17 @@ func (s *Service) Deploy(ctx context.Context, cfg *config.Config, opts RunOption
 	// Step 6: health check -- run curl on the remote so the probe is independent
 	// of DNS propagation, external port availability, and TLS certificate issuance.
 	// Use the merged config (base + prod overlay) for the correct port and TLS state.
-	healthCfg, err := LoadMergedConfig(cfg, opts.ProdConfigPath)
-	if err != nil {
-		return err
+	// When opts.ConfigPath is empty (programmatic callers that build cfg in-memory)
+	// or when the file is not on disk (unit tests), fall back to the supplied cfg.
+	healthCfg := cfg
+	if opts.ConfigPath != "" {
+		if _, statErr := os.Stat(opts.ConfigPath); statErr == nil {
+			loaded, err := LoadMergedConfig(opts.ConfigPath, opts.ProdConfigPath)
+			if err != nil {
+				return err
+			}
+			healthCfg = loaded
+		}
 	}
 	port := healthCfg.Server.Port
 	if port == 0 {
--- a/internal/app/deploy/service.go
+++ b/internal/app/deploy/service.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
@@ -349,12 +350,19 @@ func (s *Service) Deploy(ctx context.Context, cfg *config.Config, opts RunOption
 	// or when the file is not on disk (unit tests), fall back to the supplied cfg.
 	healthCfg := cfg
 	if opts.ConfigPath != "" {
-		if _, statErr := os.Stat(opts.ConfigPath); statErr == nil {
+		_, statErr := os.Stat(opts.ConfigPath)
+		switch {
+		case statErr == nil:
 			loaded, err := LoadMergedConfig(opts.ConfigPath, opts.ProdConfigPath)
 			if err != nil {
 				return err
 			}
 			healthCfg = loaded
+		case errors.Is(statErr, fs.ErrNotExist):
+			// Intentional fall-through: caller supplied an in-memory cfg
+			// (programmatic or unit-test entrypoint); no disk file to merge.
+		default:
+			return fmt.Errorf("stat config %s: %w", opts.ConfigPath, statErr)
 		}
 	}
 	port := healthCfg.Server.Port
--- a/internal/cli/cmd/validate.go
+++ b/internal/cli/cmd/validate.go
@@ -1,8 +1,10 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 
@@ -102,8 +104,18 @@ Examples:
 				}
 			}
 
-			cfg, err := config.Load(configPath)
+			// Strict mode: reject unknown keys in the base file and in the
+			// production override (when discovered) so typos like tls.dmain fail
+			// loudly at validate time instead of being silently dropped at merge
+			// time. The runtime loader (config.Load) stays lenient per ADR-065.
+			prodPath := discoverProdOverride(configPath)
+			cfg, err := config.LoadStrict(configPath, prodPath)
 			if err != nil {
+				var unknown *config.UnknownKeyError
+				if errors.As(err, &unknown) {
+					fmt.Fprintf(cmd.ErrOrStderr(), "Configuration invalid: %s\n", unknown.Error())
+					return fmt.Errorf("loading config: %w", err)
+				}
 				fmt.Fprintf(cmd.ErrOrStderr(), "Configuration invalid: %v\n", err)
 				return fmt.Errorf("loading config: %w", err)
 			}
@@ -134,6 +146,23 @@ Examples:
 	}
 
 	return cmd
+}
+
+// discoverProdOverride returns the path to vibewarden.production.yaml that
+// sits next to configPath, when it exists. The empty string is returned when
+// configPath is empty, the override file does not exist, or the directory
+// cannot be read. Strict validation still succeeds when no override is
+// present.
+func discoverProdOverride(configPath string) string {
+	if configPath == "" {
+		return ""
+	}
+	dir := filepath.Dir(configPath)
+	candidate := filepath.Join(dir, "vibewarden.production.yaml")
+	if _, err := os.Stat(candidate); err != nil {
+		return ""
+	}
+	return candidate
 }
 
 // validateConfig checks semantic constraints on cfg that cannot be expressed
--- a/internal/config/strict.go
+++ b/internal/config/strict.go
@@ -1,0 +1,209 @@
+package config
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"reflect"
+	"sort"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// UnknownKeyError reports keys present in a YAML source that do not map to any
+// field in the Config schema. File is the path the keys were read from (empty
+// when the keys come from an unnamed source). Keys contains the offending
+// dotted paths, sorted alphabetically — for example
+// []string{"tls.dmain", "unknown_plugin.enabled"}.
+//
+// Use errors.As to unwrap this error and inspect File/Keys programmatically.
+type UnknownKeyError struct {
+	File string
+	Keys []string
+}
+
+// Error returns a human-readable description of the unknown-key error.
+// The message names the offending file (when known) and lists the keys in
+// alphabetical order so the output is deterministic across runs.
+func (e *UnknownKeyError) Error() string {
+	if len(e.Keys) == 0 {
+		if e.File != "" {
+			return fmt.Sprintf("config %s: unknown key(s)", e.File)
+		}
+		return "config: unknown key(s)"
+	}
+	if e.File != "" {
+		return fmt.Sprintf("config %s: unknown key(s): %s", e.File, strings.Join(e.Keys, ", "))
+	}
+	return fmt.Sprintf("config: unknown key(s): %s", strings.Join(e.Keys, ", "))
+}
+
+// LoadStrict behaves like Load but additionally rejects any key present in the
+// base config file (configPath) or the production-override file
+// (prodConfigPath) that does not map to a field in Config. The strict check is
+// independent of viper's decoder (ADR-065 keeps ErrorUnused=false on the
+// runtime path) — it parses the YAML file(s) directly and walks the
+// mapstructure tags on *Config via reflection.
+//
+// Callers:
+//   - vibew validate (always).
+//   - vibew deploy / vibew bundle, before writing any bundle files.
+//
+// The runtime loader (config.Load, used by vibewarden serve) is unchanged for
+// forward-compat per ADR-065.
+//
+// When a file does not exist it is skipped silently, matching config.Load's
+// behaviour (viper treats a missing config file as non-fatal). Passing an
+// empty configPath or prodConfigPath skips that file.
+//
+// On unknown keys, LoadStrict returns an error that wraps *UnknownKeyError.
+// The error message names the file and the offending dotted keys.
+func LoadStrict(configPath, prodConfigPath string) (*Config, error) {
+	if configPath != "" {
+		if err := checkUnknownKeys(configPath); err != nil {
+			return nil, err
+		}
+	}
+	if prodConfigPath != "" {
+		if err := checkUnknownKeys(prodConfigPath); err != nil {
+			return nil, err
+		}
+	}
+
+	// Base load uses the same code path as vibew serve so defaults, env-var
+	// overrides, and validation rules stay in lock-step with runtime.
+	cfg, err := Load(configPath)
+	if err != nil {
+		return nil, err
+	}
+	return cfg, nil
+}
+
+// checkUnknownKeys parses file as YAML and reports any top-level or nested key
+// that does not map to a field in *Config. It returns nil when the file does
+// not exist (parity with viper's non-fatal handling of missing files) or when
+// every key maps cleanly.
+func checkUnknownKeys(file string) error {
+	data, err := os.ReadFile(file) //nolint:gosec // file is a config path supplied by the caller (vibew validate / deploy)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return fmt.Errorf("reading config %s: %w", file, err)
+	}
+
+	var tree map[string]any
+	if err := yaml.Unmarshal(data, &tree); err != nil {
+		return fmt.Errorf("parsing config %s: %w", file, err)
+	}
+	if len(tree) == 0 {
+		return nil
+	}
+
+	var unknown []string
+	walkUnknownKeys(tree, reflect.TypeOf(Config{}), "", &unknown)
+	if len(unknown) == 0 {
+		return nil
+	}
+	sort.Strings(unknown)
+	return &UnknownKeyError{File: file, Keys: unknown}
+}
+
+// walkUnknownKeys recursively compares the YAML tree against the schema type.
+// Any key in tree that has no corresponding mapstructure tag on schema (or any
+// nested struct in schema) is appended to out, prefixed by the dotted path
+// that leads to it.
+//
+// Keys whose schema field has mapstructure tag "-" (e.g. DeployMode,
+// ProjectRoot) are treated as unknown — they are never expected to appear in
+// YAML.
+//
+// Non-struct schema types accept any child (e.g. map[string]any, interface{},
+// map[string]string). In those cases recursion stops — we cannot verify
+// arbitrary map values against a fixed schema.
+func walkUnknownKeys(tree map[string]any, schema reflect.Type, prefix string, out *[]string) {
+	// Unwrap pointers; non-struct schemas accept anything.
+	for schema.Kind() == reflect.Ptr {
+		schema = schema.Elem()
+	}
+	if schema.Kind() != reflect.Struct {
+		return
+	}
+
+	fields := mapstructureFields(schema)
+	for key, val := range tree {
+		field, ok := fields[key]
+		if !ok {
+			*out = append(*out, joinKey(prefix, key))
+			continue
+		}
+
+		child, isMap := val.(map[string]any)
+		if !isMap {
+			continue
+		}
+
+		fieldType := field.Type
+		for fieldType.Kind() == reflect.Ptr {
+			fieldType = fieldType.Elem()
+		}
+		// Recurse into nested structs. For slice/map fields we cannot check
+		// individual entries without more schema metadata, so we skip them.
+		if fieldType.Kind() == reflect.Struct {
+			walkUnknownKeys(child, fieldType, joinKey(prefix, key), out)
+		}
+	}
+}
+
+// mapstructureFields returns a map from mapstructure tag → reflect.StructField
+// for every exported field of schema. Fields tagged mapstructure:"-" are
+// excluded; callers will report them as unknown if they appear in YAML. When a
+// field has no mapstructure tag, the lowercased field name is used (matching
+// viper's default behaviour).
+//
+// Embedded structs (anonymous fields) are flattened: their fields appear at the
+// parent's level, mirroring mapstructure's default squash behaviour.
+func mapstructureFields(schema reflect.Type) map[string]reflect.StructField {
+	out := make(map[string]reflect.StructField, schema.NumField())
+	for i := 0; i < schema.NumField(); i++ {
+		f := schema.Field(i)
+		if !f.IsExported() {
+			continue
+		}
+		tag, ok := f.Tag.Lookup("mapstructure")
+		if ok {
+			name, opts, _ := strings.Cut(tag, ",")
+			if name == "-" {
+				continue
+			}
+			if name == "" && strings.Contains(opts, "squash") && f.Anonymous {
+				for k, nested := range mapstructureFields(f.Type) {
+					out[k] = nested
+				}
+				continue
+			}
+			if name != "" {
+				out[name] = f
+				continue
+			}
+		}
+		if f.Anonymous {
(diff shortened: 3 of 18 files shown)


internal/cli/cmd/validate.go
```diff
@@ -102,8 +104,18 @@ Examples:
 				}
 			}
 
-			cfg, err := config.Load(configPath)
+			// Strict mode: reject unknown keys in the base file and in the
+			// production override (when discovered) so typos like tls.dmain fail
+			// loudly at validate time instead of being silently dropped at merge
+			// time. The runtime loader (config.Load) stays lenient per ADR-065.
+			prodPath := discoverProdOverride(configPath)
+			cfg, err := config.LoadStrict(configPath, prodPath)
```
tibtof, line 112: **Missing CLI-level test for the strict path** (AC: `vibew validate parity`).

The architect's design and the PM acceptance criteria explicitly call out:
> `vibew validate` run against a prod-override file containing an unknown key emits the same actionable error as `vibew deploy` would

The library-level coverage in `internal/config/strict_test.go` is solid, but the CLI wiring is untested. Three gaps:

1. No test that `vibew validate <base.yaml>` with `tls.dmain: foo` in the base file exits non-zero and prints `Configuration invalid: config …: unknown key(s): tls.dmain`.
2. No test that `vibew validate` discovers the sibling `vibewarden.production.yaml` and rejects unknown keys found there.
3. No test for `discoverProdOverride` (the new auto-discovery helper, which is extra behavior beyond what the architect specified — deserves its own unit test).

`internal/cli/cmd/validate_test.go` already has the harness pattern (`TestValidateCmd_ValidConfig`). Adding a `TestValidateCmd_RejectsUnknownKeys` and a `TestValidateCmd_RejectsUnknownKeysInProdOverride` is trivial and prevents the breaking-change behavior from regressing silently.

Fix: add the two table-driven CLI tests + a small unit test for `discoverProdOverride`.

internal/cli/cmd/validate.go
```diff
@@ -136,6 +148,23 @@ Examples:
 	return cmd
 }
 
+// discoverProdOverride returns the path to vibewarden.production.yaml that
+// sits next to configPath, when it exists. The empty string is returned when
+// configPath is empty, the override file does not exist, or the directory
+// cannot be read. Strict validation still succeeds when no override is
+// present.
+func discoverProdOverride(configPath string) string {
```
tibtof, line 156: **Scope creep — `discoverProdOverride` was not in the architect design.**

The architect's design said:
> `vibew validate` — switches `config.Load(configPath)` → `config.LoadStrict(configPath, "")`. When a prod-override path is provided (future flag) pass it through.

Here you go further: you auto-discover `vibewarden.production.yaml` next to the base config. That is a good UX decision (otherwise `vibew validate` would never catch prod-file typos until a flag exists, which defeats the loudness policy for the common case), but it is a behavioural addition the architect didn't mandate.

No change requested — just calling it out as intentional scope expansion. Please confirm this was a deliberate call rather than accidental over-engineering, and add a unit test for `discoverProdOverride` (nil path, missing file, present file, unreadable dir).

internal/config/strict.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// walkUnknownKeys recursively compares the YAML tree against the schema type.
+// Any key in tree that has no corresponding mapstructure tag on schema (or any
+// nested struct in schema) is appended to out, prefixed by the dotted path
+// that leads to it.
+//
+// Keys whose schema field has mapstructure tag "-" (e.g. DeployMode,
+// ProjectRoot) are treated as unknown — they are never expected to appear in
+// YAML.
```
tibtof, line 120: **Coverage gap in `walkUnknownKeys` for slice-of-struct fields.**

`walkUnknownKeys` recurses into `map[string]any` children but does not recurse into `[]any` children, so typos inside slice elements are silently accepted. Example: a typo in a single entry of `auth.scope_rules`, `social_providers`, `input_validation.path_overrides`, or `body_size.overrides` won't be caught by `vibew validate`.

This is not a blocker — the architect's AC is "unknown keys in the base or the prod file" and naming a key like `auth.scope_rules` exists. But a vibe coder who writes:
```yaml
auth:
  scope_rules:
    - paht: /api  # typo: should be `path`
      methods: [GET]
```
will still silently lose that rule in production. That is exactly the failure mode #1053 was filed for.

Options:
1. Document the limitation in the `walkUnknownKeys` godoc (currently says "recursively compares" without caveat) — acceptable for this PR, track slice-element coverage as a follow-up.
2. Recurse into slice entries when the schema field is `[]SomeStruct` and each YAML entry is a `map[string]any` — straightforward given the reflection already in place.

I'd accept option 1 if you open a follow-up issue; option 2 is the correct long-term fix.

internal/app/deploy/service.go
```diff
@@ -345,9 +345,17 @@ func (s *Service) Deploy(ctx context.Context, cfg *config.Config, opts RunOption
...
 	// of DNS propagation, external port availability, and TLS certificate issuance.
 	// Use the merged config (base + prod overlay) for the correct port and TLS state.
-	healthCfg, err := LoadMergedConfig(cfg, opts.ProdConfigPath)
-	if err != nil {
-		return err
+	// When opts.ConfigPath is empty (programmatic callers that build cfg in-memory)
+	// or when the file is not on disk (unit tests), fall back to the supplied cfg.
+	healthCfg := cfg
+	if opts.ConfigPath != "" {
+		if _, statErr := os.Stat(opts.ConfigPath); statErr == nil {
```
tibtof: **Is the `os.Stat` guard really the right abstraction?**

The dev agent flagged this themselves. The pattern is duplicated here and in `service.go:351`:
```go
mergedCfg := cfg
if configPath != "" {
    if _, statErr := os.Stat(configPath); statErr == nil {
        loaded, err := LoadMergedConfig(configPath, prodConfigPath)
        ...
    }
}
```

Two smells:
1. Silent fall-through on stat error: if `configPath` is set but `os.Stat` fails for a reason other than `IsNotExist` (e.g. permission denied), we silently use the in-memory `cfg` and pretend nothing happened. A permission-denied base file in production would produce a misleading bundle with no warning. At minimum, only fall back on `errors.Is(err, fs.ErrNotExist)`.
2. `LoadMergedConfig` itself already handles the empty-prodConfigPath case. It could also handle the "configPath is in-memory / missing" case by accepting baseYAML bytes directly, which would let callers drop the stat dance.

Minimal fix for this PR: replace `statErr == nil` with `errors.Is(statErr, fs.ErrNotExist)`-negated, so real errors surface.

Cleaner follow-up (not this PR): split `LoadMergedConfig` into `LoadMergedConfigFromFiles(base, prod string)` and `MergeConfigWithOverride(cfg *Config, prodPath string)` so the "in-memory cfg, prod file on disk" case has a first-class API.

tibtof:
## Reviewer Agent — changes requested

(Posted as a comment because GitHub blocks a review on own PR; treat as `request-changes` for the pipeline.)

Very close to approval. Core fix is correct: routing the prod-override through `MergeConfigYAML` kills the allow-list bug cleanly. ADR-082 is well-written, library-level tests for `LoadStrict` / `UnknownKeyError` are solid, the CHANGELOG has both a `Breaking` and a `Fixes` entry with a concrete migration hint, ADR-065's `ErrorUnused=false` runtime invariant is preserved (documentary test pins it), env-var precedence is covered by `TestLoadMergedConfig_EnvVarWinsOverProd`, no new dependencies, no license concerns, build + lint + `go test ./...` all green locally.

### Why changes requested

1. **No CLI-level test for the breaking-change path.** `vibew validate` now rejects unknown keys and auto-discovers the prod override — user-facing, breaking. Not a single test in `internal/cli/cmd/validate_test.go` asserts that `vibew validate <config-with-typo>` exits non-zero with the `Configuration invalid: …` message, nor that prod-override typos are caught via auto-discovery. The PM AC `vibew validate parity` is only covered at library level. Add table-driven CLI tests — the harness in `TestValidateCmd_ValidConfig` is already the right shape. Same gap for `vibew deploy`'s new strict pre-check.

2. **`discoverProdOverride` is scope creep with no unit test.** The architect said *"When a prod-override path is provided (future flag) pass it through."* The dev went further and auto-discovers the sibling file. I agree it's the right UX call, but an undocumented scope expansion with zero direct test coverage should either get the tests or be removed.

3. **`os.Stat` fallback silently hides non-`IsNotExist` errors** (duplicated in `bundle.go` and `service.go`). If `configPath` is set but stat fails on permission-denied, the code silently reverts to the in-memory cfg — exactly the class of "silent config drop" bug #1053 was filed for. At minimum, gate the fallback on `errors.Is(err, fs.ErrNotExist)`.

### Non-blocking observation

`walkUnknownKeys` does not recurse into slice-of-struct entries (e.g. `auth.scope_rules`, `input_validation.path_overrides`). A typo inside a slice entry will still silently load. Acceptable if you either (a) document the limitation in the godoc and open a follow-up issue, or (b) extend the walk.

### Verification performed

- `git fetch origin && gh pr checkout 1056` — clean.
- `go build ./...` — clean.
- `go test ./...` — all packages pass.
- `golangci-lint run ./internal/config/... ./internal/app/deploy/... ./internal/cli/...` — 0 issues.
- `.claude/worktrees/` is untracked local noise — not in the diff; `make check` trips on it only because `gofmt` walks the tree. Not a PR concern.
- License surface: no `go.mod` changes.
- Security: tempfile in `os.TempDir()` + `defer os.Remove` is correct. Error messages name local paths the user supplied — no sensitive info leaked.

Inline comments posted on the four hotspots.

tibtof:
## Documentation review — changes requested (writer agent)

Cannot submit a formal `--request-changes` review because this PR is authored by the same account as the reviewer; posting findings as a comment and setting `status:changes-requested`.

Ran the new strict loader against every user-facing YAML snippet and example. One blocking finding: the init/wrap templates themselves ship a key that trips strict mode.

### 1. BLOCKING: init/wrap templates reference `overrides.docker_compose` (does not exist in schema)

Files:
- `internal/cli/templates/init-vibewarden.yaml.tmpl:57`
- `internal/cli/templates/vibewarden.yaml.tmpl:106`

Both templates include this commented example block:

```yaml
# overrides:
#   kratos_config: ""      # path to a custom kratos.yml
#   docker_compose: ""     # path to a custom docker-compose.yml
```

The actual schema field (per `internal/config/config.go:1600-1602` and `docs/configuration.md:706`) is `overrides.compose_file`, not `overrides.docker_compose`. Confirmed the regression with the new strict loader on this branch:

```
$ vibew validate --config vibewarden.yaml
Configuration invalid: config vibewarden.yaml: unknown key(s): overrides.docker_compose
Error: loading config: ...
```

A fresh user who uncomments the commented-out example in the file that `vibew init` / `vibew wrap` just wrote for them will have their very next `vibew validate` or `vibew deploy` fail. This is exactly the docs-runtime drift class ADR-082 exists to prevent — shipping it in the same PR that activates strict mode is self-defeating.

**Fix**: rename `docker_compose` -> `compose_file` in both templates. Same change in both files. While there, consider a golden-style test that runs `LoadStrict` against the fully-rendered template output so future drift is caught in CI.

### 2. Minor (not blocking): no troubleshooting entry for `unknown key(s)`

`docs/troubleshooting.md` has no entry for the new error. The CHANGELOG Breaking entry is good, but a user whose `vibew deploy` failed with `unknown key(s): tls.dmain` will grep the docs for "unknown key" and find nothing.

**Suggestion**: add a short section under `docs/troubleshooting.md` with the exact error text, an explanation (typo or removed key), and the fix (consult `docs/configuration.md` or `vibewarden.reference.yaml` for the correct key). Fine as a follow-up.

### 3. Pre-existing, not blocking: `plugins.fleet` mention in llms-full.txt

`llms-full.txt:263` references `plugins.fleet` as a top-level key, but there is no `plugins` top-level in `internal/config/config.go`. The text is prose (not a YAML snippet), so it will not trip strict mode and is out of scope for this PR. Flagging for a separate issue.

### Verified as correct

All other YAML snippets parse cleanly against the schema: `README.md:218-228`, `docs/getting-started.md` (all snippets), `docs/configuration.md` (all 12 snippets including the full example at line 713), `docs/deploy-to-vps.md:140`, `docs/troubleshooting.md:173,383`, `docs/examples/AGENTS-VIBEWARDEN.md`, `vibewarden.example.yaml`, `vibewarden.reference.yaml`, `examples/demo-app/vibewarden.yaml`, `examples/demo-app/vibewarden.production.yaml`, `examples/vibewarden.prod.yaml`, `examples/{nextjs,node-express,python-flask,spring-boot}/vibewarden.yaml`.

CHANGELOG Breaking entry is clear, names the ADR, gives a concrete example, and includes the migration hint.

Please fix finding #1 and I'll re-review.

tibtof:
## Changes pushed — ready for re-review

Addressed all four findings from the reviewer and writer agents. `make check` (golangci-lint + race tests) passes cleanly.

| # | Finding | Fix | Commit |
|---|---------|-----|--------|
| 1 | No CLI-level tests for `Configuration invalid: …` path | Added `TestValidateCmd_RejectsUnknownKeys` (base-file typos: unknown top-level key, `tls.dmain`), `TestValidateCmd_RejectsUnknownKeysInProdOverride` (auto-discovered sibling `vibewarden.production.yaml`), and `TestValidateCmd_AcceptsValidProdOverride` (happy path) in `internal/cli/cmd/validate_test.go`. All three exercise the real cobra command via `root.Execute()`. | `2635982` |
| 2 | `discoverProdOverride` has no unit tests | Added `internal/cli/cmd/validate_internal_test.go` with a table-driven `TestDiscoverProdOverride` covering: empty input, missing base config, present base without sibling, happy path, sibling-is-directory edge case, plus a separate `TestDiscoverProdOverride_UnreadableDir` for the permission-error silent-swallow path. | `2635982` |
| 3 | `os.Stat` fallback silently hides non-`IsNotExist` errors in `bundle.go` and `service.go` | Replaced both `if statErr == nil { … } else { fallback }` blocks with an explicit `switch` on `errors.Is(statErr, fs.ErrNotExist)`. Non-`ErrNotExist` errors now wrap to `stat config %s: %w`. Added `TestBundle_SingleSite_ConfigPathStatError` that creates a config inside an unreadable directory and asserts the error surfaces with the new message. | `2635982` |
| 4 | **BLOCKING** — templates reference schema-less `overrides.docker_compose` | Renamed `docker_compose` → `compose_file` in both `internal/cli/templates/init-vibewarden.yaml.tmpl:57` and `internal/cli/templates/vibewarden.yaml.tmpl:106`. Added `internal/cli/templates/yaml_templates_strict_test.go` — a golden-style test that renders each template, uncomments the escape-hatch example block, and runs `config.LoadStrict` against the output. Any future template key that drifts from the schema will fail the test. | `5cb3537` |
| 5 (follow-up) | No troubleshooting entry for `unknown key(s)` | Added a section to `docs/troubleshooting.md` naming the error, explaining the cause, and pointing at `internal/config/config.go` as the source of truth. Restates ADR-065 runtime leniency. | `d7111c8` |
| 6 (lint cleanup) | `gosec G302` on test cleanup `os.Chmod(dir, 0o700)` | Annotated the two cleanup closures with `//nolint:gosec` and a comment explaining directory traversal requires the execute bit. | `48916d3` |

### Out-of-scope items deferred
- `llms-full.txt:263` stale `plugins.fleet` reference — writer flagged as out of scope (prose, not YAML, won't trip strict mode). Not opening a separate issue in this PR to keep scope tight; happy to file one on request.
- `walkUnknownKeys` non-recursion into slice-of-struct entries — reviewer flagged as non-blocking observation. The godoc already documents the limitation (`internal/config/strict.go:152` comment). Leaving the deeper fix for a follow-up issue.

### Verification
- `gofmt -l cmd internal migrations` — clean.
- `golangci-lint run ./...` — 0 issues.
- `go build ./...` — clean.
- `go test -race ./...` — all packages pass, including the new tests.
- `make check` — passes after temporarily moving untracked `.claude/worktrees/` (pre-existing local noise flagged by the reviewer, not in the diff).

Removing `status:changes-requested` and adding `status:ready-for-review`.

tibtof:
LGTM / approved (reviewer agent re-review after commits 2635982, 5cb3537, d7111c8, 48916d3).

All three prior findings are resolved:

1. **CLI-level tests for strict-path** — `internal/cli/cmd/validate_test.go` now drives the cobra command end-to-end via `root.Execute()`, asserts non-nil error, and checks stderr contains `Configuration invalid:`, `unknown key(s):`, the typo key (`bogus_section`, `tls.dmain`), and — in `TestValidateCmd_RejectsUnknownKeysInProdOverride` — the override filename. Happy-path prod-override coverage added too.
2. **`discoverProdOverride` unit tests** — `validate_internal_test.go::TestDiscoverProdOverride` covers empty input, missing base, base-without-sibling, base-with-sibling, sibling-is-a-directory, plus a separate `TestDiscoverProdOverride_UnreadableDir` for the permission-denied branch (skipped as root). All 5 branches exercised.
3. **`bundle.go` / `service.go` stat handling** — both dispatch sites now switch on `statErr == nil` / `errors.Is(statErr, fs.ErrNotExist)` / `default` with `fmt.Errorf("stat config %s: %w", …)` wrapping. `TestBundle_SingleSite_ConfigPathStatError` is a real permission-denied test (chmod 0o000 parent dir with a real file inside), asserts error returned and `"stat config"` in the message — not a flag-flipped happy path.

Verified locally: `go test ./internal/cli/cmd/... ./internal/app/deploy/... ./internal/app/generate/... ./internal/config/...` passes, `gofmt -l internal/ cmd/` clean, `go vet ./internal/... ./cmd/...` clean. (`make check` flagged a gofmt issue in `.claude/worktrees/agent-a05f512d/…` which is local worktree cruft, not PR content — ignored.)

No new findings or regressions.

tibtof:
**Writer agent re-review: LGTM**

All three prior findings resolved:

1. **BLOCKING (`docker_compose` → `compose_file`)** — **Resolved.** Verified both `internal/cli/templates/init-vibewarden.yaml.tmpl:57` and `internal/cli/templates/vibewarden.yaml.tmpl:106` now emit `compose_file`. The mapstructure tag on `OverridesConfig.ComposeFile` in `internal/config/config.go:1602` is the source of truth and matches. The new golden test `internal/cli/templates/yaml_templates_strict_test.go` renders each template with representative `InitProjectData` / `TemplateData`, uncomments the `overrides:` escape-hatch block, and invokes `config.LoadStrict` against the result. All 4 subtests pass locally (`as-rendered` and `escape-hatches-uncommented` for both templates). Wired into `make check` via the `go test -race ./...` step.

2. **Minor (troubleshooting entry)** — **Resolved.** `docs/troubleshooting.md:284–310` is searchable (literal "unknown key(s)" + the full example error string a user would grep for), actionable (points at `internal/config/config.go` via GitHub link plus `docs/configuration.md` and `vibewarden.reference.yaml` as mirrors), and agent-readable (matches the Symptom / Cause / Fix structure of sibling entries). The `docker_compose` → `compose_file` rename is explicitly called out as an example cause.

3. **Pre-existing (`llms-full.txt:263` `plugins.fleet`)** — **Out of scope as previously agreed.** Verified `llms-full.txt` contains no `docker_compose`/`overrides.compose_file` references at all, so nothing new regressed.

Spot-checks on other YAML snippets (non-worktree):
- `docs/configuration.md:706` already documents `overrides.compose_file` correctly.
- `examples/demo-app/vibewarden.yaml:121–124` uses valid keys (`kratos_config`, `identity_schema`).
- All `docs/examples/*.md` `overrides:` blocks are `body_size.overrides` (different field — path/max list), not the escape-hatch. Not affected.
- `examples/*/.vibewarden/generated/docker-compose.yml` headers reference `overrides.compose_file` correctly.

CHANGELOG breaking-change entry (lines 15–25) is migration-friendly: names a concrete typo example (`tls.dmain`), references ADR-065 for runtime forward-compat, and provides the scratch-annotation → YAML-comment migration tip. The `docker_compose` rename case is adequately covered in the troubleshooting entry rather than the CHANGELOG, which is appropriate since `docker_compose` was only ever a scaffold typo and never a documented public key.

Labels on the PR and issue already show `status:APPROVED`; no label change needed from me.

Issue #1059, opened 2026-04-20
chore: remove production-only doctor checks after deploy sunset (#1051)


tibtof:
## Context

With `vibew deploy` sunset merged (#1051 / #1064), the "Production" section of
`vibew doctor` is dead code. It only ran when the user supplied an `ssh://`
target, and that UX path no longer exists. ADR-084 explicitly deferred this
deletion to the deploy-sunset follow-up; no new architectural decision is
required.

This story performs that deletion and the adjacent cleanup — nothing more.

## User story

As a vibe coder, I want `vibew doctor` to only describe checks that actually
run so that I am not misled by help text advertising a remote-target workflow
that the CLI no longer supports.

## Scope

**Delete the five production-only doctor checks and their direct supporting code.**

Confirmed inventory (from `grep -rn "checkSSHConnectivity\|checkArchCompatibility\|checkRemoteContainerHealth\|checkDomainDNS\|checkRemoteTLSCert" internal/`):

| File | Lines to cut (approx) |
|---|---|
| `internal/app/ops/doctor.go` | five check funcs + the Layer 3 block in `runChecks` + `Target` field on `DoctorOptions` + `WithRemoteExecutor` + `remoteExecutor` field + `extractHostFromTarget` + `remoteTLSCertExpiryWarnDays` + `normalizeArch` shim + unused imports (`net/url`, `runtime`, possibly `archutil`) | ~270 LOC |
| `internal/app/ops/doctor_test.go` | `fakeRemoteExecutor` + every test using `opts.Target = "ssh://..."` (9 call-sites at lines 826, 854, 922, 969, 1009, 1061, 1325, 1367, 1405 and their surrounding cases) | ~600 LOC |
| `internal/app/ops/doctor_internal_test.go` | `TestExtractHostFromTarget`, `TestNormalizeArch` | ~120 LOC |
| `internal/cli/cmd/doctor.go` | `--target` and `--ssh-key` flags, SSH executor wiring, Layer-3 "Production" block in the `Long` help text, `sshadapter` import | ~40 LOC |

Rough total: ~1000 LOC removed.

**Also reassess for orphaning after the cuts above:**

[...]

Issue #1051, opened 2026-04-20
chore: sunset vibew deploy — replaced by vibew bundle


tibtof:
## Context

`vibew deploy` has been the #1 source of friction across 4 retros (the v0.15.0 retro, documented in this issue's sibling bug reports, makes 17+ bugs total). The feature has been consistently slower than running `docker save | ssh docker load + rsync + docker compose up -d` by hand. The 4 manual commands work; the wrapper adds failure modes without adding value commensurate with the complexity.

## Decision

Drop `vibew deploy` entirely. Users deploy the artifacts produced by `vibew bundle` (#1044) using standard Docker / SSH / compose commands. The reference is a 10-line `deploy.sh` shipped inside the bundle.

## Sequencing — do not violate

1. **#1044 (vibew bundle, compose-only) must ship first.** The current \`vibew deploy --dry-run\` generator is the only way to materialise a production bundle; its logic must be lifted into \`vibew bundle\` before deploy is deleted.
2. **Then, and only then**, this issue deletes:
   - \`cmd/vibewarden/deploy.go\` (+ subcommands: status, logs, etc.)
   - Anything in \`internal/\` that's only used by deploy (check with \`grep\`/\`go build\`)
   - All deploy docs, AGENTS fallback section, and example references
3. Retain the "manual deploy" docs as the canonical path — or better, point at the \`deploy.sh\` inside the bundle output.

## Scope of deletion

- [ ] \`cmd/vibewarden/deploy*.go\` and subcommands (status/logs/validate/etc.)
- [ ] \`internal/deploy/\` orchestration (keep generator logic — move to \`internal/bundle/\` as part of #1044)
- [ ] Root-level \`vibew deploy\` help text, completions
- [ ] Docs: \`docs/deploy.md\`, \`docs/deploy-troubleshooting.md\`, \`docs/guide/vps-deploy.md\`, AGENTS \"Manual fallback\" section (re-point to bundle)
- [ ] README references
- [ ] Example configs referencing deploy
- [ ] Tests for deploy command
- [ ] \`vibew doctor\` checks that only make sense during deploy

## Bugs this closes as won't-fix-because-deleted

- #1048 — deploy auto-builds for target arch
- #1006 — deploy --reset-db
- The v0.15.0 retro bugs: arch-check-wrong, --dry-run clobbers .env, deploy error "fix" is the command already run

## Bugs this does NOT close (survive deploy removal)

- Config merge silently drops unknown keys (file separately)
[...]

PR #1060, opened 2026-04-20
fix(#1054): doctor port/TLS/remote checks — no more false positives, no shell leaks

### LINKED ISSUE #1054: fix: vibew doctor false-positives and raw-shell error leaks
## Problems

\`vibew doctor\` produced three misleading outputs in the v0.15.0 fresh-agent retro:

### 1. Port 8443 false-positive when \`vibew dev\` is running

\`\`\`
Proxy port: port 8443 is already in use
\`\`\`

This is normal — the local dev sidecar binds 8443. Doctor should detect that the process owning the port is a sibling \`vibew dev\` and either skip the check or report it as \"in use by vibew dev (expected)\".

### 2. TLS cert false-positive when cert exists

\`\`\`
TLS certificate .vibewarden/generated/certs/server.crt not found — run 'vibew generate' to create certs
\`\`\`

Reported even when the dev sidecar is currently serving requests with a valid cert. Either the path is wrong, or the check is running against a pre-generate state it shouldn't be.

### 3. Raw shell pipe leaked into error

\`\`\`
Remote containers: could not query remote containers:
  ssh docker compose ps --format json 2>/dev/null || docker-compose ps 2>/dev/null: exit status 127
\`\`\`

Exit 127 = command not found, but Docker *is* installed on the remote. The raw pipe-with-\`2>/dev/null\` masks the real error (probably the remote \`ssh\` call itself failed, not docker). Two problems:
- Raw shell pipes must never appear in user-facing error messages
- \`2>/dev/null\` inside the command hides the diagnostic the doctor needs

## Acceptance criteria

- [ ] Port-check: if port is bound by a local \`vibew dev\` process, report as informational not error
- [ ] TLS-cert check: match whatever path \`vibew dev\` actually writes to; re-verify detection logic end-to-end with a fresh \`vibew init + vibew dev\` flow
- [ ] Remote-containers check: no raw shell pipes in output; remove \`2>/dev/null\` so we get the actual exit code+stderr; format errors as human messages with a \"what to do next\" hint
- [ ] Doctor integration test: runs against a live \`vibew dev\` and produces zero false-positives
- [ ] Every check emits a single-line summary and (when failing) a one-line \"how to fix\" — no stack traces, no shell pipes

## Found in

v0.15.0 fresh-agent retrospective, qr-andy-warhol demo deploy (2026-04-20)

## Related

This issue is NOT obsoleted by #1051 (sunset deploy). Doctor will remain; the remote-container check may be removed as part of sunset, but the local checks survive.

tibtof:
Fixes #1054 (partial). Follow-up #1059 tracks removing production-only doctor checks after deploy sunset (#1051).

## Summary

Implements ADR-084. Three long-standing doctor false-positives fixed at
the structural level — no string-munging:

- **Port ownership** (`bug 1`): new `ports.PortOwnerProbe` + `VibeWardenHealthProbe` adapter. `checkPort` now issues a TLS-protected GET to `/_vibewarden/health` when the port is busy; a sibling `vibew dev` is reported OK ("in use by local vibew dev (expected)") instead of FAIL.
- **TLS certificate** (`bug 2`): deletes the `.vibewarden/generated/certs/server.crt` filesystem scan. `checkTLSCertValid` performs a live TLS handshake against `proxyHost:proxyPort` and reads the leaf cert from `ConnectionState().PeerCertificates[0]` — the expiry/severity logic now reflects what the sidecar is actually serving, not what may or may not live in a directory doctor has no business knowing about. Unreachable sidecar → WARN "start `vibew dev`".
- **Remote container check** (`bug 3`): new `internal/app/ops/remote_error.go` — pure `formatRemoteError` returns a single line with exit code + first stderr line + keyed hint, and never emits `2>/dev/null`, `||`, raw ssh commands, or the adapter's `ssh exit:` wrapper. The remote command is simplified to `docker compose ps --format json` (Compose v1 fallback + shell redirections dropped).
- **SSH adapter micro-fix**: `Executor.Run` now wraps errors as `"ssh exit: %w"` — the raw `ssh <cmd>:` and `output: ...` fragments that were the root cause of the shell leak are gone structurally.

3 topical commits: port-ownership port+adapter, remote-error formatter + ssh micro-fix, doctor service rewrite.

## Test plan

- [x] `make check` passes (lint + vet + build + race tests on main module and demo-app)
- [x] Unit — `VibeWardenHealthProbe`: signature match → OwnerVibeWarden; foreign 200 / non-2xx / plain-TCP / port-closed → OwnerForeign; `0.0.0.0` rewritten to `127.0.0.1`; strict prefix match (non-prefix signature → foreign)
- [x] Unit — `formatRemoteError` table: nil/empty/exit 127/exit 126/exit 1 multi-line/exit 1 with shell leaks/exit 255/plain error/ssh-exit-wrapped error; every case asserts single line, ≤ 180 chars, no `2>/dev/null`, no `||`, no `ssh exit`
- [x] Unit — `checkPort` ownership matrix: port free (probe not consulted) → OK; bound + `OwnerVibeWarden` → OK; bound + `OwnerForeign` → FAIL; bound + no probe → FAIL (back-compat)
- [x] Unit — `checkTLSCertValid` against `httptest.NewTLSServer` with controlled cert `NotAfter`: valid (90d) → OK; near-expiry (3d) → WARN; expired → FAIL; unreachable (closed port) → WARN with hint
- [x] Regression — `Remote containers` detail does not contain `2>/dev/null`, `||`, or `ssh exit` even when the executor returns an error string packed with all three

## Notes

- No new dependencies (stdlib only: `crypto/tls`, `net/http`, `encoding/json`, `os/exec`).
- `NewDoctorService` signature unchanged; probe is wired via `WithPortOwnerProbe(...)` setter for back-compat with existing tests.
- Integration test (build-tagged) deferred — not in scope for this PR; tracked implicitly by #1059 which will own the full production-check sunset.

### DIFF
--- a/internal/app/ops/doctor.go
+++ b/internal/app/ops/doctor.go
@@ -531,10 +531,15 @@ func (s *DoctorService) checkRemoteContainerHealth(ctx context.Context) CheckRes
 	// of ADR-084 so stderr surfaces naturally for the error formatter.
 	output, err := s.remoteExecutor.Run(checkCtx, "docker compose ps --format json")
 	if err != nil {
+		// ssh.Executor.Run merges stdout+stderr into `output` and never
+		// populates *exec.ExitError.Stderr, so we must forward `output`
+		// to the formatter — otherwise the real stderr line (e.g.
+		// "docker: command not found") is discarded and the user only
+		// sees the canned exit-code hint.
 		return CheckResult{
 			Name:     "Remote containers",
 			Severity: SeverityFail,
-			Detail:   formatRemoteError(err, "docker compose"),
+			Detail:   formatRemoteError(err, output, "docker compose"),
 		}
 	}
 
--- a/internal/app/ops/doctor.go
+++ b/internal/app/ops/doctor.go
@@ -3,9 +3,7 @@ package ops
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"encoding/json"
-	"encoding/pem"
 	"fmt"
 	"io"
 	"net"
@@ -59,6 +57,7 @@ type DoctorService struct {
 	healthChecker  ports.HealthChecker
 	remoteExecutor ports.RemoteExecutor
 	imageChecker   ports.DockerImageChecker
+	ownerProbe     ports.PortOwnerProbe
 }
 
 // NewDoctorService creates a new DoctorService.
@@ -85,6 +84,17 @@ func (s *DoctorService) WithRemoteExecutor(executor ports.RemoteExecutor) *Docto
 func (s *DoctorService) WithImageChecker(checker ports.DockerImageChecker) *DoctorService {
 	cp := *s
 	cp.imageChecker = checker
+	return &cp
+}
+
+// WithPortOwnerProbe returns a copy of the DoctorService with the given
+// PortOwnerProbe wired for port-ownership detection. When nil (or no probe
+// is supplied) the proxy-port check falls back to "busy = FAIL" semantics —
+// this preserves back-compat with existing tests that do not set a probe.
+// See ADR-084.
+func (s *DoctorService) WithPortOwnerProbe(probe ports.PortOwnerProbe) *DoctorService {
+	cp := *s
+	cp.ownerProbe = probe
 	return &cp
 }
 
@@ -177,7 +187,7 @@ func (s *DoctorService) runChecks(ctx context.Context, cfg *config.Config, opts 
 
 	// --- Layer 2: Local Runtime ---
 	results = append(results, withSection(s.checkUpstreamReachable(ctx, cfg), sectionLocalRuntime))
-	results = append(results, withSection(checkTLSCertValid(cfg, workDir), sectionLocalRuntime))
+	results = append(results, withSection(checkTLSCertValid(ctx, cfg, proxyHost, proxyPort), sectionLocalRuntime))
 
 	// --- Layer 3: Production (only when target is set and executor is available) ---
 	if opts.Target != "" && s.remoteExecutor != nil {
@@ -257,6 +267,10 @@ func checkConfigFile(cfg *config.Config, configPath string) CheckResult {
 	}
 }
 
+// checkPort combines the TCP-bind probe (PortChecker) with the optional
+// ownership probe (PortOwnerProbe) to distinguish the expected "sibling
+// vibew dev is running" state from a real foreign-process conflict. See
+// ADR-084 for the decision table.
 func (s *DoctorService) checkPort(ctx context.Context, label, host string, port int) CheckResult {
 	checkCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
@@ -269,17 +283,29 @@ func (s *DoctorService) checkPort(ctx context.Context, label, host string, port 
 			Detail:   fmt.Sprintf("port %d check failed: %v", port, err),
 		}
 	}
-	if !available {
+	if available {
 		return CheckResult{
 			Name:     label,
-			Severity: SeverityFail,
-			Detail:   fmt.Sprintf("port %d is already in use", port),
+			Severity: SeverityOK,
+			Detail:   fmt.Sprintf("port %d is available", port),
 		}
 	}
+
+	// Port is bound. Ask the owner probe who is there.
+	if s.ownerProbe != nil {
+		if s.ownerProbe.ProbeOwner(checkCtx, host, port) == ports.OwnerVibeWarden {
+			return CheckResult{
+				Name:     label,
+				Severity: SeverityOK,
+				Detail:   fmt.Sprintf("port %d in use by local vibew dev (expected)", port),
+			}
+		}
+	}
+
 	return CheckResult{
 		Name:     label,
-		Severity: SeverityOK,
-		Detail:   fmt.Sprintf("port %d is available", port),
+		Severity: SeverityFail,
+		Detail:   fmt.Sprintf("port %d is already in use", port),
 	}
 }
 
@@ -383,10 +409,16 @@ func (s *DoctorService) checkUpstreamReachable(ctx context.Context, cfg *config.
 	}
 }
 
-// checkTLSCertValid checks the local self-signed TLS certificate for expiry.
-// It only runs when the TLS provider is "self-signed" and cert files exist in
-// the generated certs directory.
-func checkTLSCertValid(cfg *config.Config, workDir string) CheckResult {
+// checkTLSCertValid inspects the local self-signed TLS certificate by doing
+// a live TLS handshake against the running sidecar. When the sidecar is not
+// reachable the check degrades to WARN with a "start vibew dev" hint.
+//
+// Self-signed cert storage is owned by the embedded Caddy instance
+// (internal/adapters/caddy) and the only reliable way to see which cert the
+// sidecar is actually serving is to handshake with it — which is exactly
+// what we do here. See ADR-084 for the rationale and the rejected
+// filesystem-scan approach.
+func checkTLSCertValid(ctx context.Context, cfg *config.Config, host string, port int) CheckResult {
 	if cfg.TLS.Provider != "self-signed" {
 		return CheckResult{
 			Name:     "TLS certificate",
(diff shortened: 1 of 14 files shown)


tibtof:
## Writer review: changes requested

Three user-visible behavior changes ship in this PR, but the user-facing docs do not reflect them.

### 1. CHANGELOG.md [Unreleased] — missing entries

No bullets were added under the existing `### Fixes` section. Users reading release notes have no way to learn that doctor output has changed. Add three bullets:

```markdown
- **`vibew doctor` no longer flags a running sidecar as a port conflict** (#1054, ADR-084). When `vibew dev` is already running on the proxy port, doctor probes `/_vibewarden/health` and reports `[OK] in use by local vibew dev (expected)` instead of `[FAIL] port 8443 is already in use`. Foreign processes still FAIL as before.
- **`vibew doctor` TLS cert check now inspects the live sidecar** (#1054, ADR-084). The self-signed cert check performs a TLS handshake against the sidecar instead of reading a hardcoded `server.crt` path, so it no longer reports "certificate file not found" when the cert exists in Caddy-managed storage. If the sidecar is not reachable, the check is now WARN with "start 'vibew dev'".
- **Remote-container doctor errors no longer leak raw shell fragments** (#1054, ADR-084). `checkRemoteContainerHealth` previously surfaced errors containing literal `2>/dev/null` and `|| docker-compose ps`. Errors are now a single clean line with an exit-code-keyed hint (e.g. `exit 127: docker: command not found — docker compose not installed on remote`).
```

### 2. docs/troubleshooting.md — stale sample output and check description

- **Line 108 sample** ("Port conflict + Docker not running") shows `[FAIL] Proxy port port 8443 is already in use`. After this PR that exact line only fires for a **foreign** owner. Either add a companion sample showing `[OK] Proxy port in use by local vibew dev (expected)` or note explicitly that the FAIL line here refers to a non-VibeWarden process.
- **Lines 49-50** describe check #10 "TLS cert valid" as inspecting the "Local TLS certificate (when using self-signed)". That wording implies a file-on-disk check. Update to describe the live TLS handshake behaviour from ADR-084: the doctor connects to the sidecar, reads the leaf cert from the handshake, and reports WARN + "start 'vibew dev'" when the handshake fails.
- **Lines 157-190** ("Port conflict — proxy port already in use") is still correct for the foreign-process case but should mention up front that this FAIL no longer fires for a running `vibew dev`.

### 3. Other surfaces — OK as-is

- `llms-full.txt`, `AGENTS-VIBEWARDEN.md` template, `docs/getting-started.md`, `docs/index.md` — only contain generic "run `vibew doctor`" references, no stale error strings. No change needed.
- ADR-084 is correctly listed in `decisions/README.md` line 58.
- #1059 follow-up is tracked correctly as internal chore work; no public doc link needed.

### Verdict

Request changes — please add the CHANGELOG bullets and update `docs/troubleshooting.md` (sample output + check description) so the documented doctor output matches the new runtime behaviour. Approving once those land.

internal/app/ops/doctor.go
```diff
@@ -470,17 +519,22 @@ func (s *DoctorService) checkSSHConnectivity(ctx context.Context) CheckResult {
...
+	// Docker Compose v1 reached end-of-life in June 2023; the fallback
+	// ("|| docker-compose ps") and 2>/dev/null shims were dropped as part
+	// of ADR-084 so stderr surfaces naturally for the error formatter.
+	output, err := s.remoteExecutor.Run(checkCtx, "docker compose ps --format json")
 	if err != nil {
 		return CheckResult{
 			Name:     "Remote containers",
 			Severity: SeverityFail,
-			Detail:   fmt.Sprintf("could not query remote containers: %v", err),
+			Detail:   formatRemoteError(err, "docker compose"),
```
tibtof: **Gap vs ADR-084 acceptance criteria (bug 3)**: the formatter never sees real stderr from `ssh.Executor.Run`. The SSH adapter merges stdout+stderr into a single `bytes.Buffer` (`executor.go:104-106`) and returns it as the first tuple value; `c.Run()` does NOT populate `*exec.ExitError.Stderr` — Go only fills that field when using `cmd.Output()`. Therefore `extractExitInfo` always returns `(exitCode, "")` in production, and the first-non-empty-line-of-stderr path is dead code. Worse, `checkRemoteContainerHealth` discards the merged output on the error branch:

```go
output, err := s.remoteExecutor.Run(checkCtx, "docker compose ps --format json")
if err != nil {
    return CheckResult{... Detail: formatRemoteError(err, "docker compose")}
}
```

So the actual stderr (e.g. `docker: command not found`) never reaches the user — only the exit code + canned hint do. The ADR explicitly requires "the first non-empty line of stderr" to appear in the detail. Fix: either extend `formatRemoteError(err error, output string, cmdName string)` and pass `output` in, or have the SSH adapter capture stderr separately and stuff it into `exitErr.Stderr` before returning. The unit tests pass today only because `makeExitErr` manually sets `exitErr.Stderr` — that codepath does not exist in production.

tibtof:
## Reviewer agent verdict: CHANGES REQUESTED

Scope and architecture are solid — ADR-084 is a well-reasoned design. The `PortOwnerProbe` port lives correctly in `internal/ports/`, the `VibeWardenHealthProbe` adapter in `internal/adapters/ops/`, DI via `WithPortOwnerProbe`, no cross-layer imports. Godoc present on all new exported types. `InsecureSkipVerify` is commented at both usages. `httptest.NewTLSServer` lifecycle is tidy (`defer srv.Close()` / `t.Cleanup`). `make check` passes.

### Blocker — stderr loss in remote-container error path

The formatter was designed to surface "the first non-empty line of stderr" per ADR-084 AC (bug 3), but in production the SSH adapter merges stdout+stderr into the first return value and never populates `exitErr.Stderr` — Go only fills that field for `cmd.Output()`, not `cmd.Run()`. `checkRemoteContainerHealth` then discards the merged `output` on the error branch:

```go
output, err := s.remoteExecutor.Run(checkCtx, "docker compose ps --format json")
if err != nil {
    return CheckResult{... Detail: formatRemoteError(err, "docker compose")}
}
```

So the actual stderr (e.g. `docker: command not found`) never reaches the user — only `exit <code>` + canned hint do. Unit tests pass because `makeExitErr` manually sets `exitErr.Stderr = []byte(stderr)`; that codepath does not exist in the real ssh executor.

**Fix options (either works):**
1. Change signature to `formatRemoteError(err error, output string, cmdName string)` and pass the currently-discarded `output` from `checkRemoteContainerHealth`. Small patch, keeps the adapter untouched.
2. Have `ssh.Executor.Run` capture stderr into a separate buffer and set `exitErr.Stderr` before wrapping.

Option 1 is less invasive. Add a test using the real `ssh.Executor` contract (stderr in first return, `exitErr.Stderr` empty) to prevent regression.

### Nit (non-blocking)
Consider structured-logging the full raw ssh stderr at debug level in `ssh.Executor.Run` so operators can still pull the raw stream from logs when the user-facing detail is truncated.

### Approved aspects
- Port-ownership probe design and test matrix (signature match, foreign 200, plain TCP, closed port, non-2xx, wildcard-host rewrite, strict prefix)
- TLS live-handshake rewrite: replaces the broken filesystem scan, uses the embedded sidecar as source of truth (per ADR-084 decision 2)
- Cross-platform stdlib-only code paths (no darwin/linux-specific branches)
- Hexagonal boundaries respected throughout
- Test-time `doctorConfig()` helper (sets `TLS.Provider = "external"`) is scope-appropriate, not a drive-by

Inline comment with full technical detail on `doctor.go:537`.

tibtof:
Addressed both reviews. `make check` passed locally.

### Reviewer blocker (stderr discarded on remote-container error path)
`b916ea0` — `fix(#1054): pass merged ssh output to formatRemoteError so real stderr reaches user`

- Signature change: `formatRemoteError(err error, output string, cmdName string)`. The adapter merges stdout+stderr into the `output` return value and leaves `*exec.ExitError.Stderr` empty (cmd.Run, not cmd.Output); the formatter now reads `output` first, falls back to `exitErr.Stderr` for defence in depth, and then to the wrapped error string.
- Call site: `checkRemoteContainerHealth` now forwards the previously-discarded `output` to the formatter.
- Regression test: `TestFormatRemoteError_SSHAdapterContract` builds a real `*exec.ExitError` via `exec.Command("sh","-c","exit N").Run()` (so `exitErr.Stderr` is genuinely empty) and wraps with `"ssh exit: %w"` to mirror `ssh.Executor.Run`. Cases cover `docker: command not found`, permission denied, multi-line compose errors (only first line leaks), and empty output.

### Writer blocker 1 (CHANGELOG)
`0000899` — three bullets added under `[Unreleased] → Fixes` covering the port-OK case, the live-TLS-handshake cert check, and the remote-error formatting.

### Writer blocker 2 (docs/troubleshooting.md)
`bf4dcd7`
- Check #10 description rewritten to describe the live TLS handshake and the WARN "start 'vibew dev'" case.
- Stale "Port conflict + Docker not running" sample replaced with two samples: the new OK case (`in use by local vibew dev (expected)`) and the foreign-process FAIL case.
- "Port conflict — proxy port already in use" section now leads with a blockquote noting the FAIL does not fire for a running `vibew dev`.

### `make check`
`==> All checks passed!` (full suite including `-race`, lint, demo-app).

internal/app/ops/doctor.go
tibtof: Resolved in commit `b916ea0`. `formatRemoteError` signature is now `(err, output, cmdName)`; the `checkRemoteContainerHealth` call site forwards the merged ssh output that was previously discarded. `TestFormatRemoteError_SSHAdapterContract` exercises the real adapter shape (genuine `*exec.ExitError` with empty `.Stderr`, `"ssh exit: %w"` wrapper) and asserts real stderr lines like `docker: command not found` reach the user-facing detail. Confirmed against `internal/adapters/ssh/executor.go:108-109`.

tibtof:
## Reviewer agent verdict: LGTM

Round-1 blocker is truly resolved.

**Signature propagation** — single callsite: `internal/app/ops/doctor.go:542` forwards the merged `output` to `formatRemoteError(err, output, "docker compose")`. No stale callers using the dead `(err, cmdName)` pattern anywhere in `internal/`.

**Regression test verified against production contract** — `TestFormatRemoteError_SSHAdapterContract` (`remote_error_internal_test.go:210`) uses `makeSSHAdapterErr`, which:
- Runs a real `exec.Command("sh", "-c", "exit N").Run()` to produce a genuine `*exec.ExitError`.
- Asserts `len(exitErr.Stderr) == 0` as a precondition — so the fallback to `exitErr.Stderr` cannot mask a bug.
- Wraps with `fmt.Errorf("ssh exit: %w", runErr)` — byte-identical to `internal/adapters/ssh/executor.go:109`.
- Returns the merged output as the first value, exactly like `ssh.Executor.Run`.

Test cases cover `docker: command not found` (exit 127), permission denied (126), multi-line compose output (only first line surfaces, stack trace dropped), and empty merged output (falls back to default hint). All assert `"ssh exit"` does not leak.

**First-non-empty-line extraction** — `firstMeaningfulLine` uses `strings.Split(s, "\n")` + `TrimSpace` per line; `sanitizeSingleLine` collapses `\r\n`, `\n`, `\r` before length-capping. Leading blank lines (`"\n\nfirst line\n..."`) and CRLF merged output are both handled — covered by `exit 1 with multi-line stderr` case.

**No regressions** — `git show --stat b916ea0` confirms the fix touches only `doctor.go`, `remote_error.go`, `remote_error_internal_test.go`. Port-ownership probe (ADR-084 decision 1) and live-TLS-handshake cert check (decision 2) are untouched.

**`make check` green** locally — full suite including `-race`, lint, demo-app: `==> All checks passed!`.

Approving.

tibtof:
## Writer review — round 2

All three findings are addressed. Approving.

**Verified via `gh pr diff 1060`:**

1. **CHANGELOG Unreleased** (commit `0000899`) — three user-actionable `### Fixes` bullets now present: port-conflict false positive, TLS cert live-handshake, remote-container shell-leak. Each bullet names the observable before/after output (e.g. `[OK] Proxy port  in use by local vibew dev (expected)` vs the old `[FAIL]`), which is exactly the format a user grepping the changelog needs.

2. **`docs/troubleshooting.md` sample output** (commit `bf4dcd7`) — now has both an OK case ("Proxy port already owned by a running `vibew dev`") and a relabelled FAIL case ("Port conflict (foreign process) + Docker not running"). The OK narrative correctly attributes the new behaviour to `/_vibewarden/health` probe + ADR-084.

3. **Check #10 description** (commit `bf4dcd7`) — correctly describes the live TLS handshake, the 7-day expiry threshold, and the WARN fallback (`sidecar not reachable — start 'vibew dev'`). No longer claims the check reads `server.crt` on disk.

**Minor polish (non-blocking — fine to ship as-is or address in follow-up):**

- CHANGELOG bullet 2 cites `tls.Conn.ConnectionState().PeerCertificates` — that's Go-level detail most users won't parse. A phrase like "reads the certificate the sidecar is actually serving" conveys the same thing in user language.
- `docs/troubleshooting.md` check #10 row is 4 sentences and includes ADR history ("so the check no longer depends on a file-on-disk at a hardcoded path (ADR-084)"). Reference tables should answer "what it tests" in one line; migration notes belong in the changelog (already there).
- `docs/troubleshooting.md` line 112 sample output shows `[OK]  Proxy port  in use by local vibew dev (expected)` but the actual detail string emitted by `doctor.go:300` is `port 8443 in use by local vibew dev (expected)` (with the `port 8443` prefix, matching the format of the other Proxy-port details in the doc). Worth syncing so copy-paste regex matches.

None of these block merge. Docs are consistent with the code shipped in this PR.

Writer review: APPROVED.

Issue #1086, opened 2026-04-21
fix: vibew add <feature> destructively rewrites vibewarden.production.yaml — wipes all commented stanzas


tibtof:
## Problem

`vibew add tls` (and any future `vibew add` subcommand that touches `vibewarden.production.yaml`) destructively rewrites the file via `yaml.Unmarshal` → `map[string]any` → `yaml.Marshal`, deleting every commented-out stanza, losing key ordering, and dropping whitespace. A user running `vibew add tls` on a carefully-curated production config silently loses WAF block-mode overrides, auth overrides, and header overrides. Retro 2026-04-21, v0.16.0. Release-blocker.

## Root cause

`internal/cli/cmd/add_tls.go::upsertTLSFieldsInProdConfig` reads `vibewarden.production.yaml` into a `map[string]any`, mutates it, and re-marshals. `gopkg.in/yaml.v3`'s map path does not preserve comments, ordering, or whitespace.

The existing `internal/adapters/yamlmod/Toggler` uses `yaml.v3` node trees (HeadComment/LineComment/FootComment preserved, original ordering preserved) and is the correct path — but it was not wired up for production.yaml. The feature-block builders (`upsertTLSBlock`, `upsertWAFBlock`, etc.) already use node trees and already work correctly for `vibewarden.yaml`.

Scope today: only `add tls` writes to production.yaml. But the bug class — ad-hoc map-round-trip YAML edits — must be forbidden everywhere, so this story also audits every `add` subcommand and forbids the pattern.

## User story

As a vibe coder with a curated `vibewarden.production.yaml` containing active TLS config plus commented-out WAF block-mode and auth overrides, when I run `vibew add tls --domain foo.com` I want only the `tls.domain` key to change and every other stanza — active or commented — to survive byte-for-byte, so that I never silently lose production security decisions.

## Acceptance criteria

- [ ] Every `vibew add <feature>` subcommand that reads or writes `vibewarden.yaml` OR `vibewarden.production.yaml` routes the edit through `yaml.v3` node trees (comments, ordering, whitespace preserved). No `yaml.Unmarshal` into `map[string]any` followed by `yaml.Marshal` anywhere in `internal/cli/cmd/add_*.go` or `internal/app/scaffold/add_*.go`.
[...]

PR #1096, opened 2026-04-21
fix(#1078,#1090): TLS state accuracy in vibew status + doctor

### LINKED ISSUE #1078: fix: vibew doctor reports TLS cert expires in 0 day(s) on fresh scaffold
## Problem

On a freshly scaffolded project with \`vibew dev\` running, \`vibew doctor\` reports:

\`\`\`
[WARN] TLS certificate expires in 0 day(s) on 2026-04-21
\`\`\`

The certificate should be fresh (self-signed, just issued by Caddy). Either:
- The cert's expiry time is wrong
- The doctor check's "days remaining" math is off
- The message should distinguish "dev self-signed cert regenerated daily" from "production cert expiring"

## Fix

- If dev certs genuinely have 1-day TTL, reword doctor's message: "TLS cert expires in N days (dev self-signed — auto-renews)"
- If not, debug Caddy's internal issuer's NotAfter vs NotBefore

## Found in

Pre-release e2e test (2026-04-21). The live-handshake check added in ADR-084 works correctly (it reads the leaf cert), so this may be a Caddy internal config issue with the self-signed cert's validity period.

### LINKED ISSUE #1090: fix: vibew status shows ✓ TLS while Let's Encrypt cert is still being obtained
## Problem

After deploy, \`vibew status\` reports:
\`\`\`
✓ TLS enabled
\`\`\`
even when Caddy hasn't yet issued the Let's Encrypt cert. The curl to the domain fails with a TLS handshake error while the cert is still being obtained. There's no intermediate state ("pending", "obtaining", "retrying") exposed to the operator.

## Retro evidence (2026-04-21)

> \`curl: exit 35 (TLS handshake fail while cert was still being issued)\` — bad. \`vibew status\` showed \`✓ TLS enabled\` while certs weren't obtained yet. There's no "cert pending / obtaining" state exposed.

## Acceptance criteria

\`vibew status\` must distinguish:

- \`✓ TLS obtained (expires 2026-07-21)\` — cert exists, not expiring soon
- \`⏳ TLS obtaining (ACME in progress)\` — Caddy is mid-HTTP-01 or mid-TLS-ALPN-01 exchange
- \`⚠ TLS failing (N retries, last error: <first line of Caddy error>)\` — cert obtain keeps failing; show the upstream reason
- \`✗ TLS disabled\` — as today
- \`⚠ TLS near expiry (expires in 5 days)\` — existing check

Source of truth: query Caddy's admin API (\`/pki/...\` or \`/config/\` endpoints) for cert state rather than guessing from config.

## Related

- #1078 — \`vibew doctor\` has a related "expires in 0 days" bug (same class of issue, different command). Consider co-implementing.

## Found in

v0.16.0 fresh-agent retro.

tibtof:
Closes #1090, #1078. Part of v0.17.0. Supersedes the ad-hoc handshake check in ADR-084 with a shared domain type.

## Summary

- New `internal/domain/tls.State` value object with variants: `Disabled`, `SelfSignedLocal`, `Obtaining`, `Obtained{ExpiresAt}`, `ExpiringSoon{DaysLeft}`, `Failing{LastError}`, `Unknown`. `String()` emits the PM-spec display strings.
- New `ports.TLSStateResolver` + `ports.ErrNotInProcess` sentinel.
- Two adapters: `caddy.InProcessResolver` (primary, inspects Caddy's in-memory state; short-circuits `ErrNotInProcess` when Caddy is not embedded) and `caddy.HandshakeResolver` (fallback, live TLS handshake — issuer CN `Caddy Local Authority` → `SelfSignedLocal` without consulting NotAfter).
- `ops.ChainResolver` composes primary → fallback → `Unknown`.
- Pure renderers `renderTLSStatusLine` and `renderTLSDoctorCheck` are the single source of truth for status and doctor output.
- `StatusService` + `DoctorService` gain `WithTLSStateResolver`. CLI composition roots (`doctor`, `status`, `mcp`) wire the chain.
- The old `checkTLSCertValid` handshake logic is moved behind the resolver; `vibew doctor` no longer prints `expires in 0 day(s)` for self-signed dev certs (fixes #1078); `vibew status` replaces `✓ TLS enabled` with state-aware detail (fixes #1090).

## Test plan

- Table-driven unit tests per variant: `internal/domain/tls/state_test.go`, `internal/adapters/caddy/tls_state_test.go` (fake cert provider), `internal/adapters/caddy/tls_state_handshake_test.go` (httptest TLS server, self-signed + external issuer + unreachable rows), `internal/app/ops/chain_resolver_test.go`, `internal/app/ops/status_tls_render_test.go`, `internal/app/ops/doctor_tls_render_test.go`.
- Explicit regression test `TestRenderTLSDoctorCheck_NoExpiringZeroDaysForSelfSigned` locks bug #1078.
- `make check` — full build + lint + tests on both modules.

### DIFF
--- a/internal/app/ops/doctor_tls_render.go
+++ b/internal/app/ops/doctor_tls_render.go
@@ -7,11 +7,6 @@ import (
 	tlsdomain "github.com/vibewarden/vibewarden/internal/domain/tls"
 )
 
-// nowFn is the clock used by renderTLSDoctorCheck. It is a package-level
-// variable so tests can replace it with a fixed time without threading a
-// clock through the renderer signature.
-var nowFn = time.Now
-
 // renderTLSDoctorCheck converts a TLSState into a CheckResult for the
 // `vibew doctor` report. Severity mapping follows the PM spec for
 // #1090 / #1078:
@@ -21,7 +16,10 @@ var nowFn = time.Now
 //
 // The Name is always "TLS certificate" and the Section is left to the
 // caller (the doctor service wraps it with the Local Runtime section).
-func renderTLSDoctorCheck(state tlsdomain.State) CheckResult {
+//
+// now is injected so callers can fix the clock in tests without mutating
+// any package-level variable.
+func renderTLSDoctorCheck(state tlsdomain.State, now func() time.Time) CheckResult {
 	result := CheckResult{Name: "TLS certificate"}
 
 	switch state.Kind() {
@@ -32,7 +30,7 @@ func renderTLSDoctorCheck(state tlsdomain.State) CheckResult {
 		result.Severity = SeverityOK
 		result.Detail = "self-signed dev cert (rotates automatically)"
 	case tlsdomain.KindObtained:
-		daysLeft := int(state.ExpiresAt().Sub(nowFn()).Hours() / 24)
+		daysLeft := int(state.ExpiresAt().Sub(now()).Hours() / 24)
 		if daysLeft < 0 {
 			daysLeft = 0
 		}
@@ -42,7 +40,7 @@ func renderTLSDoctorCheck(state tlsdomain.State) CheckResult {
 		// An ExpiringSoon state with NotAfter already in the past is
 		// really "expired" — escalate to FAIL so automation treats it
 		// as a hard failure.
-		if !state.ExpiresAt().IsZero() && nowFn().After(state.ExpiresAt()) {
+		if !state.ExpiresAt().IsZero() && now().After(state.ExpiresAt()) {
 			result.Severity = SeverityFail
 			result.Detail = fmt.Sprintf("expired on %s", state.ExpiresAt().Format("2006-01-02"))
 		} else {
--- a/internal/app/ops/doctor_tls_render.go
+++ b/internal/app/ops/doctor_tls_render.go
@@ -1,0 +1,71 @@
+package ops
+
+import (
+	"fmt"
+	"time"
+
+	tlsdomain "github.com/vibewarden/vibewarden/internal/domain/tls"
+)
+
+// nowFn is the clock used by renderTLSDoctorCheck. It is a package-level
+// variable so tests can replace it with a fixed time without threading a
+// clock through the renderer signature.
+var nowFn = time.Now
+
+// renderTLSDoctorCheck converts a TLSState into a CheckResult for the
+// `vibew doctor` report. Severity mapping follows the PM spec for
+// #1090 / #1078:
+//   - Disabled, SelfSignedLocal, Obtained (>7d) → OK
+//   - Obtaining, ExpiringSoon, Unknown        → WARN
+//   - Failing                                  → FAIL
+//
+// The Name is always "TLS certificate" and the Section is left to the
+// caller (the doctor service wraps it with the Local Runtime section).
+func renderTLSDoctorCheck(state tlsdomain.State) CheckResult {
+	result := CheckResult{Name: "TLS certificate"}
+
+	switch state.Kind() {
+	case tlsdomain.KindDisabled:
+		result.Severity = SeverityOK
+		result.Detail = "TLS plugin disabled"
+	case tlsdomain.KindSelfSignedLocal:
+		result.Severity = SeverityOK
+		result.Detail = "self-signed dev cert (rotates automatically)"
+	case tlsdomain.KindObtained:
+		daysLeft := int(state.ExpiresAt().Sub(nowFn()).Hours() / 24)
+		if daysLeft < 0 {
+			daysLeft = 0
+		}
+		result.Severity = SeverityOK
+		result.Detail = fmt.Sprintf("valid until %s (%d days)", state.ExpiresAt().Format("2006-01-02"), daysLeft)
+	case tlsdomain.KindExpiringSoon:
+		// An ExpiringSoon state with NotAfter already in the past is
+		// really "expired" — escalate to FAIL so automation treats it
+		// as a hard failure.
+		if !state.ExpiresAt().IsZero() && nowFn().After(state.ExpiresAt()) {
+			result.Severity = SeverityFail
+			result.Detail = fmt.Sprintf("expired on %s", state.ExpiresAt().Format("2006-01-02"))
+		} else {
+			result.Severity = SeverityWarn
+			result.Detail = fmt.Sprintf("expires in %d day(s) on %s", state.DaysLeft(), state.ExpiresAt().Format("2006-01-02"))
+		}
+	case tlsdomain.KindObtaining:
+		result.Severity = SeverityWarn
+		result.Detail = "ACME exchange in progress — rerun in 1-2 minutes"
+	case tlsdomain.KindFailing:
+		result.Severity = SeverityFail
+		if state.LastError() == "" {
+			result.Detail = "ACME exchange failed"
+		} else {
+			result.Detail = fmt.Sprintf("ACME exchange failed: %s", state.LastError())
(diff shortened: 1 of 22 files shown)


tibtof:
**Writer agent review — changes requested.**

Three doc gaps found:

**1. CHANGELOG.md — no entry for this PR**
`[Unreleased] > Fixed` has #1086 but nothing here. Required entries:
- `vibew status` now shows TLS state detail (Obtaining / Obtained / Failing / SelfSignedLocal) instead of `✓ TLS enabled` (#1090)
- `vibew doctor` no longer prints `expires in 0 day(s)` for self-signed dev certs (#1078)

**2. `llms-full.txt` — `vibew doctor` checks list is stale**
"vibew doctor checks (in order)" (line 1445) stops at check 6 and omits the TLS state check. Add:
```
  7. TLS state — cert state (Obtaining / Obtained / SelfSignedLocal / Failing / Unknown)
```
`vibew status` table entry (line 1301) does not indicate state-aware TLS output. An agent cannot predict the new output strings.

**3. `docs/examples/AGENTS-VIBEWARDEN.md` — TLS state variants absent**
TLS section covers config keys but not what `vibew status` / `vibew doctor` emit per state. An AI agent cannot pattern-match `TLS: obtaining certificate…` or `TLS: certificate valid, expires in N days` without this.

internal/app/ops/doctor_tls_render.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"time"
+
+	tlsdomain "github.com/vibewarden/vibewarden/internal/domain/tls"
+)
+
+// nowFn is the clock used by renderTLSDoctorCheck. It is a package-level
+// variable so tests can replace it with a fixed time without threading a
+// clock through the renderer signature.
+var nowFn = time.Now
```
tibtof: **Global state violation**: `nowFn` is a package-level variable used as a mutable clock hook. CLAUDE.md mandates "no global state — everything passed via dependency injection". This pattern also makes the test in `doctor_tls_render_test.go` racy if tests ever run in parallel (two tests both reassign `nowFn`).

**Why it matters**: global mutable state is the #1 cause of flaky parallel tests and breaks the DI contract.

**Fix**: add a `nowFn func() time.Time` field to the renderer or pass the clock as a parameter. Since `renderTLSDoctorCheck` is a pure function today, the simplest fix is to accept a `clock func() time.Time` argument: `func renderTLSDoctorCheck(state tlsdomain.State, now func() time.Time) CheckResult`. The two call sites in `doctor.go` pass `time.Now` (or an injected clock), and the test passes its fixed clock directly without touching a global.

internal/app/ops/status.go
```diff
@@ -51,6 +52,14 @@ func (s *StatusService) WithLogs(logs ports.ComposeLogs) *StatusService {
 	return s
 }
 
+// WithTLSStateResolver attaches a TLS state resolver used to render the
+// TLS row with state-aware detail (obtained/obtaining/self-signed/...).
+// When nil, the TLS row falls back to the legacy config-only rendering.
+func (s *StatusService) WithTLSStateResolver(r ports.TLSStateResolver) *StatusService {
```
tibtof, line 63: **Inconsistent copy-on-write**: `WithCompose`, `WithLogs`, and `WithTLSStateResolver` on `StatusService` all mutate the receiver in-place and return `s`. `DoctorService.WithTLSStateResolver` (and `WithPortOwnerProbe`) correctly do `cp := *s; cp.field = r; return &cp`.

**Why it matters**: the mutation pattern is safe only when the caller never retains the original pointer after calling `With*`. The copy pattern is safe always. The inconsistency within the same package will confuse future contributors, and if `StatusService` is ever shared across goroutines a data race will result.

**Fix**: align `StatusService.WithTLSStateResolver` (and ideally `WithCompose`/`WithLogs` for consistency) with the copy-on-write pattern already established in `DoctorService`:
```go
func (s *StatusService) WithTLSStateResolver(r ports.TLSStateResolver) *StatusService {
    cp := *s
    cp.tlsState = r
    return &cp
}
```
Note: `WithCompose`/`WithLogs` already exist with the mutation pattern, so fixing only `WithTLSStateResolver` would still be inconsistent. Fix all three in this PR since it touches the file.

internal/app/ops/tls_handshake_fallback.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/vibewarden/vibewarden/internal/config"
+	tlsdomain "github.com/vibewarden/vibewarden/internal/domain/tls"
+	"github.com/vibewarden/vibewarden/internal/ports"
+)
+
+// caddyLocalIssuerCN is the issuer Common Name Caddy stamps on internal
+// self-signed dev leaves. Duplicated here (also defined in the caddy
+// adapter) to avoid an app → adapter import violation. If the constant
+// ever changes, bump it in both places.
```
tibtof: **Duplicated constant with manual sync note**: `caddyLocalIssuerCN` is defined here in `internal/app/ops/tls_handshake_fallback.go` and also in `internal/adapters/caddy/tls_state.go` (shared within `package caddy`). The comment explicitly acknowledges this and instructs developers to "bump it in both places" if it changes.

**Why it matters**: constants with a manual-sync requirement are a maintenance hazard and a future bug waiting to happen. The existence of the inline fallback itself is the root cause.

**Fix**: the `inlineHandshakeResolver` in `tls_handshake_fallback.go` exists purely to avoid importing `internal/adapters/caddy` from `internal/app/ops`. A cleaner approach is to export `CaddyLocalIssuerCN` from `internal/adapters/caddy` and have the constant be the single source of truth — but that imports an adapter from app, which is also wrong. The real fix is to move the constant to `internal/domain/tls` (it is a domain fact about how Caddy identifies its local CA, not an adapter-internal detail), then both the caddy adapter and the ops fallback import from there. Alternatively, eliminate `inlineHandshakeResolver` entirely by making `caddy.HandshakeResolver` constructible with a nil config (it already guards `r.cfg != nil`) and exporting it so the fallback wiring in `doctor.go` does not need the inline copy. Either approach removes the duplicated constant.

tibtof:
## Reviewer Agent verdict — CHANGES REQUESTED

3 issues block merge.

**1. Global state violation** — `var nowFn = time.Now` in `internal/app/ops/doctor_tls_render.go:13` is mutable package-level state. CLAUDE.md: "No global state — everything passed via dependency injection." It also makes parallel tests racy. Fix: pass clock as a parameter to `renderTLSDoctorCheck`.

**2. Inconsistent copy-on-write on StatusService** — `WithTLSStateResolver` (and `WithCompose`, `WithLogs`) mutate the receiver in-place while `DoctorService.WithTLSStateResolver` uses the correct copy pattern (`cp := *s`). Fix: apply copy-on-write to all three `With*` methods on `StatusService`.

**3. Duplicated constant with manual-sync comment** — `caddyLocalIssuerCN` is defined in both `internal/adapters/caddy/tls_state.go` and `internal/app/ops/tls_handshake_fallback.go` with a note saying "bump it in both places". Fix: move the constant to `internal/domain/tls` so both packages import a single source of truth.

Everything else is correct: domain layer is stdlib-only, `ErrNotInProcess` sentinel is in `ports`, `SelfSignedLocal` is keyed on `Caddy Local Authority` issuer CN (not NotAfter math), the regression test `TestRenderTLSDoctorCheck_NoExpiringZeroDaysForSelfSigned` is present, all 7 `Kind` variants exist, `String()` matches the PM spec, table-driven tests cover every variant, no new unapproved licenses.

PR #1118, opened 2026-04-23
feat(#1057): LE rate-limit preflight via crt.sh — avoid blind ACME retries on exhausted budgets

### LINKED ISSUE #1057: feat: LE rate-limit preflight before prod TLS issuance
## Problem

When Let's Encrypt's per-identifier rate limit (5 certs / 168h) is hit, the sidecar silently retries through the ACME fallback chain. By the time the chain is exhausted, 10+ minutes of production downtime may have passed. Operators have no way to know in advance that the domain is rate-limited.

## Split from #1055

This issue is split out from the ACME fallback chain hardening work (#1055) because it needs a separate design call:

- Where does preflight live? (subcommand `vibew tls preflight`? part of `vibew doctor`? a pre-deploy hook?)
- What data source? (`crt.sh` public CT log search, LE's own API, local cert cache?)
- Should preflight be blocking or advisory?
- How do we avoid false negatives when the user legitimately needs to re-issue?

## Acceptance criteria (placeholder — PM agent to refine)

- [ ] Design doc or ADR proposing a preflight mechanism for LE rate-limit state
- [ ] Implementation gated on design approval
- [ ] Preflight must work without side-effects (no real ACME request)
- [ ] Clear operator message when rate-limit is detected (what, when, how long to wait)

## Out of scope

- ZeroSSL email preflight (handled in #1055)
- Removing Buypass (handled in #1055)
- Generic per-CA rate-limit preflight — LE is the only CA with a well-known per-identifier budget that we can query externally

## Related

- #1055 — parent ACME chain hardening issue
- #1026 / #1042 — original fallback chain implementation
- ADR-079 — ACME fallback chain multi-issuer

tibtof:
## Summary

- Adds LE rate-limit preflight via crt.sh CT log query.
- Runs under `vibew doctor`; `--skip-le-preflight` + `tls.skip_rate_limit_check` opt-out.
- ADR-090.
- Closes #1057.

### New packages

- `internal/domain/tlspreflight/` — pure domain types: `Status`, `Result`, `Budget` constants, error sentinels, `CountIssuedSince` parser. 100% coverage.
- `internal/ports/cert_transparency.go` — `CertTransparencyQuerier` interface and `CrtShRecord` projection.
- `internal/adapters/crtsh/` — HTTP adapter calling `crt.sh/?Identity=<domain>&output=json`. Handles 429/500/malformed/empty/context-cancel.
- `internal/app/tlspreflight/` — `Service.Check(ctx, domains)` orchestrates queries, maps all error sentinels to WARN, preserves input order. 100% coverage.

### Existing file edits

- `internal/app/ops/doctor.go` — `WithLERateLimitService` setter, `checkLERateLimit` helper, six ADR-090 guards in `runChecks`.
- `internal/cli/cmd/doctor.go` — `--skip-le-preflight` flag, crt.sh adapter wiring (10s timeout per AC-8), `deriveRegisteredDomains` via `publicsuffix`.
- `internal/config/config.go` — `TLSConfig.SkipRateLimitCheck bool` field.
- `go.mod` — `golang.org/x/net` promoted from indirect to direct (BSD-3).
- `docs/troubleshooting.md` — flags table row, check table row, full "LE rate-limit preflight" section.
- `llms-full.txt` — doctor checks list and flags updated.
- `CHANGELOG.md` — [Unreleased] Added entries.

### Grep contract

`grep -rn "crtsh\." internal/plugins/ internal/adapters/caddy/` → 0 matches.  
The crt.sh adapter is never instantiated from the sidecar runtime — only from `internal/cli/cmd/doctor.go`.

## Test plan

- [x] `make check` passes (gofmt, golangci-lint, build, `go test -race ./...`)
- [x] Domain + app coverage 100% (well above 85% target)
- [x] crt.sh adapter unit tests cover 429 / 500 / empty / malformed / context-cancel / User-Agent / URL escaping / malformed-row-skipped
- [x] App service tests cover all threshold transitions (0–5/5), all three error sentinels, multi-domain mixed outcomes, result order, next-slot time
- [x] `DoctorService` tests cover WARN / FAIL / SkipRateLimitCheck / SkipFlag / non-LE provider / ACME CA override / no domains / network error / JSON output
- [x] CLI tests verify `--skip-le-preflight` flag registered, long description mentions rate-limit, help text contains flag name

### DIFF
--- a/internal/adapters/crtsh/client.go
+++ b/internal/adapters/crtsh/client.go
@@ -21,6 +21,11 @@ const (
 	// userAgent is sent on every request so crt.sh operators can identify the
 	// tool. The URL is informational — it helps them contact us if we cause load.
 	userAgent = "vibew-doctor/1 (+https://vibewarden.dev)"
+
+	// maxResponseBytes is the upper bound on crt.sh response body size.
+	// 10 MiB is generous for even the largest CT query result sets; a response
+	// that exceeds this cap is treated as malformed (ErrCTResponseMalformed).
+	maxResponseBytes = 10 * 1024 * 1024
 )
 
 // Client implements ports.CertTransparencyQuerier by calling the crt.sh
@@ -93,9 +98,17 @@ func (c *Client) Query(ctx context.Context, registeredDomain string) ([]ports.Cr
 		return nil, fmt.Errorf("%w: Content-Type %q", tlspreflight.ErrCTResponseMalformed, ct)
 	}
 
-	body, err := io.ReadAll(resp.Body)
+	// Bound the response body to maxResponseBytes to prevent memory-exhaustion
+	// from a spoofed or misbehaving server. io.LimitReader returns io.EOF at
+	// the limit — we detect an over-limit response by checking whether the
+	// returned body is exactly maxResponseBytes after a successful read.
+	limited := io.LimitReader(resp.Body, maxResponseBytes+1)
+	body, err := io.ReadAll(limited)
 	if err != nil {
 		return nil, fmt.Errorf("%w: read body: %v", tlspreflight.ErrCTUnavailable, err)
+	}
+	if int64(len(body)) > maxResponseBytes {
+		return nil, fmt.Errorf("%w: response exceeds %d-byte limit", tlspreflight.ErrCTResponseMalformed, maxResponseBytes)
 	}
 
 	// crt.sh returns an empty array "[]" for unknown domains — treat as 0 certs.
--- a/internal/adapters/crtsh/client.go
+++ b/internal/adapters/crtsh/client.go
@@ -1,0 +1,159 @@
+package crtsh
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/vibewarden/vibewarden/internal/domain/tlspreflight"
+	"github.com/vibewarden/vibewarden/internal/ports"
+)
+
+const (
+	// defaultBase is the crt.sh base URL used by NewClient.
+	defaultBase = "https://crt.sh"
+
+	// userAgent is sent on every request so crt.sh operators can identify the
+	// tool. The URL is informational — it helps them contact us if we cause load.
+	userAgent = "vibew-doctor/1 (+https://vibewarden.dev)"
+)
+
+// Client implements ports.CertTransparencyQuerier by calling the crt.sh
+// public JSON API over HTTPS.
+//
+// Construct via NewClient or NewClientWithBase (for tests). The client is
+// stateless and safe for concurrent use.
+type Client struct {
+	http *http.Client
+	base string
+}
+
+// NewClient returns a Client that calls the production crt.sh endpoint.
+// httpClient must have a Timeout set (the caller in cmd/doctor.go uses 10s
+// per AC-8).
+func NewClient(httpClient *http.Client) *Client {
+	return &Client{http: httpClient, base: defaultBase}
+}
+
+// NewClientWithBase returns a Client with a custom base URL, used in tests to
+// point at an httptest.Server instead of the real crt.sh.
+func NewClientWithBase(httpClient *http.Client, base string) *Client {
+	return &Client{http: httpClient, base: base}
+}
+
+// crtShRow is the JSON shape returned by crt.sh for each certificate record.
+// Only the fields the preflight consumes are decoded; others are ignored.
+type crtShRow struct {
+	NotBefore  string `json:"not_before"`
+	IssuerName string `json:"issuer_name"`
+	CommonName string `json:"common_name"`
+	NameValue  string `json:"name_value"`
+}
+
+// Query fetches certificates from crt.sh for the given registered domain and
+// returns them as a slice of ports.CrtShRecord.
+//
+// Error mapping:
+//   - network / DNS / context cancel  → tlspreflight.ErrCTUnavailable (wrapped)
+//   - HTTP 429                        → tlspreflight.ErrCTThrottled (wrapped)
+//   - HTTP non-200 (except 429)       → tlspreflight.ErrCTUnavailable (wrapped)
+//   - Content-Type not JSON           → tlspreflight.ErrCTResponseMalformed
+//   - JSON decode error               → tlspreflight.ErrCTResponseMalformed (wrapped)
+//   - empty body / empty array        → nil error, empty slice
+func (c *Client) Query(ctx context.Context, registeredDomain string) ([]ports.CrtShRecord, error) {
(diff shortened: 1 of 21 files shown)


internal/adapters/crtsh/client.go
```diff
@@ -0,0 +1,159 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%w: HTTP %d", tlspreflight.ErrCTUnavailable, resp.StatusCode)
+	}
+
+	ct := resp.Header.Get("Content-Type")
+	if ct != "" && !strings.HasPrefix(strings.ToLower(ct), "application/json") {
+		return nil, fmt.Errorf("%w: Content-Type %q", tlspreflight.ErrCTResponseMalformed, ct)
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
tibtof: **Missing response body size limit — potential memory exhaustion**

`io.ReadAll(resp.Body)` at line 96 reads the entire crt.sh response into memory with no upper bound. With a 10-second timeout a fast link can deliver tens of megabytes. If crt.sh is spoofed (MITM, DNS poisoning) or returns a pathologically large body, this could exhaust heap on the operator's machine.

Fix: wrap the body before reading:

```go
const maxBodyBytes = 10 << 20 // 10 MiB — crt.sh responses are ~50 KB at most
body, err := io.ReadAll(io.LimitReader(resp.Body, maxBodyBytes))
```

Risk is low because (a) the request goes to a hardcoded HTTPS endpoint, not a user-supplied URL, and (b) the 10-second timeout limits exposure. Still, this is a missing defensive measure that costs one line.

internal/cli/cmd/doctor.go
```diff
@@ -121,6 +148,30 @@ Examples:
...
+// In multi-site mode (ADR-068), the caller would extend this function to
+// iterate over all site domains; for now, only the primary TLS domain is used.
+func deriveRegisteredDomains(cfg *config.Config) []string {
+	if cfg.TLS.Domain == "" {
+		return nil
+	}
+	registered, err := publicsuffix.EffectiveTLDPlusOne(cfg.TLS.Domain)
+	if err != nil {
+		// Single-label hostnames (e.g. "localhost") cannot be normalised.
+		// The guard in runChecks will catch this — we return nil so the
```
tibtof: **ADR-090 discrepancy: single-label domain produces silence instead of WARN**

ADR-090 §(error cases) table specifies:

> `publicsuffix.EffectiveTLDPlusOne` errors (e.g. single label) → SeverityWarn, detail: `cannot derive registered domain from "<x>" — rate-limit check skipped`

The code path today:
1. `deriveRegisteredDomains(cfg)` in `doctor.go` returns `nil` when `publicsuffix.EffectiveTLDPlusOne` fails (e.g. domain = `"localhost"`).
2. The outer guard in `runChecks` still passes because `cfg.TLS.Domain != ""` is true.
3. `checkLERateLimit` receives `opts.LERegisteredDomains == nil` and returns immediately with no results.
4. `NewSkipped` is defined but never called anywhere.

Result: a user who somehow configures `provider: letsencrypt` + `domain: localhost` gets no check result and no warning — silent skip, not a WARN.

Fix: return the raw FQDN from `deriveRegisteredDomains` as a sentinel-domain when normalisation fails, and have `checkLERateLimit` detect it and emit `NewSkipped`. Alternatively, have `checkLERateLimit` compare `opts.LERegisteredDomains` to `cfg.TLS.Domain` directly. Either way, `NewSkipped` must be reachable.

Real-world impact is low (no LE domain is single-label), but the ADR contract is broken and `NewSkipped` is dead code as shipped.

tibtof:

Two issues require fixes before merge.

**1. Missing body size limit (adapter/crtsh/client.go:96)** — `io.ReadAll(resp.Body)` has no upper bound. A spoofed or misbehaving crt.sh response could exhaust heap. Fix with `io.LimitReader`. One line, inline comment posted.

**2. Dead code + ADR-090 contract violation (cli/cmd/doctor.go:172)** — `NewSkipped` is defined in `internal/domain/tlspreflight/result.go` but never called. The ADR-090 error-cases table specifies a `SeverityWarn` result when `publicsuffix.EffectiveTLDPlusOne` fails (e.g. `domain: localhost`); the code produces silence instead. `NewSkipped` is unreachable dead code as shipped. Inline comment with fix guidance posted.

**What passed:**
- Hexagonal boundaries clean: domain layer imports stdlib only; interface in `internal/ports/`; adapter in `internal/adapters/crtsh/`; app service in `internal/app/tlspreflight/`.
- ADR-090 grep contract holds: zero matches for `crtsh.` under `internal/plugins/` or `internal/adapters/caddy/`.
- All tests pass. Coverage: domain 100%, app 100%, adapter 89.6% (above the 70% floor ADR-090 sets).
- Both opt-out paths (`--skip-le-preflight` flag + `tls.skip_rate_limit_check`) tested and working.
- crt.sh base URL hardcoded (not user-configurable) — no SSRF vector.
- Context propagated through all I/O calls. HTTP timeout set (10s per AC-8).
- Outgoing headers are User-Agent and Accept only — no sensitive data leakage beyond domain name (accepted by ADR-090 §h).
- `golang.org/x/net` promoted from indirect to direct; BSD-3 license confirmed.
- No global variables, no `init()` side effects, no `panic` outside `main()`.

tibtof:
All three reviewer/writer changes addressed:

**1. crt.sh body size limit** (`internal/adapters/crtsh/client.go`)
`io.ReadAll` replaced with `io.LimitReader(resp.Body, 10MiB+1)`. Reads exceeding the cap return `ErrCTResponseMalformed`. Added `TestClient_Query_OversizedBody_MalformedError` which streams a 10MiB+1 body and asserts the error.

**2. Single-label domain WARN** (`internal/cli/cmd/doctor.go`, `internal/app/ops/doctor.go`)
`deriveRegisteredDomains` now returns `(registered []string, skipped []string)` — un-normalisable FQDNs (e.g. `localhost`) land in `skipped` instead of being dropped silently. `DoctorOptions` gained `LESkippedDomains`; `checkLERateLimit` calls `tlspreflight.NewSkipped` for each entry, emitting a `SeverityWarn` CheckResult. Added `TestDoctorService_Run_LERateLimit_SingleLabelDomain_WARN` asserting the WARN output.

**3. Reference YAML** (`vibewarden.reference.yaml`)
Added `tls.skip_rate_limit_check: false` after `acme_ca`, with a two-line comment explaining it bypasses the preflight after revocation within the 168h window and that it is equivalent to `--skip-le-preflight`.

tibtof:
**Reviewer Agent: APPROVED**

Both fixes verified: `io.LimitReader` + oversize test present; single-label domains now emit `SeverityWarn` via `LESkippedDomains`. No regressions observed.

PR #1156, opened 2026-04-26
fix(#1141): bundle uses cwd-basename when name: unset; --build agrees on the tag

### LINKED ISSUE #1141: fix: vibew bundle image-name resolution — no-name fallback broken; bundle --build builds wrong-named image
## Problem

From the qr-dali retro (project dir was `qr-dali`, no `name:` field set):

- `vibew bundle` looked for `vibewarden-app:latest` instead of the expected `qr-dali-app:latest`.
- `AGENTS-VIBEWARDEN.md` claims the fallback is the directory name. Behavior and docs disagree.
- Adding `name: qr-dali` to `vibewarden.yaml` fixed it.
- `vibew bundle --build`'s suggested fix is misleading: running it builds `qr-dali-app:latest` (correct name) and then complains about the missing `vibewarden-app:latest` (wrong name) — same outcome as `vibew bundle` alone.

This is the second-most-impactful UX issue in the retro after the deploy.sh false healthcheck.

## Possible causes

This may be a regression of #1084 (project-scoped tag work, ADR-089) or doc drift on `AGENTS-VIBEWARDEN.md`.

## Proposal

Pick one and stick to it:

**Option A — fix the default**: with no `name:` set, derive from `cwd` directory basename. Make this work end-to-end (init scaffold, build, bundle, --build).

**Option B — make `name:` mandatory and scaffold it at init**: `vibew init` writes `name: <dirname>` into the generated `vibewarden.yaml` so users never hit the fallback.

PM should pick A or B based on which is closer to the existing code path. Either way, document the chosen behavior accurately in `AGENTS-VIBEWARDEN.md`.

## Acceptance criteria

- `vibew init` in `qr-dali/` followed by `vibew bundle` (no `name:` edit) succeeds and bundles `qr-dali-app:latest`.
- `vibew bundle --build` builds the same image name that `vibew bundle` looks for (no name divergence).
- Docs match the chosen behavior.
- Regression test: integration test in a temp dir asserts the end-to-end name flow.

- CLAUDE.md §Architecture principles → Artifact policy (originally seeded as #1137; closed via PR #1152).

## Source

- `~/notes/vibewarden/retro-0.17.0.md`, retro items 2 + 4.

tibtof:
Closes #1141

## Summary

- `deriveProjectName` is now the single project-name authority inside `runBundle`. Previously `runBundle` computed `imageTag` via `cfg.ComposeProjectName()` which silently returned `"vibewarden"` when `ProjectRoot` was unpopulated by `LoadRaw` — producing `vibewarden-app:latest` regardless of the actual directory name.
- `imageTag` default is now `projectName + "-app:latest"` where `projectName` comes from `deriveProjectName`, which correctly chains `name:` → `app.image:` strip → cwd-basename fallback.
- Added explicit guard: `runBundle` returns an error if `deriveProjectName` returns empty (previously implicit via the `"vibewarden"` literal).
- `BuildOptions.ImageTag` added to `internal/app/ops/build.go`. When set, `BuildService.Run` uses it verbatim and skips `resolveImageTag`. `runBundle` passes `ImageTag: imageTag` in the `--build` branch so the build step and bundle lookup always agree byte-for-byte.
- Standalone `vibew build` behavior is unchanged — it calls `BuildService.Run` with `opts.ImageTag == ""`, so `resolveImageTag` runs as before.

## Test plan

- `TestRunBundle_ImageTagDerivation` (new, `internal/cli/cmd/bundle_test.go`) — table-driven: `name:` set, `app.image:` strip, cwd-basename fallback (`qr-dali`); asserts `VIBEWARDEN_APP_IMAGE=<expected>-app:latest` in both `sample.env` and `.env`.
- `TestBuildService_Run_UsesPreResolvedImageTag` (new, `internal/app/ops/build_test.go`) — when `BuildOptions.ImageTag` is set, the recorded builder tag equals it regardless of `cfg`.
- `TestComposeProjectName_FallsBackToVibewardenWhenProjectRootEmpty` (new, `internal/config/config_test.go`) — guard test pinning the latent `LoadRaw` misbehavior per ADR-093.
- `TestDeriveProjectName_SanitisesAdversarialInput` — existing, still passes.
- All 100+ package tests pass; `make check` passes including golangci-lint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ This initial entry was written by hand to summarise the work leading up to v0.1.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`vibew bundle` no longer ignores the cwd-basename fallback for project name** (#1141, [ADR-093](decisions/adr-093-bundle-image-name-cwd-basename-fallback.md)). With no `name:` field set, `vibew bundle` and `vibew bundle --build` now both look for `<dirname>-app:latest`, matching the documented behavior. Previously `vibew bundle` resolved the image tag via `cfg.ComposeProjectName()` which silently fell through to the literal `"vibewarden"` when `ProjectRoot` was not populated by the loader, while `vibew bundle --build` correctly derived the cwd-basename — two different names from the same input. `deriveProjectName` is now the single project-name authority inside `runBundle`; its result is fed to both the image-tag default and the `--build` step.
+
 ### Changed
 
 - **`vibew init` no longer generates a placeholder `Dockerfile` or `.dockerignore`** (#1139). `AGENTS-VIBEWARDEN.md` now carries a Dockerfile contract checklist; agents and users write their own Dockerfile against that contract. Migration: existing projects that already have a Dockerfile are unaffected. New projects must write a Dockerfile after `vibew init` — see `AGENTS-VIBEWARDEN.md` §Dockerfile contract.
--- a/decisions/adr-093-bundle-image-name-cwd-basename-fallback.md
+++ b/decisions/adr-093-bundle-image-name-cwd-basename-fallback.md
@@ -1,0 +1,257 @@
+# ADR-093: bundle image-name resolution — cwd-basename fallback unified across `vibew bundle` and `--build`
+
+**Date**: 2026-04-23
+**Issue**: #1141
+**Status**: Accepted
+
+## Context
+
+`vibew bundle` and `vibew bundle --build` walk different project-name
+derivation chains when neither `name:` nor `app.image:` is set in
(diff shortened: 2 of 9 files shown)


tibtof:
**Reviewer Agent: APPROVED**

ADR-093 implemented correctly. `deriveProjectName` is the single authority inside `runBundle`; its result feeds both the default `imageTag` and `BuildOptions.ImageTag`. `ComposeProjectName()` is untouched — leaf fix as scoped. Guard test, table-driven regression, and `TestBuildService_Run_UsesPreResolvedImageTag` all present. Explicit empty-name error guard added. Standalone `vibew build` behavior unchanged.

PR #1171, opened 2026-04-27
feat(#1144): vibew validate catches real next-command failures

### LINKED ISSUE #1144: feat: vibew validate catches real failure modes (or drop the command)
## Problem

From the retro:

> `vibew validate` said "Configuration valid" while my project was about to fail bundling on image-name resolution (`vibewarden-app:latest` vs `qr-dali-app:latest`). A validator that doesn't catch a configuration that breaks the next command is decoration. Either make it actually catch real failure modes (missing `name:`, port mismatch between Dockerfile EXPOSE and `upstream.port`, image-tag drift, ACME-incompatible domain) or drop it. "Schema-valid YAML" is what your editor's language server already does.

## Proposal

Pick one direction and commit:

**Option A — make `vibew validate` useful**:

- Detect missing `name:` (or whatever D / #1141 settles on).
- Detect EXPOSE/`upstream.port` mismatch when a Dockerfile is present (overlaps with C2 / #1140's lint contract — bundle them).
- Detect image-tag drift between `.env` `VIBEWARDEN_APP_IMAGE` and the project-scoped tag the bundle would produce.
- Detect ACME-incompatible domain (e.g., `localhost`, IP literals, RFC 1918 ranges in `tls.domains` when `provider: letsencrypt`).
- Detect `WAF: log` in `vibewarden.production.yaml` (likely-wrong default for prod).
- Output uses the same glyph contract as F / #1143: `OK`, `OFF`, `FAIL`.

**Option B — drop the command**:

- Schema validation (which is what `validate` does today) is what the editor language server does already.
- Remove from CLI; re-purpose `vibew doctor` for the same checks.

## Recommendation

Option A. The command's discoverability is good ("validate") and several of the checks above are uniquely vibew's responsibility (image tag, ACME compat). Worth investing.

## Acceptance criteria (Option A path)

- `vibew validate` in qr-dali's project state (no `name:`) returns a `FAIL` with a clear remediation hint.
- Catches at least: missing name, EXPOSE/port mismatch, image-tag drift, ACME-incompatible domain.
- Tests for each rule.

- CLAUDE.md §Architecture principles → Artifact policy (see #1137 / PR #1152 — mostly-accurate validators are the worst middle ground).

## Related

- C2 / #1140 — Dockerfile lint check; overlaps and could be implemented as the same engine.

## Source

- `~/notes/vibewarden/retro-0.17.0.md`, agent additional drop #3.

tibtof:
Closes #1144

## Summary

- New package `internal/app/validate/` with 5 runtime checks run after strict schema validation:
  1. **name.go** — FAIL when `name:` unset and cwd dir is exactly `"vibewarden"` (tag collision)
  2. **dockerfile.go** — FAIL when last `EXPOSE` port in `Dockerfile` does not match `upstream.port`
  3. **imagetag.go** — FAIL when `.env` `VIBEWARDEN_APP_IMAGE` drifts from `<project>-app:latest`
  4. **acme.go** — FAIL when ACME provider configured with localhost, IP literal, or reserved TLD (`.local`, `.test`, etc.)
  5. **waf.go** — FAIL when WAF `mode: log` in production config; OK when `waf.acknowledge_log_mode: true`
- `DeriveProjectName` and `SanitiseProjectName` moved from `internal/cli/cmd/bundle.go` to `internal/app/bundle/projectname.go` (exported). The CLI shim and the new imagetag check both delegate to the same function — no drift possible.
- New config key `waf.acknowledge_log_mode: bool` (default false) added to `WAFConfig`.
- `validate.go` loads merged config (base + production overlay) for runtime checks so WAF and other production-only settings are visible.
- 5 fixture directories under `internal/cli/cmd/testdata/validate/`, one per FAIL scenario.
- Table-driven unit tests for each check; 6 integration tests via cobra command in-process.
- `vibew validate --help` long description updated; CHANGELOG, `llms-full.txt`, agents template, docs example, and `vibewarden.reference.yaml` updated.

## Test plan

- [ ] `make check` passes (confirmed locally — pre-push hook ran it)
- [ ] `go test ./internal/app/validate/...` — 29 unit test cases, all pass
- [ ] `go test ./internal/cli/cmd/... -run TestValidateRuntime` — 6 integration tests, all pass
- [ ] `go test ./internal/app/bundle/... -run TestDeriveProjectName -run TestSanitiseProjectName` — passes
- [ ] Manual: `vibew validate` on a fixture with a Dockerfile EXPOSE mismatch → exits 1 with FAIL row
- [ ] Manual: `vibew validate` with `waf.acknowledge_log_mode: true` in production override → exits 0 with OK row

### DIFF
--- a/internal/app/validate/acme.go
+++ b/internal/app/validate/acme.go
@@ -20,18 +20,6 @@ var acmeProviders = map[string]bool{
 	"letsencrypt-staging": true,
 }
 
-// rfc1918Nets is the list of RFC 1918 private IPv4 ranges.
-var rfc1918Nets []*net.IPNet
-
-func init() {
-	for _, cidr := range []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"} {
-		_, n, err := net.ParseCIDR(cidr)
-		if err == nil {
-			rfc1918Nets = append(rfc1918Nets, n)
-		}
-	}
-}
-
 // reservedTLDs is the set of TLD suffixes that ACME CAs will never issue for.
 // Includes RFC 6761 reserved names and the .local mDNS TLD.
 var reservedTLDs = map[string]bool{
--- a/internal/app/validate/acme.go
+++ b/internal/app/validate/acme.go
@@ -1,0 +1,114 @@
+package validate
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/vibewarden/vibewarden/internal/app/ops"
+	"github.com/vibewarden/vibewarden/internal/config"
+)
+
+// acmeProviders is the set of tls.provider values that use ACME (Let's Encrypt
+// and compatible CAs). These providers cannot issue certificates for IP
+// addresses, localhost, RFC 1918 addresses, or reserved TLDs.
+var acmeProviders = map[string]bool{
+	"letsencrypt":         true,
+	"zerossl":             true,
+	"buypass":             true,
+	"letsencrypt-staging": true,
+}
+
+// rfc1918Nets is the list of RFC 1918 private IPv4 ranges.
+var rfc1918Nets []*net.IPNet
+
+func init() {
+	for _, cidr := range []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"} {
+		_, n, err := net.ParseCIDR(cidr)
+		if err == nil {
+			rfc1918Nets = append(rfc1918Nets, n)
+		}
+	}
+}
+
+// reservedTLDs is the set of TLD suffixes that ACME CAs will never issue for.
+// Includes RFC 6761 reserved names and the .local mDNS TLD.
+var reservedTLDs = map[string]bool{
+	".local":     true,
+	".localhost": true,
+	".test":      true,
+	".invalid":   true,
+	".example":   true,
+}
+
+// CheckACME iterates the configured TLS domains and reports any that are
+// incompatible with ACME certificate issuance. Only fires when tls.provider is
+// one of the ACME providers (letsencrypt, zerossl, buypass, letsencrypt-staging).
+//
+// Per spec: checks cfg.TLS.Domain (singular). The config struct does not have a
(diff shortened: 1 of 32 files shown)


internal/app/validate/acme.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// rfc1918Nets is the list of RFC 1918 private IPv4 ranges.
+var rfc1918Nets []*net.IPNet
+
+func init() {
+	for _, cidr := range []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"} {
+		_, n, err := net.ParseCIDR(cidr)
+		if err == nil {
+			rfc1918Nets = append(rfc1918Nets, n)
```
tibtof: **Dead code + prohibited `init()` side effect**: `rfc1918Nets` is populated in `init()` but never read. RFC 1918 detection already works correctly because `net.ParseIP(d) != nil` returns true for any IP literal — private ranges included — so the slice is never consulted.

This violates two rules from CLAUDE.md:
- "No global state" — `rfc1918Nets` is a mutable package-level slice.
- "No `init()` side effects" — the `init()` function exists solely to populate it.

**Fix**: delete `var rfc1918Nets []*net.IPNet` and the entire `init()` block. No behaviour changes because the logic path that would use the slice does not exist.

tibtof:

One violation in `internal/app/validate/acme.go`:

`var rfc1918Nets []*net.IPNet` (line 24) and `func init()` (lines 26–32) are dead code — the slice is populated but never read. RFC 1918 detection is already handled correctly by `net.ParseIP(d) != nil`. This also violates CLAUDE.md's "no global state" and "no `init()` side effects" rules.

**Fix**: delete both declarations. No behaviour change. All other checks (architecture, tests, ACME classifier, Dockerfile parser, .env parser, imagetag drift, WAF key, `deriveProjectName` move) are correct and tests pass.

tibtof:
**Reviewer Agent: APPROVED**

re-review ok: no `init()` body, globals are immutable map literals — blocker resolved.

PR #1172, opened 2026-04-27
feat(#1140): vibew doctor Dockerfile lint — alpine, EXPOSE, HEALTHCHECK, USER, multi-stage, toolchain match

### LINKED ISSUE #1140: feat: vibew doctor / lint check that validates a user-written Dockerfile against the contract
## Problem

After dropping the placeholder Dockerfile (C1 / sibling issue), agents are expected to write their own. Without a feedback loop they'll get details wrong (alpine vs slim base, EXPOSE port mismatch, missing HEALTHCHECK, root user at runtime) and fail silently in `docker build` or worse, in production.

## Proposal

Add a check (preferably a new sub-check inside `vibew doctor`, or a dedicated `vibew lint` if doctor doesn't fit) that asserts the Dockerfile in cwd:

- Base image is alpine-based (so the sidecar healthcheck `wget`s without curl).
- `EXPOSE` value matches `upstream.port` from merged `vibewarden.yaml`.
- `HEALTHCHECK` directive present and probes the same port.
- Final stage does not run as `USER root` (warn, not block).
- Multi-stage build for Go / Rust / Java / etc. (warn if single-stage and binary is large).
- For Go: base image major.minor matches `go.mod`'s `go` directive (catches the 1.24 vs 1.26 case from the retro).

## Acceptance criteria

- New `vibew doctor` check (or `vibew lint` command) implemented as an outbound port + adapter (per ADR-064 / ADR-067).
- Failures provide actionable hints, not regex diagnostics (the retro flagged "go mod download exit 1" as the worst error of the run; the new check pre-empts that with "your Dockerfile uses golang:1.24 but go.mod requires 1.26 — bump the FROM line").
- Unit tests + integration test against a fixture Dockerfile.

- CLAUDE.md §Architecture principles → Artifact policy (originally seeded as #1137; closed via PR #1152).

## Depends on

- C1 (sibling issue) for the contract definition.

## Source

- `~/notes/vibewarden/retro-0.17.0.md`.

tibtof:
Closes #1140

## Summary

- New package `internal/app/dockerfile/` with `Parse()`, six rule helpers (`RuleAlpineBase`, `RuleExposeMatchesPort`, `RuleNoHealthcheck`, `RuleNonRootUser`, `RuleMultiStageForCompiled`, `RuleToolchainMatch`), and toolchain manifest detectors (Go/Node/Python).
- `internal/app/ops/doctor_dockerfile.go` wires `checkDockerfile()` into `DoctorService.runChecks()` as a new "Dockerfile" section. Section is omitted when no Dockerfile is present.
- `internal/app/validate/dockerfile.go` refactored to a thin shim over the shared parser (no duplicate logic).
- Non-root USER rule returns `SeverityWarn` (non-blocking per architect spec) — does not flip `allOK` to false.
- Toolchain-match check is the qr-dali regression guard: `go.mod` says `go 1.26` but `FROM golang:1.24-alpine` → FAIL with actionable hint.
- CHANGELOG `[Unreleased] ### Added`, `llms-full.txt` doctor check list, and both agents template files updated.

## Test plan

- `make check` passes (gofmt, golangci-lint 0 issues, full test suite).
- `internal/app/dockerfile/` — 20+ parse test cases, 6 rule tables, 17 toolchain tables.
- `internal/app/ops/doctor_dockerfile_test.go` — integration tests: no Dockerfile → section omitted; all-OK; each of the 5 FAIL checks; WARN-only fixture keeps `allOK=true`; JSON output includes `"section":"Dockerfile"`.
- `internal/app/validate/...` — existing EXPOSE shim tests unchanged and passing.


internal/app/ops/doctor_dockerfile.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	defer func() { _ = f.Close() }()
+
+	parsed, err := dockerfile.Parse(f)
+	if err != nil {
+		// Unparseable — silent skip.
+		return nil
+	}
+
+	tc, _, _ := dockerfile.DetectToolchain(projectRoot)
```
tibtof: **Code quality: swallowed error**

`tc, _, _ := dockerfile.DetectToolchain(projectRoot)` discards the error return, violating the project rule of never swallowing errors (`CLAUDE.md`: "Errors wrapped with context — never swallow errors").

A corrupt `go.mod` or unreadable `package.json` silently returns `Toolchain{}` here, meaning rules 5 and 6 both emit `SeverityOff` with no indication to the developer that their manifest is unreadable.

Fix: surface the error as a `SeverityWarn` check result instead of discarding it:

```go
tc, _, tcErr := dockerfile.DetectToolchain(projectRoot)
if tcErr != nil {
    out = append(out, withSection(CheckResult{
        Name:     "Dockerfile: toolchain detection",
        Severity: SeverityWarn,
        Detail:   fmt.Sprintf("could not read toolchain manifest: %v", tcErr),
    }, sectionDockerfile))
}
```

This keeps rules 5 and 6 in `Off` state (correct) while surfacing the read failure to the developer.

internal/app/ops/doctor.go
```diff
@@ -215,6 +215,12 @@ func (s *DoctorService) runChecks(ctx context.Context, cfg *config.Config, opts
 	results = append(results, withSection(s.checkUpstreamReachable(ctx, cfg), sectionLocalRuntime))
 	results = append(results, withSection(s.checkTLSCertValid(ctx, cfg, proxyHost, proxyPort), sectionLocalRuntime))
 
+	// --- Dockerfile contract checks ---
+	// Omitted entirely when no Dockerfile is present in the project root.
+	if workDir != "" {
+		results = append(results, s.checkDockerfile(ctx, workDir, cfg)...)
```
tibtof: **Code quality: dead-code guard**

The `if workDir != ""` condition at this line is never false. In `Run()` (15 lines above), `workDir` is already normalized:

```go
workDir := opts.WorkDir
if workDir == "" {
    workDir = "."
}
```

So by the time `runChecks` calls `checkDockerfile`, `workDir` is always at least `"."`. The guard provides false safety documentation (a reader infers there is a code path where `workDir` can be empty here, which is misleading).

The `checkDockerfile` function itself already has a redundant `if projectRoot == "" { return nil }` guard on its first line, so the double guarding is doubly unnecessary.

Fix: remove the `if workDir != ""` wrapper and call `checkDockerfile` unconditionally:

```go
// Dockerfile contract checks — omitted when no Dockerfile is present.
results = append(results, s.checkDockerfile(ctx, workDir, cfg)...)
```

tibtof:

2 issues require fixes before approval:

1. **Swallowed error** — `tc, _, _ := dockerfile.DetectToolchain(projectRoot)` in `internal/app/ops/doctor_dockerfile.go:43` silently discards manifest read errors. Per `CLAUDE.md`, errors must never be swallowed. Surface as a `SeverityWarn` row or at minimum capture and log it.

2. **Dead-code guard** — `if workDir != ""` in `internal/app/ops/doctor.go:220` is unreachable: `workDir` is normalized to `"."` unconditionally before `runChecks` is called. Misleads readers into thinking empty `workDir` is possible at that point. Remove the wrapper.

Everything else is solid: architecture is clean (dockerfile package imports only stdlib), all six rules implemented correctly, non-root USER correctly emits `SeverityWarn` (not `SeverityFail`), exit-code contract verified by test, fixture files not gitignored, validate shim correctly delegates to shared parser, no new dependencies.

tibtof:
**Reviewer Agent: APPROVED**

Both blockers resolved: `DetectToolchain` error is now logged via `slog.Warn` with graceful fallback to zero `Toolchain{}`; dead `workDir != ""` guard replaced by the correct `projectRoot == ""` guard inside `checkDockerfile`. Re-review clean.

PR #1168, opened 2026-04-27
chore(#1149): extract --observability flag into vibew obs up/down

tibtof:
Closes #1149.

## Summary

- Removes `--observability` flag from `vibew dev`.
- Adds new `vibew obs up` / `vibew obs down` subcommand pair that gates the Prometheus + Grafana stack behind a Docker Compose profile (`--profile observability`).
- Reuses `ports.Compose` adapter; no new ports.
- Reuses the existing observability compose template and scrape config — zero behavior change in what the obs stack does.

## Why

Per CLAUDE.md §Architecture principles → Artifact policy and the qr-dali deploy retro: trim CLI surface where features are heavyweight defaults that confuse the core loop. `vibew dev` is the day-loop verb; observability is opt-in tooling that belongs in its own subcommand pair, mirroring the `vibew dev` / `vibew down` symmetry.

## What changed

- New `internal/cli/cmd/obs.go` — cobra `obs` parent + `up` + `down`.
- New `internal/app/ops/obs.go` — `ObsService` with `Up` / `Down`, prints sidecar advisory if main stack isn't up.
- New `internal/app/ops/obs_test.go` — 14 table-driven tests covering both paths.
- New `internal/cli/cmd/obs_test.go` — wiring + help text tests.
- `internal/cli/cmd/dev.go` — removed `--observability` flag, godoc + Long updated.
- `internal/app/ops/dev.go` — removed `DevOptions.Observability` field and profile-passing branches.
- `internal/app/ops/dev_test.go` — observability-path tests deleted.
- `internal/cli/cmd/ops_test.go` — assertions updated.
- `internal/cli/cmd/root.go` — `NewObsCmd()` registered.
- `docs/observability.md`, `docs/getting-started.md`, `docs/architecture.md`, `docs/index.md` — invocation updated.
- `internal/cli/templates/agents/agents-vibewarden.md.tmpl` AND `docs/examples/AGENTS-VIBEWARDEN.md` — both mention the new commands; agree.
- `llms-full.txt` — CLI table row swap.
- `CHANGELOG.md` `[Unreleased]` — `### Removed` (flag) + `### Added` (commands).

## Test plan

- [x] `make check` passes
- [x] `vibew dev --observability` is no longer a flag (cobra unknown-flag error)
- [x] `vibew obs up` invokes compose with `--profile observability up -d`
- [x] `vibew obs down` invokes compose with `--profile observability down`
- [x] `vibew dev` (without the removed flag) keeps working unchanged

Issue #1149, opened 2026-04-26
chore: move --observability from vibew dev flag to a dedicated vibew obs up plugin command


tibtof:
## Problem

From the retro:

> `--observability` flag bundling Prometheus + Grafana into the dev stack. Heavyweight default. Most devs already have a stack or don't need one for local. Either make it a separate `vibew obs up` plugin or document it in passing — don't keep it as a flag on the core dev command. Flags are surface area.

Flags on `vibew dev` are noise for the 80% of users who never need them.

## Proposal

1. Move the observability stack into a dedicated subcommand: `vibew obs up` (and `vibew obs down` symmetric to `vibew down`).
2. Remove `--observability` from `vibew dev`.
3. `vibew obs up` brings up Prometheus + Grafana as a separate compose project that talks to the running sidecar via the metrics endpoint.
4. Document in `docs/observability.md` (already exists per #1129's docs entry).

## Acceptance criteria

- `vibew dev --observability` is no longer a flag (or aliases to `vibew obs up` with a deprecation notice).
- `vibew obs up` brings up the stack; `vibew obs down` tears it down.
- Existing `docs/observability.md` updated.
- AGENTS-VIBEWARDEN updated.
- CHANGELOG `### Changed` entry.

## Anchored by (policy reference)

- CLAUDE.md §Architecture principles → Artifact policy (originally seeded as #1137; closed via PR #1152).

## Source

- `~/notes/vibewarden/retro-0.17.0.md`, agent additional drop #5.

tibtof:
## Spec

### Problem statement

`vibew dev --observability` bundles Prometheus + Grafana into the core dev command as a flag, making the CLI surface heavier for the 80% of users who never need local observability. Flags are surface area; this one belongs behind a dedicated subcommand pair.

### User story

As a vibe coder, I want `vibew obs up` / `vibew obs down` to control the observability stack independently, so that `vibew dev` stays minimal and I can start or stop Prometheus + Grafana whenever I need them without restarting my dev session.

### Inventory (current state)

| Location | What it does |
|---|---|
| `internal/cli/cmd/dev.go` L106 | Registers `--observability` bool flag |
| `internal/app/ops/dev.go` `DevOptions.Observability` | Passes the flag to `compose up --profile observability` |
| `internal/config/templates/docker-compose.yml.tmpl` | Defines `profile: observability` for Prometheus, Loki, Promtail, OTel Collector, Jaeger, Grafana services |
[...]

PR #1182, opened 2026-04-28
fix(#1176,#1177): vibew obs up actually starts containers; vibew obs down only stops obs services

### LINKED ISSUE #1176: fix: vibew obs up silently no-ops on fresh projects (release blocker, v0.18.0)
**Severity: CRITICAL** — release blocker for v0.18.0.

## Symptom

\`vibew obs up\` exits 0 and prints success messages including the Grafana / Prometheus URLs, but launches zero containers. A user (or AI agent) following the docs visits \`http://localhost:3001\` and gets \`connection refused\` with no diagnostic path forward.

Reproduced on the v0.18.0-candidate smoke test (Go project, fresh \`vibew init\`):
- \`vibew dev\` succeeded; main stack up.
- \`vibew obs up\` → exit 0 → prints "Observability stack started" + URLs.
- \`docker compose -f .vibewarden/compose.yaml ps\` shows ONLY the main app + sidecar; no Grafana, no Prometheus.

## Root cause

\`vibew obs up\` calls \`docker compose up --profile observability\`. The generated \`docker-compose.yml\` only contains Grafana / Prometheus services when \`cfg.Observability.Enabled\` is true. The default generated config does not set this — \`observability\` is treated as opt-in via a config key the user must add manually.

Compose semantics: \`docker compose up --profile X\` succeeds silently when profile \`X\` matches no services in the file. No warning, no non-zero exit. The \`vibew obs up\` UX is sitting on top of this silent no-op.

## Affected: #1149 / PR #1168

Introduced when we extracted the \`--observability\` flag into the \`vibew obs up/down\` subcommand pair. The flag-mode probably had similar issues but they weren't caught because no smoke test exercised it end-to-end with a fresh project (the unit tests use mocked compose adapters).

## Fix options

1. **Always include observability services in the generated compose, profile-gated**: the \`docker-compose.yml\` generator unconditionally writes Grafana + Prometheus services tagged with \`profiles: [observability]\`. \`vibew obs up\` becomes "compose has the services; activate the profile." Cleanest. The services are inert until \`obs up\` runs.

2. **Detect-and-error**: \`vibew obs up\` reads the rendered compose, checks whether \`observability\` profile has any services, and fails with a clear hint ("add \`observability.enabled: true\` to vibewarden.yaml then run \`vibew dev\` to regenerate the compose"). Less clean — pushes work onto the user — but smaller code change.

Recommended: **Option 1**. Profiles are designed for this exact use case. The current "config flag gates compose generation" pattern is a leftover from before Compose v2 profiles existed.

## Acceptance criteria

- \`vibew init\` → \`vibew dev\` → \`vibew obs up\` brings up Grafana on \`localhost:3001\` and Prometheus on \`localhost:9090\` without ANY config edits.
- \`vibew obs up\` followed by \`docker compose -f .vibewarden/compose.yaml ps\` lists 4 containers (sidecar, app, grafana, prometheus).
- An end-to-end test in \`test/integration/\` verifies the full lifecycle.

## Discovery

V0.18.0-candidate smoke test, 2026-04-27.

### LINKED ISSUE #1177: fix: vibew obs down tears down main stack — compose --profile X doesn't scope down (release blocker, v0.18.0)
**Severity: CRITICAL** — release blocker for v0.18.0.

## Symptom

\`vibew obs down\` stops the entire compose project, not just the observability profile services. After running it, the main app + sidecar are also stopped, leaving the user with a fully dead stack.

Reproduced on the v0.18.0-candidate smoke test:
- \`vibew dev\` → 4 containers up (sidecar, app, grafana, prometheus — assuming BUG-1 fixed).
- \`vibew obs down\` → "Stopped 2 containers" / both app and vibewarden stopped.

## Root cause

\`docker compose down --profile observability\` does NOT scope the action to profile-only services. Compose's \`--profile\` flag selects which profile-gated services are *included* in the considered set, but \`down\` removes ALL services in the project regardless. The \`vibew obs down\` adapter call ends up nuking everything.

## Affected: #1149 / PR #1168

The PR added \`Profiles []string\` to \`ComposeDownOptions\` and threaded \`--profile observability\` to the down call. That fixed the test (which mocked the adapter) but the real-world Compose semantics make the flag a no-op for scoping.

## Fix options

1. **Stop + remove only the named obs services**: replace \`compose down --profile X\` with \`compose stop grafana prometheus\` followed by \`compose rm -f grafana prometheus\`. Explicit service list. Adapter needs to know which services belong to the obs profile.

2. **Use a separate compose project for obs**: bring up Grafana + Prometheus as their own \`docker compose -p <project>-obs\` invocation, share the network with the main project. Then \`compose -p <project>-obs down\` cleanly tears down only obs. Bigger change; we'd reverse the architect's "profile overlay on the existing compose project" decision from #1149.

Recommended: **Option 1** (smaller; keeps the architect's profile-overlay decision). The obs service list is known statically; passing it to compose stop/rm is straightforward.

## Acceptance criteria

- \`vibew dev\` → \`vibew obs up\` → \`vibew obs down\` leaves the main stack running.
- The down path stops + removes ONLY the obs services (grafana, prometheus, anything else gated by the observability profile).
- Integration test in \`test/integration/\` exercises the full lifecycle and asserts main containers are still running after \`obs down\`.

## Discovery

V0.18.0-candidate smoke test, 2026-04-27.

tibtof:
Closes #1176, closes #1177

## Summary

Two release-blocker bugs introduced by #1149 / PR #1168, surfaced by the v0.18.0-candidate smoke test.

**Bug #1176 — `vibew obs up` silently no-ops:**
The compose template wrapped all obs services in `{{- if .Observability.Enabled }}`. `docker compose --profile observability up -d` succeeds with zero containers when no service matches the profile. Fix: emit obs services unconditionally; they carry `profiles: [observability]` so they are inert until the profile is activated.

**Bug #1177 — `vibew obs down` nukes the main stack:**
`obs down` called `compose down --profile observability`. Compose's `--profile` is an activation flag for `up`, not a scope limiter for `down` — it removes ALL services in the project. Fix: service-targeted `compose stop + rm` against the static obs service list only.

## Changes

- `internal/ports/ops.go` — `ComposeDownOptions` gains `Services []string` and `VolumeNames []string`; removes `Profiles []string` (only caller was the broken path).
- `internal/adapters/ops/compose.go` — `Down` branches on `len(opts.Services)`; service-targeted path runs `stop + rm + best-effort volume rm`; full project path unchanged. New helpers: `downProject`, `downServices`, `isNoOpError`, `resolveProjectName`.
- `internal/app/ops/obs.go` — static `obsServices` and `obsVolumeNames` lists; `Down` rewired to use `Services`/`VolumeNames`; `printObsDownSummary` replaces the generic `printDownSummary` call.
- `internal/config/templates/docker-compose.yml.tmpl` — three `if .Observability.Enabled` guards removed (header comment, services block, volumes block). OTLP env vars guard kept (per ADR-097).
- `decisions/adr-097-obs-up-down-fix.md` — ADR authored by architect; committed here.
- `test/integration/obs_lifecycle_test.go` — new full-lifecycle integration test (gated by `//go:build integration`).
- Tests updated: replaced `TestObsService_Down_PassesObservabilityProfile` with `TestObsService_Down_PassesObsServices` + `TestObsService_Down_PassesObsVolumeNames` + drift test `TestObsServices_MatchTemplate`; updated generate service tests for always-present obs services.
- `docs/observability.md` — footnote updated to document `stop+rm` teardown.
- `CHANGELOG.md` — `[Unreleased] / Fixed` entry for both issues.

## Test plan

- [ ] `make check` passes (lint, build, race tests) — confirmed locally.
- [ ] `TestObsService_Down_PassesObsServices` asserts `Services == ["prometheus","loki","promtail","otel-collector","jaeger","grafana"]`.
- [ ] `TestObsService_Down_DoesNotPassProfiles` asserts `Profiles` is empty.
- [ ] `TestObsServices_MatchTemplate` drift test parses the template and asserts the static list matches.
- [ ] `TestGenerate_Observability_AlwaysPresent_WithProfile` asserts obs services always appear and OTLP vars are absent when disabled.
- [ ] Integration test `TestObsLifecycle_DownDoesNotNukeMainStack` (run with `-tags integration` on a machine with Docker and `vibew` on PATH).


internal/adapters/ops/compose.go
```diff
@@ -140,6 +158,90 @@ func (c *ComposeAdapter) Down(ctx context.Context, composeFile string, opts port
...
+			}
+			return ports.DownResult{}, fmt.Errorf("docker compose rm: %w", err)
+		}
+	}
+
+	result := parseDownOutput(stopStderr.String() + rmStderr.String())
+
+	// Step 3 (optional): remove named volumes best-effort.
+	if opts.Volumes && len(opts.VolumeNames) > 0 {
+		projectName := resolveProjectName(composeFile)
```
tibtof: **Bug: `resolveProjectName` returns `generated`, not the actual compose project name — `vibew obs down --volumes` silently never removes volumes.**

The compose file lives at `.vibewarden/generated/docker-compose.yml`. `filepath.Dir` of that path is `generated`, so `resolveProjectName` returns `"generated"`. But the compose file sets `name: {{ .ComposeProjectName }}` explicitly (e.g. `myapp`), so Docker names the volumes `myapp_prometheus-data`, not `generated_prometheus-data`. The `docker volume rm generated_prometheus-data` command will always get "no such volume" and silently skip — `RemovedVolumes` stays 0 and the user sees an incorrect summary.

The doc comment acknowledges the mismatch but marks it "best-effort fallback" — that framing understates the severity: this path is **always wrong** for the generated compose file.

Fix options (in order of preference):
1. Pass the project name from the app layer: add a `ProjectName string` field to `ComposeDownOptions`. `ObsService` already has access to the config and can call `cfg.ComposeProjectName()`.
2. Derive it from the compose `name:` field by parsing the YAML with a minimal struct before running stop/rm.

Either way, `resolveProjectName` should not be the live code path for the known compose file location.

internal/adapters/ops/compose_test.go
```diff
@@ -353,3 +353,96 @@ func TestCommandArgsConstruction(t *testing.T) {
...
+		// What matters is that a "no such service" scenario is tolerated.
+		// This test documents the intent; the real tolerance is verified in
+		// the unit test for isNoOpError below.
+		t.Logf("Down() returned error (acceptable with non-existent file): %v", err)
+	}
+}
+
+// TestIsNoOpError_RecognisesKnownMessages is a table-driven unit test for the
+// isNoOpError helper that classifies docker stderr messages as no-ops.
+func TestIsNoOpError_RecognisesKnownMessages(t *testing.T) {
```
tibtof, line 503: **Missing test: `TestComposeAdapter_Down_ServicesPath_VolumesCallsVolumeRm` required by ADR-097.**

The ADR test strategy explicitly lists this test. Its absence means the volume-removal path has no coverage, which is compounded by the `resolveProjectName` bug above (the test would have caught the wrong project name).

Add a unit test that:
1. Stubs `docker` to a fake that records invocations (or uses the `export_test.go` pattern).
2. Calls `Down` with `Services: ["grafana"]`, `Volumes: true`, `VolumeNames: ["grafana-data"]`, and a known compose file path whose parent directory name can be predicted.
3. Asserts `docker volume rm <project>_grafana-data` was invoked.

This test will also surface the `resolveProjectName` issue immediately.

internal/ports/ops.go
```diff
@@ -40,20 +40,38 @@ type ComposeUpOptions struct {
...
+	// `docker compose down`.
+	Services []string
+
+	// VolumeNames lists the named volumes to remove when Volumes is true and
+	// Services is non-empty. Each name is relative (e.g. "prometheus-data");
+	// the adapter resolves the full volume name using the compose project name.
+	// When VolumeNames is empty and Volumes is true with a non-empty Services
+	// list, no volumes are removed (the caller must supply the list explicitly).
+	VolumeNames []string
 }
```
tibtof, line 82: **Interface doc gap: `ComposeRunner.Down` does not mention the `Services`-path behavior.**

The `ComposeDownOptions` struct doc is thorough, but the `Down` method comment on the interface only says "opts controls --volumes and --remove-orphans" — nothing about service-targeted teardown or the `--profile` non-scope-limiter caveat.

ADR-097 explicitly required: "The adapter doc comment must call out: 'When Services is non-empty, this method performs stop + rm rather than down, because docker compose down --profile does not scope teardown by profile.'"

The implementation docs have this, but the **interface** is what callers read. Add to the `// Down` comment: 'When opts.Services is non-empty, teardown is scoped to those services via stop+rm rather than a full project down — docker compose --profile does not scope teardown by profile.'

tibtof:

Three issues found — one is a functional bug, one is a missing required test, one is an interface doc gap.

**1. Bug — `vibew obs down --volumes` silently never removes volumes** (inline on `internal/adapters/ops/compose.go` line 210)

`resolveProjectName` returns `generated` (the parent dir of `.vibewarden/generated/docker-compose.yml`). The compose file has `name: {{ .ComposeProjectName }}` set (e.g. `myapp`), so Docker names volumes `myapp_prometheus-data`. The `docker volume rm generated_prometheus-data` call always gets "no such volume" and silently skips. The doc marks this as "best-effort fallback" but it is always wrong for the generated compose path. Fix: add a `ProjectName string` field to `ComposeDownOptions`; have `ObsService` pass `cfg.ComposeProjectName()`. `resolveProjectName` can stay as a true last-resort fallback.

**2. Missing test — `TestComposeAdapter_Down_ServicesPath_VolumesCallsVolumeRm`** (inline on `internal/adapters/ops/compose_test.go` line 428)

ADR-097 test strategy explicitly lists this test. Without it the volume-removal path has zero coverage, and the project-name bug above would have been caught immediately.

**3. Interface doc gap — `ComposeRunner.Down` method comment** (inline on `internal/ports/ops.go` line 75)

The `Down` method doc on the interface says only "opts controls --volumes and --remove-orphans" — nothing about the `Services`-path or the `--profile` non-scope-limiter caveat. ADR-097 required this callout. The struct docs are thorough; the method doc needs one sentence added.

Everything else is correct: template guards removed properly, `obsServices` static list matches template, drift test present, `Profiles` fully removed, ADR-097 indexed, `vibew down` full-project path unaffected, OTLP env var gate preserved.

tibtof:
**Reviewer Agent: APPROVED**

Re-review confirmed all three blockers resolved:
1. Project name sourced from `cfg.ComposeProjectName()` in `internal/cli/cmd/obs.go` — no hardcoded strings.
2. `TestComposeAdapter_Down_ServicesPath_VolumesCallsVolumeRm` asserts correct `<project>_<volume>` construction.
3. `ComposeRunner.Down` interface doc in `internal/ports/ops.go` explicitly states `--profile` is not used on `Down`.

PR #1185, opened 2026-04-28
fix(#1184): obs config files written as files, not empty directories

### LINKED ISSUE #1184: fix(obs): observability config files generated as empty directories instead of files
## Bug report

**Found during**: v0.18.0 smoke test pass-2 (post-fix verification)

## What happened

After running \`vibew obs up\`, the file generator creates empty **directories** at paths that should be config **files**:

- \`.vibewarden/generated/observability/prometheus/prometheus.yml\` — is a directory
- \`.vibewarden/generated/observability/loki/loki-config.yml\` — is a directory

Docker Compose then fails to bind-mount these into the containers with:

\`\`\`
Error response from daemon: failed to create task for container: ... mount src=.../prometheus.yml, dst=/etc/prometheus/prometheus.yml: not a directory: Are you trying to mount a directory onto a file (or vice-versa)?
\`\`\`

## What was expected

Both paths should be regular files containing YAML configuration for Prometheus and Loki.

## Impact

- Prometheus, Loki, Grafana, Promtail, and otel-collector all fail to start.
- \`vibew obs up\` exits with error code 1 on every fresh project.
- Grafana is unreachable on port 3001.
- Jaeger starts successfully (it has no file-mount dependency).
- \`vibew obs down\` is unaffected (BUG-2 fix still works correctly).

## Reproduction

\`\`\`bash
mkdir /tmp/test-obs-bug && cd /tmp/test-obs-bug
vibew init --non-interactive
# create Dockerfile and app
vibew dev
vibew obs up   # fails with mount error
file .vibewarden/generated/observability/prometheus/prometheus.yml
# => .../prometheus.yml: directory
\`\`\`

## Severity

**Release blocker** for v0.18.0 — the observability feature is completely broken.

## Suggested fix

The generator should write YAML file content to these paths, not call \`os.MkdirAll\` with the full file path. A test asserting \`os.Stat(path).IsDir() == false\` for every generated file would catch this regression.

tibtof:
Closes #1184

## Summary

- Removes the `if cfg.Observability.Enabled` guard at `internal/app/generate/service.go:292` so `generateObservability` always runs, regardless of the config flag.
- Symmetric with PR #1182 which removed the same gate from the compose template. The mismatch caused Docker to auto-create bind-mount sources as empty directories, making all obs containers fail on start with a mount-type mismatch error.
- Adds `TestGenerate_Observability_FilesAreRegular_WhenDisabled` (table-driven, 7 paths) as the regression anchor — asserts each obs config path is a regular non-empty file when `Observability.Enabled: false`.
- Updates two stale tests (`TestGenerate_Observability_WhenDisabled`, `TestGenerate_OtelCollector_NotPresent_WhenDisabled`) that asserted the opposite (obs dir must not exist when disabled) to match the new contract.
- CHANGELOG `[Unreleased] ### Fixed` updated.

## Test plan

- `make check` passes (lint + build + all tests including new regression test).
- `TestGenerate_Observability_FilesAreRegular_WhenDisabled` exercises all 7 obs config paths with `Observability.Enabled: false` and asserts `IsRegular() == true` and `Size() > 0`.
- Revert `internal/app/generate/service.go` to the old guarded version and the new test fails — confirms it is a valid regression anchor.


tibtof:
Documentation review: docs consistent with code changes. Fixed two stale references in docs/ (getting-started.md line 416, configuration.md line 447) where 'when enabled' / 'Generate the observability stack' no longer held after the gate removal. Both now reflect that obs configs are always written.

tibtof:
**Reviewer Agent: APPROVED**

Gate removal is surgical and symmetric with PR #1182. `generateObservability` body unchanged. `cfg.Observability.Enabled` field retained. Compose template and obs adapter not touched. Regression test `TestGenerate_Observability_FilesAreRegular_WhenDisabled` covers all 7 paths with IsRegular + size > 0 assertions. `make check` passes — zero lint issues, all tests green including `test/architecture`.

PR #1206, opened 2026-04-28
fix(#1200): vibew bundle fails on image arch vs target platform mismatch

### LINKED ISSUE #1200: fix(retro-0.18.1): vibew bundle must fail on image-arch vs target-arch mismatch
## Retro evidence

> If I'd skipped `--platform` on an Apple Silicon host, the bundle would have shipped an arm64 image to an amd64 server and failed at runtime. The `--target-platform` flag exists; it should warn loudly when the image arch doesn't match.

## Problem

`vibew bundle` prints `Arch: linux/amd64` (or whatever) in the health block but doesn't compare against any declared target architecture. Apple Silicon dev → amd64 server is the silent-fail path: bundle succeeds, deploy succeeds, container fails to start at runtime with `exec format error`.

## Acceptance criteria

- [ ] `vibewarden.production.yaml` has `deploy.target_platform: linux/amd64` (default written by `vibew init`).
- [ ] `--target-platform` flag on `vibew bundle` overrides the yaml value.
- [ ] If the bundled image arch doesn't match the resolved target, bundle **fails** (not warns) with an actionable error: `image arch is linux/arm64, target is linux/amd64. Rebuild with: vibew build --platform linux/amd64`.
- [ ] `vibew init` includes `deploy.target_platform: linux/amd64` (visible default) in the production yaml it writes — as an actual field, not a comment.
- [ ] Test: bundle on macOS arm64 host with target amd64 (no `--platform` flag during build) → error.
- [ ] Test: bundle on macOS arm64 host with `vibew build --platform linux/amd64 && vibew bundle` → success.

## Scope notes

- Default is `linux/amd64` because Hetzner / most cloud VMs are amd64.
- Flag overrides yaml; both can be set.
- Do NOT auto-cross-build via buildx — that's scope creep. Just fail fast and tell the user the rebuild command.

## Severity

**HIGH.** Silent runtime failures after a successful `vibew bundle` + `docker compose up -d` are worse than misleading warnings.

## Source

tibtof:
Closes #1200

## Summary

- **New config field** `deploy.target_platform` (default `linux/amd64`) in `internal/config/deploy.go`. Added to `Config` struct with `mapstructure:"deploy"` tag and viper default so `LoadStrict` reflection accepts the field in production yaml files.
- **Init template** `internal/cli/templates/init-vibewarden.production.yaml.tmpl` now writes `deploy.target_platform: linux/amd64` as an active yaml entry (not a comment).
- **CLI flag** `--target-platform` default changed from `"linux/amd64"` to `""` (empty sentinel). Resolution order: flag (if set) → `cfg.Deploy.TargetPlatform` (yaml or viper default) → `"linux/amd64"`.
- **Hard failure** on arch mismatch. `ErrPlatformMismatch` sentinel added to `internal/app/bundle/image_health.go`. `runImageHealthCheck` renders the health block first (so the user sees Tag/Arch/Target side-by-side), then returns the error if `health.ArchMismatch` is true. No bundle files are written. CLI maps it to exit code 1.
- **Exact error copy** (pinned by `TestPlatformMismatchMessage_ExactWording`):
  ```
  image arch is linux/arm64, target is linux/amd64. Rebuild with: vibew build --platform linux/amd64
  Then re-run: vibew bundle
  ```
- **Docs**: `vibewarden.reference.yaml` + `docs/production-hardening.md` + `llms-full.txt` + `CHANGELOG.md` all updated.

## Multi-arch note

`docker image inspect` against a local image always returns the single architecture that was loaded into the local daemon (buildx `--load` writes only one arch). There is no manifest-list case for `vibew bundle`. This assumption is documented in the `ErrPlatformMismatch` godoc.

## Test plan

- `TestLoad_DeployTargetPlatform_DefaultIsAmd64` — viper default resolves correctly
- `TestLoad_DeployTargetPlatform_FromYAML` — yaml field overrides default
- `TestLoadStrict_DeployTargetPlatform_Accepted` — field accepted by strict loader
- `TestLoadStrict_DeployUnknownKey_Rejected` — unknown `deploy.foo` rejected
- `TestRunImageHealthCheck_ArchMismatch_ReturnsErrPlatformMismatch` — sentinel returned, health block rendered before error
- `TestBundle_ArchMismatch_NoFilesWritten` — no files written on mismatch
- `TestBundle_ArchMatch_Succeeds` — happy path unchanged
- `TestBundle_NoInspector_SkipsArchCheck` — nil-inspector path preserved
- `TestPlatformMismatchMessage_ExactWording` — exact error copy pinned
- `TestInitProject_ProductionYAML_HasDeployTargetPlatform` — two tests (template layer + scaffold integration layer)
- `TestBundleCmd_TargetPlatform_FlagRegistered` — flag registered with empty default
- `TestBundleCmd_NoYAML_NoFlag_DefaultsToAmd64` — default path
- `TestBundleCmd_YAMLTargetPlatform_AcceptedByStrict` — yaml field accepted end-to-end
- `TestBundleCmd_YAMLUnknownDeployKey_RejectedByStrict` — unknown sibling rejected

Architect's design comment: https://github.com/VibeWarden/vibewarden/issues/1200

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -3452,6 +3452,12 @@ func TestLoad_DeployTargetPlatform_DefaultIsAmd64(t *testing.T) {
 
 // TestLoad_DeployTargetPlatform_FromYAML verifies that deploy.target_platform
 // is populated when the yaml carries the field.
+//
+// Note on empty-string yaml: when the yaml contains `deploy.target_platform: ""`
+// (explicit empty string), viper returns "" — the explicit empty overrides the
+// viper default. The caller (CheckImageHealth / runBundle) is responsible for
+// treating "" as "use defaultTargetPlatform". This test documents that
+// config.Load returns "" in that case so callers are not surprised.
 func TestLoad_DeployTargetPlatform_FromYAML(t *testing.T) {
 	tests := []struct {
 		name     string
@@ -3467,6 +3473,14 @@ func TestLoad_DeployTargetPlatform_FromYAML(t *testing.T) {
 			name:     "amd64 explicit in yaml",
 			yaml:     "deploy:\n  target_platform: linux/amd64\n",
 			wantPlat: "linux/amd64",
+		},
+		{
+			// Explicit empty string in yaml: viper returns "" (does not fall
+			// back to the default "linux/amd64"). CheckImageHealth treats ""
+			// as defaultTargetPlatform; config.Load must not silently hide it.
+			name:     "empty string in yaml returns empty",
+			yaml:     "deploy:\n  target_platform: \"\"\n",
+			wantPlat: "",
 		},
 	}
 	for _, tt := range tests {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -3431,3 +3431,70 @@ func TestValidate_RolePaths(t *testing.T) {
 		})
 	}
 }
+
+// TestLoad_DeployTargetPlatform_DefaultIsAmd64 verifies that when no yaml
+// contains deploy.target_platform the viper default resolves to linux/amd64.
+func TestLoad_DeployTargetPlatform_DefaultIsAmd64(t *testing.T) {
+	dir := t.TempDir()
+	cfgPath := filepath.Join(dir, "vibewarden.yaml")
+	if err := os.WriteFile(cfgPath, []byte("server:\n  port: 8443\n"), 0o600); err != nil {
+		t.Fatalf("writing config: %v", err)
+	}
+
+	cfg, err := config.Load(cfgPath)
+	if err != nil {
+		t.Fatalf("Load() error = %v", err)
+	}
+	if cfg.Deploy.TargetPlatform != "linux/amd64" {
+		t.Errorf("Deploy.TargetPlatform = %q, want %q", cfg.Deploy.TargetPlatform, "linux/amd64")
+	}
+}
+
+// TestLoad_DeployTargetPlatform_FromYAML verifies that deploy.target_platform
+// is populated when the yaml carries the field.
+func TestLoad_DeployTargetPlatform_FromYAML(t *testing.T) {
+	tests := []struct {
+		name     string
+		yaml     string
+		wantPlat string
+	}{
+		{
+			name:     "arm64 from yaml",
+			yaml:     "deploy:\n  target_platform: linux/arm64\n",
+			wantPlat: "linux/arm64",
+		},
+		{
+			name:     "amd64 explicit in yaml",
+			yaml:     "deploy:\n  target_platform: linux/amd64\n",
+			wantPlat: "linux/amd64",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			dir := t.TempDir()
+			cfgPath := filepath.Join(dir, "vibewarden.yaml")
+			if err := os.WriteFile(cfgPath, []byte(tt.yaml), 0o600); err != nil {
+				t.Fatalf("writing config: %v", err)
+			}
+			cfg, err := config.Load(cfgPath)
+			if err != nil {
+				t.Fatalf("Load() error = %v", err)
+			}
+			if cfg.Deploy.TargetPlatform != tt.wantPlat {
+				t.Errorf("Deploy.TargetPlatform = %q, want %q", cfg.Deploy.TargetPlatform, tt.wantPlat)
+			}
+		})
+	}
+}
+
+// TestLoad_DeployBlock_MissingEntirelyUsesDefault verifies that when the yaml
+// contains no deploy: block at all the viper default (linux/amd64) applies.
+func TestLoad_DeployBlock_MissingEntirelyUsesDefault(t *testing.T) {
+	cfg, err := config.Load("")
+	if err != nil {
+		t.Fatalf("Load(\"\") error = %v", err)
+	}
+	if cfg.Deploy.TargetPlatform != "linux/amd64" {
+		t.Errorf("Deploy.TargetPlatform = %q, want %q", cfg.Deploy.TargetPlatform, "linux/amd64")
+	}
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ yet completed (boot gap of ~5–10s). After that it is always `"ok"` or `"failin
 
 ### Fixed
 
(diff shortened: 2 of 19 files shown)


internal/config/config_test.go
```diff
@@ -3431,3 +3431,70 @@ func TestValidate_RolePaths(t *testing.T) {
...
+			cfg, err := config.Load(cfgPath)
+			if err != nil {
+				t.Fatalf("Load() error = %v", err)
+			}
+			if cfg.Deploy.TargetPlatform != tt.wantPlat {
+				t.Errorf("Deploy.TargetPlatform = %q, want %q", cfg.Deploy.TargetPlatform, tt.wantPlat)
+			}
+		})
+	}
+}
```
tibtof, line 3502: **Missing test case — empty-string yaml field**: The architect's design explicitly requires a test for `deploy.target_platform: ""` (empty string in yaml) falling through to `linux/amd64`. This case is not in `TestLoad_DeployTargetPlatform_FromYAML`.

Why it matters: viper's `SetDefault` does NOT override an explicit empty string set in YAML — when the yaml contains `target_platform: ""`, `cfg.Deploy.TargetPlatform` will be `""`. The resolution then falls through to `CheckImageHealth`'s own empty-check (not viper), which correctly defaults to `linux/amd64`. The end result is correct, but it's via a different path than the code comments imply. The test is needed to prevent a future refactor from breaking this path silently.

Add to `TestLoad_DeployTargetPlatform_FromYAML`:

```go
{
    name:     "empty string in yaml falls to default",
    yaml:     "deploy:\n  target_platform: \"\"\n",
    wantPlat: "linux/amd64",  // viper default applies when yaml value is empty
},
```

Note: viper will actually set the field to `""` and NOT apply the default. The default is applied by `CheckImageHealth`. The test needs to go at the integration level (calling `Bundle` with an inspector) rather than config load level — or document the actual behavior: the config layer returns `""`, and the bundle layer catches it.

tibtof:

Two required fixes before merge. The core behavior (sentinel error, no files written, exit code 1, resolution order, init template) is correct. The issues are a stale godoc and a missing test case that the architect explicitly mandated.

---

**Issue 1 — Stale godoc in `RenderImageHealth`** (file: `internal/app/bundle/image_health.go`, line 188)

The godoc example for `RenderImageHealth` still shows:
```
//	    - image arch linux/arm64 != target linux/amd64 (rebuild: vibew build --platform linux/amd64)
```
This line no longer appears in `RenderImageHealth` output — `collectWarnings` does not emit the arch-mismatch line since this PR converted it to a hard error. On an arch mismatch, `RenderImageHealth` is called by `runImageHealthCheck` to render the block (Tag/Arch/Target/etc.) and then the error is returned separately. The rendered block will show `Warnings: none` (or only the stale warning if stale). Remove or replace the stale example line.

**Fix:** Remove line 188 from the godoc example, or replace the `- image arch...` line with a comment explaining that on arch mismatch the block is rendered first and then `ErrPlatformMismatch` is returned by the caller — the block itself does not include the rebuild command.

---

**Issue 2 — Missing test for `deploy.target_platform: ""` (empty-string yaml)** (file: `internal/config/config_test.go`, inline comment posted)

The architect's design explicitly states: "yaml present with `deploy.target_platform: ""` (empty string) → fall through to viper default. Document this in the test."

This test case is absent. The behavior is currently correct — an empty string falls through to `linux/amd64` via `CheckImageHealth`'s own empty-check — but the path is not via viper as the `runBundle` comment implies. A future refactor removing `CheckImageHealth`'s `if target == ""` guard would silently break this.

**Fix:** Add an empty-string test case to `TestLoad_DeployTargetPlatform_FromYAML` (documenting that config.Load returns `""` for explicit empty yaml) AND add a bundle-level test showing that `Bundle` with `TargetPlatform: ""` resolves to `linux/amd64` behavior (i.e., does not mismatch against an amd64 image). The `runBundle` comment that says "cfg.Deploy.TargetPlatform is always populated by the viper default" is inaccurate for the empty-string case — fix the comment to be precise.

---

**Non-blocking observation** (no fix required, flagged for awareness):

`internal/cli/templates/agents/agents-vibewarden.md.tmpl` line 194 still reads:
```
Exit codes: 0=success (with or without warnings), 2=image missing, 3=docker daemon unreachable.
```
The PR adds exit code 1 for arch mismatch. The architect called out "Add one line explaining yaml fallback" to this template. The `docs/examples/AGENTS-VIBEWARDEN.md` (pre-rendered) was updated in commit `08eec14`, but the template that `vibew init` uses to generate new projects' agent guides was not. New users initialized after this change will get an outdated agents guide with no mention of exit 1 (arch mismatch). This should be fixed — I'm marking it non-blocking only because it does not affect the correctness of the feature itself, but it is a CLAUDE.md violation (docs-runtime drift is a known failure mode). Recommend fixing it in this PR.

tibtof:
@reviewer addressed: godoc fixed, empty-string test added, init template exit-codes updated. Ready for re-review.

tibtof:
**Reviewer Agent: APPROVED**

Re-review of commit `6a9c528` (third commit — new commit, not amend, confirmed).

---

**Finding 1 — Stale godoc: RESOLVED**

`internal/app/bundle/image_health.go` lines 177–195: the old arch-mismatch warning line is gone. Replaced with accurate prose describing the warn→fail split and a corrected stale-only example block. The godoc now correctly states that `RenderImageHealth` renders the block and the caller returns `ErrPlatformMismatch` separately.

**Finding 2 — Missing tests: RESOLVED**

Both test functions are present and correct:

- `TestCheckImageHealth_EmptyStringTargetPlatform` (`internal/app/bundle/image_health_test.go` line 210): passes `TargetPlatform: ""`, asserts `h.Target == "linux/amd64"` and `h.ArchMismatch == false`. Correctly pins the guard path.
- `TestLoad_DeployTargetPlatform_FromYAML` (`internal/config/config_test.go` line 3461): table-driven with three cases — arm64 explicit, amd64 explicit, and empty-string-in-yaml returns `""`. The empty-string case is documented precisely (viper does not fall back to default on explicit empty; caller handles it). Companion test `TestLoad_DeployBlock_MissingEntirelyUsesDefault` (line 3504) covers the absent-field path.

**Finding 3 — Init template exit codes: RESOLVED**

`internal/cli/templates/agents/agents-vibewarden.md.tmpl` line 194 now reads:
`Exit codes: 0=success (with or without warnings), 1=image arch does not match target platform (rebuild with `vibew build --platform <target>`), 2=image missing, 3=docker daemon unreachable.`

**runBundle comment: RESOLVED**

`internal/cli/cmd/bundle.go` lines 198–204: the comment now accurately states that viper returns `""` for explicit-empty yaml (does not fall back to default) and that `CheckImageHealth`'s own guard handles it. The inaccurate "always populated" wording is gone.

---

**CI status (as of review time):** `Build & Test`, `Lint`, `Integration Tests`, `Semgrep SAST`, `Trivy`, and `CodeQL` are still in progress. No failures reported. Do not merge until all checks complete.

**No new findings.** Architecture, error handling, test structure, and template are all correct.

PR #1207, opened 2026-04-28
fix(#1199): docker compose project name consistent local vs remote

### LINKED ISSUE #1199: fix(retro-0.18.1): docker compose project name inconsistent local vs remote
## Retro evidence

> Locally the docker-compose prefix is the directory name (`qr-code-blackhole-*`); on the server after deploying it became `vibewarden-app-*`. Same project, two identities.

## Problem

Locally `vibew dev` runs `docker compose` from cwd → compose defaults the project name to the directory name. On the server, the bundled `docker-compose.yml` has a hardcoded top-level `name: vibewarden-app`. Same project, two identities. Container/network names diverge between environments. Breaks `docker ps | grep <app>` debugging.

## Acceptance criteria

- [ ] Both the dev compose template and the bundled compose include explicit `name: <app.name>` (where `<app.name>` is the value from `vibewarden.yaml`).
- [ ] `app.name` is the canonical source. Fall back to dirname only if `app.name` is empty (today's `vibew init` always populates it).
- [ ] Search the codebase for the literal string `vibewarden-app`. Anything depending on it (sidecar config network refs, future `vibew remote *`, etc.) must read from `app.name`.
- [ ] Test: `docker ps` after `vibew dev` shows containers prefixed with `<app.name>-`.
- [ ] Test: `docker ps` on the remote after deploy shows containers prefixed with `<app.name>-`.

## Severity

**HIGH.** Not a crash, but actively misleads ops/agents debugging the deployed app.

## Source

tibtof:
Closes #1199

## Root cause

`vibew init` did not populate `app.name` unless `--name` was passed. Without `name:` set, `ComposeProjectName()` resolution fell through differently per environment:

- **Local (`vibew dev`):** `cfg.Name=""`, `cfg.ProjectRoot` set from config file dir → resolved to `dirname` (e.g. `qr-code-blackhole`)
- **Remote bundle:** `cfg.ProjectRoot` not set from `loadAndResolve`; old `App.Image` branch yielded `vibewarden-app` for legacy configs

Two separate resolver chains (`ComposeProjectName()` and `DeriveProjectName()`) with subtly different fallbacks compounded the problem.

## What changed

1. **`init-vibewarden.yaml.tmpl`** — dropped the `{{ if .Name }}` guard; `name:` is always rendered
2. **`vibewarden.yaml.tmpl`** (wrap template) — added `name: "{{ .ProjectName }}"` before `server:`
3. **`init_project.go`** — when `opts.Name == ""`, defaults to `opts.ProjectName` (dirname) before rendering
4. **`config.go:ComposeProjectName()`** — removed the `App.Image` derivation branch. Chain is now: `Name` → `dirname(ProjectRoot)` → `"vibewarden"`
5. **`projectname.go:DeriveProjectName()`** — collapsed to `SanitiseProjectName(cfg.ComposeProjectName())`; one canonical resolver
6. **`bundle.go:runBundle()`** — sets `cfg.ProjectRoot` from the config file dir when empty, enabling the dirname fallback for legacy projects
7. **`build.go:resolveImageTag()`** — same: populates `cfg.ProjectRoot` from `workDir` before calling `ComposeProjectName()`
8. **`bundle_extras.go:renderBundleReadme()`** — orphan-cleanup paragraph added for upgraders

## Upgrade note

> Existing remote deployments running under the old project name `vibewarden-app` should run `docker compose -p vibewarden-app down` ONCE on the remote before bringing up the new stack. The bundle README includes this note.

## Test plan

- `TestInitCmd_NoNameFlag_DefaultsToDirname` — without `--name`, yaml has `name: <dirname>`
- `TestInitProject_NoNameFlag_DefaultsToDirname` — same at service layer
- `TestComposeProjectName` — updated table (removed `App.Image` cases, added dirname cases)
- `TestDeriveProjectName` — updated table + new `TestDeriveProjectName_MatchesComposeProjectName` parity test
- `TestBundle_ComposeProjectName_MatchesAppName` — real generator: compose `name:` matches `cfg.Name`
- `TestBundle_Extras_Readme_OrphanCleanupHint` — README contains the upgrade note
- `TestRunBundle_ImageTagDerivation` — updated: `app.image` no longer drives project name; dirname does

Architect's design comment: https://github.com/VibeWarden/vibewarden/issues/1199#issuecomment-1


internal/app/ops/build.go
```diff
@@ -142,24 +142,31 @@ func (s *BuildService) probeShell(ctx context.Context, image string, out io.Writ
...
+	abs, err := filepath.Abs(workDir)
+	if err != nil {
+		return "", fmt.Errorf("resolving work directory: %w", err)
+	}
+
 	if cfg != nil {
+		// Populate ProjectRoot from workDir so that cfg.ComposeProjectName()
+		// can use the dirname fallback for projects without name: set.
+		if cfg.ProjectRoot == "" {
+			cfg.ProjectRoot = abs
```
tibtof: **Architectural concern: `resolveImageTag` mutates the caller's `cfg` via pointer.**

Setting `cfg.ProjectRoot = abs` at this line mutates the caller's `*config.Config` struct. Any subsequent call to `ComposeProjectName()` in the same request will see the mutated `ProjectRoot` value — even if the caller did not intend it. This is an implicit side effect that makes the function harder to reason about in isolation.

The immediate behaviour is correct (the caller does use the mutated value), but the pattern is fragile. Consider one of two fixes:
1. Copy the config before mutating: `cfgCopy := *cfg; cfgCopy.ProjectRoot = abs; name := cfgCopy.ComposeProjectName()`
2. Add a doc comment above the mutation block noting the intentional side effect.

Option 1 is preferred since `resolveImageTag` is a private helper and the extra allocation is trivial.

tibtof:

## Summary

The core logic is correct: template guard removed, wrap template gains `name:`, `ComposeProjectName()` loses the `App.Image` branch, `DeriveProjectName()` collapses to a sanitising wrapper, bundle README gets the orphan-cleanup paragraph. All 9 checklist items from the review brief are implemented. Tests are well-structured and the new assertions are correct inversions of the old ones.

Three issues require fixes before approval.

---

### Finding 1 — BLOCKING: `validate.go` line 172 — missed callsite for `cfg.ProjectRoot`

`internal/cli/cmd/validate.go:172` calls `cfg.ComposeProjectName()` without setting `cfg.ProjectRoot` first. `ProjectRoot` is only assigned at line 185 (`projectRoot := filepath.Dir(absCheck)`) and is never written back to `cfg`.

For a legacy project with no `name:` in `vibewarden.yaml`, `ComposeProjectName()` now falls through to `"vibewarden"` (last resort). The migration hint then emits:

```
sed -i 's/vibewarden-app:latest/vibewarden-app:latest/g' .env
```

A no-op command printed as actionable guidance. This is the one callsite the dev missed; `bundle.go` and `build.go` both set `ProjectRoot` before calling `ComposeProjectName()`. Fix:

```go
// internal/cli/cmd/validate.go — before detectLegacyAppImage() at ~line 168
if cfg.ProjectRoot == "" {
    cfg.ProjectRoot = filepath.Dir(absCheck)  // absCheck is already resolved above
}
```

---

### Finding 2 — BLOCKING: `resolveImageTag` mutates the caller's `*config.Config` via pointer (inline comment on `build.go:161`)

The mutation `cfg.ProjectRoot = abs` is an implicit side effect on the caller's struct pointer. The current callers happen to be fine, but any future caller that passes a shared `*config.Config` and calls `ComposeProjectName()` after `Run()` will see a mutated `ProjectRoot`. This is the exact class of bug that caused the pre-existing confusion between resolver chains.

Fix: copy before mutating in the private helper:

```go
cfgCopy := *cfg
cfgCopy.ProjectRoot = abs
name := cfgCopy.ComposeProjectName()
if name != "" && name != "vibewarden" {
    return name + "-app:latest", nil
}
```

---

### Finding 3 — Advisory: `ComposeProjectName()` branch 1 does not sanitize `cfg.Name`

If a user writes `name: "My App"` in `vibewarden.yaml`, `ComposeProjectName()` returns `"My App"` verbatim (uppercase + space). Docker Compose requires `[a-z0-9_-]+` for project names. The bundle path adds `SanitiseProjectName` on top via `DeriveProjectName`, which strips the space and preserves case — yielding `"MyApp"`, which is still not lowercase-valid.

This pre-dates the PR, but making `cfg.Name` the unconditional canonical source elevates the risk. At minimum, `ComposeProjectName()` should apply `sanitizeProjectName` to branch 1 the same way branch 2 does. Otherwise the generated `docker-compose.yml` will carry an invalid project name for any user who writes a mixed-case `name:`.

This is advisory (not a new regression introduced by this PR), but should be addressed before v0.19.0 ships.

---

### Architectural question: per-callsite vs. centralized `ProjectRoot` injection

The architect suggested centralizing in `loadAndResolve`. The dev chose per-callsite, which is why `validate.go` was missed. With three separate callsites now independently setting `cfg.ProjectRoot` (`bundle.go`, `build.go`, `generate/service.go`), the probability of a fourth callsite forgetting is high. If centralizing in `loadAndResolve` is impractical (e.g. config path isn't always known there), at least document the invariant in a comment on the `ProjectRoot` field so future devs know to set it.

---

### What's correct

- Template guard removal in `init-vibewarden.yaml.tmpl`: correct.
- `vibewarden.yaml.tmpl` (wrap) now has `name: "{{ .ProjectName }}"` and `ProjectName` is set to `filepath.Base(filepath.Clean(dir))` in `service.go:119`: correct.
- `init_project.go` defaults `opts.Name` to `opts.ProjectName` when empty: correct.
- `ComposeProjectName()` chain is Name → dirname → "vibewarden" with App.Image branch removed: correct.
- `DeriveProjectName()` collapses to `SanitiseProjectName(cfg.ComposeProjectName())`: correct.
- Bundle README orphan-cleanup paragraph: correct. `docker compose -p vibewarden-app down` on a host that never had the project exits 0 (no-op).
- `TestInitCmd_NoNameFlag_DefaultsToDirname`: assertion correctly inverted from old test.
- `TestDeriveProjectName_MatchesComposeProjectName`: good parity guard.
- Branch name `fix/1199-compose-project-name`, PR targets `main`, conventional commit title: all correct.
- No new dependencies.

tibtof:
@reviewer addressed: validate callsite centralized in loader (also removes the build.go pointer-mutation bug); branch 1 sanitization added with test; @writer's doc commits unchanged. Ready for re-review.

tibtof:
**Reviewer Agent: APPROVED**

Re-review of commit `b869711` and the full commit stack (3 commits total on `fix/1199-compose-project-name`).

All four findings from the first review are resolved. `make check` passes clean.

---

### Finding-by-finding confirmation

**Finding 1 (BLOCKING) — validate.go missed callsite.**
Resolved by centralizing in `loadInternal`. `validate.go:124` calls `config.LoadStrict`, which calls `Load`, which calls `loadInternal(configPath, true)`. `loadInternal` now sets `cfg.ProjectRoot = filepath.Dir(abs)` after `ReadInConfig` whenever `v.ConfigFileUsed()` is non-empty. The migration hint at `validate.go:172` therefore calls `cfg.ComposeProjectName()` with `ProjectRoot` already populated. Confirmed by reading `load_raw.go:72-76` and tracing the call chain.

**Finding 2 (BLOCKING) — resolveImageTag pointer mutation.**
Resolved. `build.go:165` does `cfgCopy := *cfg` (struct copy). The conditional fill is on `cfgCopy.ProjectRoot`, not on the caller's pointer. The caller's `*config.Config` is never mutated. Comment at line 160–164 explains the invariant correctly.

**Finding 3 (BLOCKING-MINOR) — ComposeProjectName branch 1 unsanitized.**
Resolved. `config.go:195` returns `sanitizeProjectName(c.Name)`. Three subtests pin the behavior: `"My App" → "my-app"`, `"MyProject" → "myproject"`, `"foo_bar!" → "foo-bar"`. Traced manually — all correct.

**Finding 4 (Architectural) — centralization.**
Fully addressed. `loadInternal` is the single authority. `generate/service.go:120-125` retains a pre-existing `if cfg.ProjectRoot == ""` guard (pre-dates this PR, no change); it now acts as defense-in-depth and only fires for manually constructed `Config` values (no loader involved), which is the correct pattern matching `build.go`.

---

### Brief verification points

1. **loadInternal sets ProjectRoot exactly once** — confirmed at `load_raw.go:72-76`. No "only when not already set" guard needed: `cfg` is created fresh inside `loadInternal`, so no caller can have set `ProjectRoot` before this line. The doc comment on the field states "do not set this field manually".

2. **bundle.go redundant block removed** — confirmed from commit `b869711` stat: `bundle.go` has 9 deletions, no new `ProjectRoot` assignment.

3. **build.go struct copy** — confirmed at `build.go:165`: `cfgCopy := *cfg` with `cfgCopy.ProjectRoot = abs` only when empty.

4. **validate.go no explicit ProjectRoot assignment** — confirmed. Only the loader sets it. `TestValidateCmd_MigrationHint_UsesProjectDirname` passes a config file path, loader sets `ProjectRoot`, hint emits `mylegacyapp-app:latest`. Correct.

5. **ComposeProjectName branch 1 sanitization** — `sanitizeProjectName` lowercases and replaces non-`[a-z0-9-]` with hyphens. Underscores → hyphens in both branch 1 and branch 2 (both go through `sanitizeProjectName`). Intentional and consistent. Docker Compose allows underscores but the choice is conservative and uniform.

6. **ProjectRoot doc comment** — `config.go:136-141` updated: states "Set by loadInternal (and therefore by Load, LoadRaw, and LoadStrict)", "Callers of ComposeProjectName() MUST use the Config returned by a loader function — do not set this field manually."

7. **Test count +4** — 3 new subtests in `TestComposeProjectName` (`config_test.go:3235-3237`) + 1 new function `TestValidateCmd_MigrationHint_UsesProjectDirname` (`validate_test.go:785`). Confirmed.

8. **Adversarial name double-sanitization** — `ComposeProjectName()` returns `"myproject-----rm--rf"` (hyphens from `sanitizeProjectName`); `SanitiseProjectName` passes hyphens through (bundle-layer allowlist includes `'-'`). Result is `"myproject-----rm--rf"` — matches test expectation. No surprising output.

---

### make check

Passes clean. All packages `ok`. No lint errors.

---

### No new concerns

The centralization in `loadInternal` is clean. The one remaining pre-existing mutation in `generate/service.go:120-125` is now defense-in-depth, not the primary setter, and is guarded with `if cfg.ProjectRoot == ""` — consistent with the pattern established in `resolveImageTag`. No unintended side effects on test loaders: all tests that need specific `ProjectRoot` values construct `&config.Config{ProjectRoot: ...}` directly without going through a loader.

Issue #1203, opened 2026-04-28
feat(retro-0.18.1): canonical agent kickoff prompt in repo + vibew prompt-template CLI


tibtof:
## Retro evidence

The qr-code-blackhole agent's confusion in the v0.18.1 retro ("docs say the bundle contains a `deploy.sh`") turned out NOT to be a vibew docs issue. The official docs are correct. The bad `bash deploy.sh` instruction came from the **human user's hand-rolled kickoff prompt** — which is outside the repo, version-drifted, and re-pasted to every new agent task.

Three retros now (qr-dali on v0.17.0, qr-code-blackhole on v0.18.1) have hit version drift in the kickoff instructions. A versioned canonical prompt closes the gap permanently.

## Problem

The "spin up VibeWarden with an agent" prompt template has no canonical home. Each user maintains it locally, and it rots silently every time the deploy contract or CLI surface changes.

## Acceptance criteria

- [ ] New canonical prompt content lives in the repo (e.g. as a section embedded in `llms-full.txt` so it's discoverable via the URL the kickoff prompt itself references — `vibewarden.dev/llms-full.txt`).
- [ ] Optionally also exposed as `docs/agent-kickoff.md` for direct linkability.
- [ ] New CLI command: `vibew prompt-template [--deploy] [--name <prjname>] [--describe "<desc>"] [--domain <fqdn>]` prints the kickoff prompt to stdout, generated from the current binary's knowledge — so the prompt always matches the binary version.
- [ ] **Two flavors:**
  - **Dev only** (default): install + `mkdir <name> && cd <name>` + `vibew init --name <prjname> --describe "<desc>"` + `vibew add tls --domain <fqdn>` + `vibew dev`
  - **Dev + deploy** (`--deploy` flag): adds the bundle / scp / ssh / `docker compose up -d` / healthcheck steps using the current contract from the bundle README (companion issue).
- [ ] Init step in the prompt **always uses both `--name` and `--describe` flags explicitly** (per user direction during retro triage).
- [ ] No `deploy.sh` references. Use the literal command sequence from the bundle README.
- [ ] Test: `vibew prompt-template --deploy --name foo --describe "bar" --domain demo.example.com` produces a copy-pasteable agent prompt that takes the agent zero-to-deployed without any other docs.
- [ ] ADR documenting the decision: vibew owns the canonical agent kickoff prompt (binary-versioned, regenerable, stdin-pasteable).

## Severity

[...]

PR #1209, opened 2026-04-28
feat(#1204): bundle README command-first + auto MANIFEST.md + stdout deploy commands

### LINKED ISSUE #1204: feat(retro-0.18.1): bundle README command-first + vibew bundle prints deploy commands + auto manifest
## Retro evidence

> The bundle/README.md is essentially the runbook, but the steps are spread across prose paragraphs (different sections describe load vs compose vs healthcheck). I had to read it carefully and assemble the command myself.

> A `vibew remote logs` against the bundle's host would have been worth a lot.

## Problem

After `deploy.sh` was retired in #1138, the bundle README became the deploy runbook — but the commands are scattered across prose paragraphs and the agent has to assemble them. Plus there's no manifest of what's in the bundle.

A separate triage proposal for `vibew remote *` CLI commands was rejected (would re-open the orchestration door closed by ADR-086 / #1138). The read-only ssh recipe is folded into this issue's docs work instead.

## Acceptance criteria

- [ ] **Top of `bundle/README.md`:** a single fenced shell block with the literal copy-pasteable command sequence:
  ```bash
  ssh user@host 'mkdir -p /opt/<app>'
  scp -r .vibewarden/bundle/* user@host:/opt/<app>/
  ssh user@host "cd /opt/<app> && docker load -i image.tar && docker compose up -d"
  curl -fsSL https://<domain>/_vibewarden/health
  ```
- [ ] Prose explanation moves *below* the command block ("why mkdir first", "why -i", etc.) — for context, not for sequencing.
- [ ] Inline read-only ssh recipe documented near the bottom: `ssh user@host docker compose -f /opt/<app>/compose.yml logs --tail 50` (and `ps`, and `curl …/_vibewarden/health`).
- [ ] `vibew bundle` stdout prints the literal commands at the end, with `<app>`, `<domain>`, and bundle path filled in from the live config.
- [ ] Bundle includes an auto-generated `MANIFEST.md` (or section in README) listing every file with one-line descriptions. Generated at bundle time, deterministic, no rot.
- [ ] Test: an agent given only `vibew bundle` stdout (no other docs) deploys successfully.

## Scope notes

- Folds in the dropped `vibew remote logs` proposal — solved with a documented one-line ssh recipe instead of a new CLI command.
- Depends on the deploy contract from #1138 (already shipped).
- Does NOT re-introduce `deploy.sh` (per `CLAUDE.md` §Architecture principles → Artifact policy).

## Severity

**MEDIUM.** Workaround exists (read the prose carefully). Fix improves agent throughput.

## Source

tibtof:
Closes #1204

## Summary

Three changes to the `vibew bundle` output pipeline:

- **Bundle README opens with a fenced deploy block.** The README was prose-only; now the first section after the heading is a `\`\`\`bash` block with the four literal commands (ssh mkdir, scp, docker load+compose up, curl health). `app.name` and `tls.domain` are substituted at render time; empty values produce `<your-app>`/`<your-domain>` placeholders. With `--skip-image` the `docker load -i image.tar &&` clause is omitted so the block is always copy-pasteable.

- **Auto-generated `MANIFEST.md`** in every bundle. Written last in `writeBundleExtras` so it walks the real output directory (not an in-memory list). Entries are sorted alphabetically. Known files get one-line descriptions from `manifestFileDescriptions`; unknown files get `"bundle artifact"` so future generators don't silently drop entries. `MANIFEST.md` does not list itself.

- **`vibew bundle` stdout "Next: deploy" block** replaces the single `see README.md` line. Prints the four literal commands with `app.name` and `tls.domain` substituted — an agent has a copy-pasteable next step without opening a file.

## New stdout sample

```
Bundle written to /home/user/myapp/.vibewarden/bundle

Contents:
  .env
  MANIFEST.md
  README.md
  docker-compose.yml
  sample.env
  vibewarden.yaml

Next: deploy
    ssh user@host 'mkdir -p /opt/myapp'
    scp -r /home/user/myapp/.vibewarden/bundle/* user@host:/opt/myapp/
    ssh user@host "cd /opt/myapp && docker load -i image.tar && docker compose up -d"
    curl -fsSL https://myapp.example.com/_vibewarden/health

See /home/user/myapp/.vibewarden/bundle/README.md for context and read-only inspection commands.
```

## Key test: `TestBundle_Readme_AlignsWithDeployTmpl`

Loads `internal/cli/templates/prompts/deploy.tmpl` (the #1203 canonical deploy recipe), renders the bundle README with concrete substitution values, extracts the four deploy commands from both sources, and asserts they are equivalent after normalising template variables. If a future change touches one source and not the other, this test fails with a precise diff. This is the drift guard between #1203 and #1204.

## Test plan

- `TestBundle_Extras_Readme_DeployContract` — inverted: positive on `scp`/`ssh`/`docker compose up`, negative only on `./deploy.sh` (still forbidden).
- `TestBundle_Extras_Readme_FencedDeployBlock` — fenced block within first 30 lines; exact command strings; skip-image omits docker load.
- `TestBundle_Extras_Readme_ReadOnlyRecipes` — three read-only inspection commands in README.
- `TestBundle_Extras_Readme_PlaceholdersWhenAppOrDomainMissing` — empty values produce `<your-app>`/`<your-domain>`.
- `TestBundle_Extras_Manifest_DeterministicAndSorted` — body identical across two renders; entries sorted; MANIFEST.md not listed in itself.
- `TestBundle_Extras_Manifest_UnknownFilesGetGenericDescription` — unknown file gets `"bundle artifact"` description.
- `TestBundle_Readme_AlignsWithDeployTmpl` — forensic drift guard (see above).
- `TestBundle_Stdout_PrintsLiteralDeployCommands` — stdout contains `Next: deploy` + substituted commands; `--skip-image` omits docker load.
- All existing bundle tests still pass. `make check` clean (gofmt + golangci-lint + go test -race ./...).


internal/app/bundle/bundle_extras.go
```diff
@@ -19,8 +20,32 @@ const (
...
+
+// manifestFileDescriptions maps known bundle filenames to their one-line
+// descriptions for MANIFEST.md. Unknown files receive a generic description.
+var manifestFileDescriptions = map[string]string{
+	"vibewarden.yaml":            "merged production config",
+	"vibewarden.production.yaml": "production overrides",
+	"docker-compose.yml":         "deterministic compose file",
+	"image.tar":                  "docker-saved app image (omit when --skip-image)",
+	"README.md":                  "deploy contract + command reference",
+	"MANIFEST.md":                "this file",
```
tibtof: **Dead code**: `manifestFileDescriptions["MANIFEST.md"]` maps to `"this file"` (line 44) but `MANIFEST.md` is explicitly excluded from its own listing at line 291:
```go
if e.Name() == fileManifest {
    continue // exclude self
}
```
This map entry will never be read. Remove it to avoid confusing future maintainers who might wonder why the self-description is needed.

internal/app/bundle/bundle_extras_test.go
```diff
@@ -537,3 +549,411 @@ func TestBundle_Extras_ImageSaverError_PropagatesWrapped(t *testing.T) {
...
+	if len(tmplCommands) == 0 {
+		t.Fatal("deploy.tmpl: could not extract any deploy commands — check regex / template format")
+	}
+	if len(readmeCommands) == 0 {
+		t.Fatal("bundle README: could not extract any deploy commands — check regex / README format")
+	}
+
+	// Assert each tmpl command is also in the README (order independent, since
+	// the README may split steps differently). This is the drift guard.
+	for _, cmd := range tmplCommands {
```
tibtof, line 910: **Forensic test is one-directional**: `TestBundle_Readme_AlignsWithDeployTmpl` asserts every command from `deploy.tmpl` appears in the bundle README, but NOT the reverse. If a deploy command is added to the README that has no equivalent in `deploy.tmpl`, the test passes silently.

In practice this is low-risk because the README commands come from the same hardcoded block, and the read-only recipes (logs, ps) match the `ssh ` prefix of the regex but are not in `tmplCommands` so they're never checked. Still, consider a comment above the assertion loop explaining the one-directional intent, so future readers understand why the asymmetry is deliberate:
```go
// One-directional by design: we guard that README contains every tmpl command.
// Extra commands in the README (e.g., read-only recipes) are intentionally not
// checked against the template — they serve a different purpose.
```
The current test is effective as a drift guard for the critical 4-command sequence. This note is non-blocking.

tibtof:

## Required fix (blocking)

**Unpushed commit not in PR**: Local branch `feat/1204-bundle-readme-manifest` has a second commit (`bf08494`) that was never pushed and is NOT included in the PR. This commit:
- Fixes line 62 of `internal/cli/templates/prompts/deploy.tmpl`: changes the stale `# The bundle README is prose-only; ADR-099 is the authoritative source for this sequence.` to the correct `# The bundle README opens with this same fenced block (app.name and tls.domain substituted).`
- Adds `MANIFEST.md` to the `llms-full.txt` Step 6 file listing.

As submitted, `deploy.tmpl` contains a comment that directly contradicts this PR's implementation. Agents parsing that file will be told the README is prose-only immediately before reading the canonical deploy recipe — this is the documentation drift CLAUDE.md requires the writer agent to guard. Push the fix:
```
git push origin feat/1204-bundle-readme-manifest
```

## Minor fix (non-blocking, fix before merge)

**Dead code in `manifestFileDescriptions`**: `"MANIFEST.md": "this file"` at line 44 of `bundle_extras.go` is unreachable — `MANIFEST.md` is excluded from its own listing at line 291 before the description lookup. Remove the entry.

## Findings (non-blocking, no changes required)

**Forensic test direction**: `TestBundle_Readme_AlignsWithDeployTmpl` is one-directional — it guards that every `deploy.tmpl` command appears in the README, not the reverse. The asymmetry is safe because the read-only recipes that exist in the README but not in `deploy.tmpl` are intentionally different (they serve inspection, not deployment). A short comment explaining this above the assertion loop would help future reviewers. Inline comment posted.

**MANIFEST.md content gap via memBundleFS**: Full-pipeline tests (`TestBundle_Extras_WritesExpectedFileSet`, etc.) use `memBundleFS` for extras but `renderBundleManifest` walks the real `outDir` on disk. Because the extras pipeline writes README/env via `memBundleFS` (not disk), the MANIFEST.md generated in those tests will only list files the generator wrote to disk — not the extras. The manifest render function itself is well-tested in isolation via `TestBundle_Extras_Manifest_DeterministicAndSorted`, which uses `os.WriteFile` directly. This is a known design tension (documented in the code comment) rather than a bug, but there is no test that verifies the MANIFEST.md actually lists all bundle files in an end-to-end run. Acceptable for this PR; worth a follow-up issue if the MANIFEST content completeness becomes load-bearing.

## What passes cleanly

- All tests: `go test ./...` — clean
- Race detector: `go test -race ./internal/app/bundle/... ./internal/cli/cmd/...` — clean
- `gofmt -l` — no output (all changed files clean)
- No new `go.mod` dependencies
- Branch name: `feat/1204-bundle-readme-manifest` — correct convention
- Architecture: `bundle_extras.go` imports stdlib only — hexagonal boundary preserved
- `RenderBundleReadme` and `RenderBundleManifest` exported for direct test use — correct pattern
- Forensic test (`TestBundle_Readme_AlignsWithDeployTmpl`) passes and correctly uses `runtime.Caller` to locate the source file — the drift guard works
- Substitution placeholders (`<your-app>`, `<your-domain>`) correct for empty inputs
- `--skip-image` removes `docker load -i image.tar &&` from both README fenced block and stdout — consistent
- `writeBundleExtras` writes MANIFEST.md last so it walks the actual file set — correct ordering
- The `"bash"` token correctly removed from the forbidden list in `TestBundle_Extras_Readme_DeployContract` (README now has ```bash fenced blocks)
- README line count: 55 lines — within the `<= 60` test limit
- `#1200` alignment correct: no arch note in the stdout deploy block; arch pre-flight already hard-fails before `runBundle` reaches the stdout print

tibtof:
@reviewer addressed: dead `MANIFEST.md` description entry removed. Ready for re-review.

Issue #1220, opened 2026-04-29
feat(retro-0.18.2): vibew dev --rebuild — stop+rmi+rebuild+start in one command


tibtof:
## Retro evidence

From the v0.18.2 retrospective (qr-code-blackhole Node/Express deploy):

> Expected but missing: `vibew dev --rebuild` flag (I had to `vibew down && docker rmi && vibew build && vibew dev` — four commands to do "rebuild from scratch")

## Coupling with #1225 (image-tag collision)

The recovery path for #1225's image-tag-collision block is "rebuild from scratch." Without this flag, every collision recovery is a 4-command dance. With it, the collision message says `vibew dev --rebuild` and that's a one-liner.

## Severity

**HIGH** when paired with #1225. Standalone, this is a DX nicety. Filed as HIGH.

## Acceptance criteria

- [ ] `vibew dev --rebuild` performs the equivalent of: stop the dev stack → remove the resolved app image (`<project>-app:latest`) → rebuild via `vibew build` → start the dev stack.
- [ ] Volumes are **preserved by default**. Add `vibew dev --rebuild --volumes` for explicit volume reset (named, opt-in, surprising-data-loss prevention).
- [ ] Verbose stdout naming each destructive action, in order:
  ```
  Stopping stack...
  Removing image qr-code-blackhole-app:latest
  Rebuilding image...
  Starting stack...
  ```
- [ ] Exit codes match `vibew dev`'s existing semantics; new sentinel only if needed for `--rebuild` failures (probably not).
- [ ] Test: rebuild on a clean project (no stack running) succeeds. Rebuild while stack is up tears down + recreates cleanly. `--volumes` removes named volumes; default does not.
- [ ] No ADR. CHANGELOG entry under Added.

## Out of scope

- Bundle-time `--rebuild` symmetry: handled differently by `vibew bundle --build` (already exists). Don't extend bundle here.
- `--no-cache` for the build step: separate flag if needed; default uses cache.

## Source

Retro: `~/notes/vibewarden/retro-0.18.2.md` (2026-04-28)

tibtof:
Status: READY_FOR_DEV

Design (no ADR — pure CLI composition over existing primitives; per the threshold rule and #1220's spec, decisions live here + CHANGELOG).

## Strategy

[...]

PR #1225, opened 2026-04-29
fix(#1219): vibew dev blocks on stale image from prior project (org.vibewarden.project-root-hash label)

### LINKED ISSUE #1219: fix(retro-0.18.2): vibew dev silently uses stale <project>-app:latest from prior unrelated project
## Retro evidence

> `qr-code-blackhole-app:latest` already existed on my machine from a previous experiment (a Go smoke binary), and `vibew dev` happily reused it. The stack came up "healthy", `/health` returned 200, but the served HTML was from the other project. I burned ~5 minutes thinking my Express routes were misordered before I `docker exec`'d in and found `/app` didn't even exist — the container was running `/usr/local/bin/app`, a Go binary.

> There is no warning anywhere that an existing image with the conventional tag will shadow your build.

## Mechanism

- Image tag = `<project>-app:latest` where `<project>` = `app.name` (post-#1199 = dirname unless `--name` overridden).
- Two projects with the same dirname → same tag.
- Compose's default `image_pull_policy` is `if-not-present` → existing image wins. No rebuild, no warning, silent serve of foreign content.

## Severity

**CRITICAL.** Most dangerous failure mode in the retro: silent success producing wrong result. Health endpoint returned 200; agent had to `docker exec` to discover the image was from a different project entirely.

## Acceptance criteria

- [ ] `vibew build` stamps a Docker label `org.vibewarden.project-root=<absolute-path>` on the produced image. (Path can be hashed if exposing absolute path is undesirable; both must be supported behaviors with a fallback.)
- [ ] On `vibew dev` startup, before bringing the stack up, `docker image inspect` reads the label from the resolved app image. If absent or mismatched against the current project root, **block with exit 1** and a clear message:
  ```
  Error: image qr-code-blackhole-app:latest was built from a different project (or is missing the project label). 
    Built from: /Users/foo/old-project    (or: <unknown — image predates this check>)
    Current:    /Users/tibtof/qr-code-blackhole
  Rebuild with: vibew dev --rebuild   (paired with #1220)
  ```
- [ ] Unlabelled images are treated as "could be from anywhere" → block with the same message.
- [ ] No volume reset, no destructive actions on block — just exit 1.
- [ ] Backwards compatibility note: existing local images don't have the label. Document in CHANGELOG that first post-upgrade `vibew dev` will block until `--rebuild` is used; this is intentional.
- [ ] Test: a project with a foreign-labelled image at the expected tag → `vibew dev` exits 1 with the actionable message. Project with the right label → `vibew dev` proceeds normally. Project with no label → blocks like the foreign case.
- [ ] Architect: assess whether this small-but-load-bearing pattern (Docker label as image identity) deserves an ADR. My read: borderline. Decide and justify in the status comment.

## Out of scope

- Rebuild flow itself — handled by #1220 (the `--rebuild` flag).
- Bundle-time arch+image inspection — already covered post-#1200; this issue is dev-time.

## Source

tibtof:
Closes #1219
ADR: [ADR-100](decisions/adr-100-image-identity-via-project-root-label.md)

## Summary

Implements the image-identity label scheme from ADR-100. Closes the silent-wrong-result bug from the v0.18.2 retrospective: `vibew dev` silently used a stale `<project>-app:latest` image from a prior unrelated project sharing the same directory name, served foreign content, reported "healthy".

**Label scheme:**
- `org.vibewarden.project-root-hash` = `sha256:<64-hex>` of `filepath.EvalSymlinks(projectRoot)` — authoritative, compared for equality.
- `org.vibewarden.project-root` = realpath verbatim — informational, shown in error messages.

**Behaviour change (BREAKING for existing local images):**
Every image built by VibeWarden ≤ v0.18.2 is unlabelled. The first `vibew dev` after upgrading blocks with Variant 2 (see below). Recovery: `vibew build` (or `vibew dev --rebuild` when #1220 lands).

Custom images via `app.image:` that differ from the vibew-derived canonical tag are skipped with one INFO line — the labels likely belong to a different toolchain.

**Two error variants (byte-exact, golden-tested):**

Variant 1 — labelled but different project:
```
Error: app image qr-code-blackhole-app:latest was built from a different project.
  Built from: /Users/foo/old-project
  Current:    /Users/tibtof/qr-code-blackhole

Rebuild with: vibew dev --rebuild
```

Variant 2 — unlabelled (legacy/foreign builder):
```
Error: app image qr-code-blackhole-app:latest is missing the vibew project-root label.
  This image was built before VibeWarden v0.19.0 OR by something other than vibew build.
  Current project: /Users/tibtof/qr-code-blackhole

Rebuild with: vibew dev --rebuild
```

Note: `vibew dev --rebuild` references #1220 which is not yet merged. Until #1220 ships, recovery is `vibew build && vibew dev`.

## Files changed

**New:**
- `internal/app/ops/image_identity.go` — `ImageIdentity`, `ProjectRootHash`, `VerifyAppImageIdentity`, `ErrProjectRootMismatch`, label constants
- `internal/app/ops/image_identity_test.go` — hash determinism, symlink resolution, mismatch/missing label detection
- `internal/app/ops/dev_format.go` — `formatProjectRootMismatch` (pinned wording)
- `internal/app/ops/dev_format_test.go` — golden tests for both message variants
- `internal/app/ops/export_test.go` — exposes `formatProjectRootMismatch` for tests
- `decisions/adr-100-image-identity-via-project-root-label.md`

**Modified:**
- `internal/ports/ops.go` — `DockerBuildOptions` struct replaces positional args on `DockerBuilder.Build`
- `internal/ports/image_inspector.go` — `+Labels map[string]string` on `ImageInfo`
- `internal/adapters/ops/build.go` — emit `--label` flags sorted by key
- `internal/adapters/ops/build_test.go` — assert label flags in arg slice
- `internal/adapters/ops/image_inspect.go` — parse `Config.Labels`
- `internal/adapters/ops/image_inspect_test.go` — labels round-trip tests
- `internal/adapters/ops/export_test.go` — expose `buildDockerArgs` for tests
- `internal/app/ops/build.go` — compute and pass labels via `BuildOptions`
- `internal/app/ops/build_test.go` — assert both label keys present and correct
- `internal/app/ops/dev.go` — `verifyAppImageIdentity` step, `WithImageInspector`, `isUserSetImage`
- `internal/app/ops/dev_test.go` — 8 new cases (match/mismatch/unlabelled/user-set/build-set/nil-inspector/daemon-down/generic-error)
- `internal/cli/cmd/dev.go` — wire `NewImageInspectAdapter` onto `DevService`
- `CHANGELOG.md` — Behavior changes entry under Unreleased

## Test plan

- `make check` clean (gofmt + golangci-lint + build + race tests, verified on branch)
- `go test -race ./...` — all packages pass
- Golden tests in `dev_format_test.go` pin the exact byte output of both message variants
- Table-driven `image_identity_test.go` covers: stable hash, sha256-hex format, different dirs → different hash, symlink → same hash, path is absolute, match→nil, mismatch→ErrProjectRootMismatch, missing label→ErrProjectRootMismatch, inspector error→wrapped error
- `dev_test.go` integration-shaped cases assert compose.Up is not called on mismatch or unlabelled image; user-managed and build-set images proceed without calling the inspector; daemon-down/generic errors degrade gracefully
- `build_test.go` asserts both label keys are always passed to the fake builder, hash matches `ProjectRootHash(dir)`, nil-cfg uses workDir


internal/adapters/ops/image_inspect_test.go
```diff
@@ -131,3 +131,61 @@ func TestImageInspectAdapter_HappyPath(t *testing.T) {
 		t.Errorf("SizeBytes = %d, want 170_000_000", info.SizeBytes)
 	}
 }
+
```
tibtof, line 134: **Missing test: real adapter JSON parsing for Config.Labels**

The new tests added to `internal/adapters/ops/image_inspect_test.go` (e.g. `TestImageInfo_Labels_RoundTrip`, `TestImageInfo_EmptyLabelsMap`) exercise the port struct via `fakeImageInspector`, not the actual adapter. The real JSON-parsing path — specifically the `Config.Labels` field now wired in `dockerInspectOutput` — has no unit test that feeds it a real JSON blob and asserts the output.

**Why it matters**: The ADR specifies "golden tests cover labels round-trip" for the adapter test file. The actual code change that matters is in `image_inspect.go`:
```go
labels := out.Config.Labels
if labels == nil {
    labels = make(map[string]string)
}
```
If someone accidentally breaks the JSON field name (e.g. renames `Config` struct or the `Labels` json tag), the existing unit tests will not catch it — they bypass `json.Unmarshal` entirely.

**Fix**: Export a `ParseInspectOutputForTest` helper in `internal/adapters/ops/export_test.go` (same pattern as `ParseDownOutputForTest` and `BuildDockerArgsForTest` already there). Add two table-driven test cases that call it with:
1. A JSON blob containing `"Config":{"Labels":{"org.vibewarden.project-root-hash":"sha256:abc"}}` — asserts `info.Labels["org.vibewarden.project-root-hash"] == "sha256:abc"`.
2. A JSON blob with no `Config` key or `"Config":{"Labels":null}` — asserts `info.Labels` is non-nil and empty.

internal/app/ops/build_test.go
```diff
@@ -464,3 +463,100 @@ func TestBuildService_Run_ImageNameMatchesComposeProjectName(t *testing.T) {
 		})
```
tibtof, line 463: **Missing: architecture invariant test that BuildService always stamps both labels**

The ADR's test strategy explicitly calls out:
> `test/architecture/...` (existing pattern from ADR-087): add an invariant test that `BuildService.Run` always passes both label keys to the fake builder.

No such test was added. The unit tests in this file cover the happy-path label stamping well, but an architecture invariant test in `test/architecture/` is the regression guard that catches a future PR adding a second build path (e.g. fast-path for #1220 rebuilds) that forgets to call `ProjectRootHash`.

**Fix**: Add `test/architecture/image_labels_test.go` (or append to an existing arch test file) asserting that `BuildService.Run` always populates both `LabelProjectRootHash` and `LabelProjectRoot` in the captured `BuildOptions.Labels`, using the same fake-builder pattern already established in ADR-087's architecture tests.

internal/app/ops/dev_format.go
```diff
@@ -0,0 +1,52 @@
...
+			"Error: app image %s is missing the vibew project-root label.\n"+
+				"  This image was built before VibeWarden v0.19.0 OR by something other than vibew build.\n"+
+				"  Current project: %s\n"+
+				"\n"+
+				"Rebuild with: vibew dev --rebuild",
+			tag, currentProjectRoot,
+		)
+	}
+	return fmt.Errorf("%s", msg) //nolint:err113 // dynamic user-facing error message; wording is golden-tested
+}
```
tibtof, line 63: **Aspirational `vibew dev --rebuild` wording — CHANGELOG hedges correctly, error message does not**

The CHANGELOG entry reads:
> Recovery: rebuild via `vibew build` (or `vibew dev --rebuild` once #1220 lands).

That hedge is correct and I noted it. But the golden-tested error message in `dev_format.go` is unhedged:
```
Rebuild with: vibew dev --rebuild
```
If #1220 does not ship in the same release as this PR, users who hit this error will run `vibew dev --rebuild`, get "unknown flag: --rebuild", and be worse off than before — they now have a broken suggestion on top of the original problem.

**This is a BLOCKING concern if #1220 is not guaranteed to ship atomically with this PR.** The architect's status comment says "dev agents must coordinate so #1220 ships in the same release. If #1220 slips, fall back wording: `vibew down && docker rmi <image> && vibew build && vibew dev`".

**Fix options** (in order of preference):
1. If #1220 is already open and will be merged in the same release cut: add a GitHub Closes reference in the commit and verify #1220 is merged before v0.19.0 is tagged.
2. If #1220 is not ready: change the message to the fallback wording the architect spec'd: `vibew down && docker rmi <image> && vibew build && vibew dev`. Update the golden test.

CHANGELOG.md
```diff
@@ -12,6 +12,34 @@ This initial entry was written by hand to summarise the work leading up to v0.1.
...
+
+**Breaking for existing local images.** Every image built by VibeWarden ≤ v0.18.2 is
+unlabelled. The first `vibew dev` after upgrading will block with:
+
+```
+Error: app image <tag> is missing the vibew project-root label.
+  This image was built before VibeWarden v0.19.0 OR by something other than vibew build.
+  Current project: <path>
+
+Rebuild with: vibew dev --rebuild
```
tibtof, line 36: **Docs gap: `docs/cli.md` and `AGENTS-VIBEWARDEN.md` not updated**

The ADR specifies the following docs changes in the same PR:
- `docs/cli.md` — vibew dev blocking-behaviour section; cross-link to `vibew dev --rebuild` (#1220).
- `AGENTS-VIBEWARDEN.md` template — one paragraph under "vibew dev" telling the agent that an unlabelled image will block and the recovery is `vibew dev --rebuild`.

Neither file was modified in this PR. Note: the writer agent is the authority on doc consistency, but since this gap is explicitly enumerated in the ADR I am flagging it here for completeness.

The missing `AGENTS-VIBEWARDEN.md` update is the higher-priority one: without it, agents running on this repo post-upgrade will not know that `vibew dev` now blocks on unlabelled images and may loop on the error rather than immediately running `vibew build`. This is exactly the docs-runtime drift that three retrospectives identified as the #1 source of agent failures (CLAUDE.md).

tibtof:

Reviewed commit `0323788`. Architecture, domain logic, port/adapter layering, and the core identity check implementation are all correct and clean. Three issues require changes before merge; the `--rebuild` item is the highest-risk one.

---

### Finding 1 — BLOCKING (if #1220 is not ready): Aspirational `vibew dev --rebuild` in golden-tested error message

The error message in `internal/app/ops/dev_format.go` reads `Rebuild with: vibew dev --rebuild` — but `--rebuild` is defined in issue #1220, not yet merged. The CHANGELOG entry correctly hedges with "once #1220 lands", but the actual user-visible message does not. A user who upgrades to v0.19.0 before #1220 ships will run `vibew dev --rebuild`, get "unknown flag: --rebuild", and be left with no actionable path.

The architect's spec explicitly states: *"If #1220 is not merged in time, the message can substitute `vibew down && docker rmi <image> && vibew build && vibew dev` as a fallback."*

**Required**: Either confirm #1220 merges atomically in the same release, or change both the message and its golden test to the fallback wording before merging this PR. See inline comment on `dev_format.go:52`.

---

### Finding 2 — REQUIRED: Missing adapter-layer JSON-parse test for `Config.Labels`

The new tests in `internal/adapters/ops/image_inspect_test.go` exercise `Labels` via `fakeImageInspector` — they test the port struct, not the adapter. The actual code path that changed is in `image_inspect.go`: `json.Unmarshal` → `out.Config.Labels` → `ImageInfo.Labels`. That path has no unit test today. If the JSON field name or struct tag were accidentally renamed, all tests would still pass.

**Required**: Export a `ParseInspectOutputForTest` helper in `internal/adapters/ops/export_test.go` (matching the existing `ParseDownOutputForTest`/`BuildDockerArgsForTest` pattern) and add two test cases: (1) JSON with `Config.Labels` populated, (2) JSON with `Config.Labels` null/absent. See inline comment on `image_inspect_test.go:134`.

---

### Finding 3 — REQUIRED: Architecture invariant test missing from `test/architecture/`

ADR-100's test strategy section explicitly specifies:
> `test/architecture/...`: add an invariant test that `BuildService.Run` always passes both label keys to the fake builder.

This test was not added. The unit tests in `build_test.go` cover the current happy path, but the arch-invariant test is the regression guard that catches a future PR (e.g. a rebuild fast-path in #1220) that adds a new build code path without calling `ProjectRootHash`. ADR-087 established this pattern precisely for this class of regression.

**Required**: Add `test/architecture/image_labels_test.go` (or extend an existing arch test file) with the invariant. See inline comment on `build_test.go:463`.

---

### Informational (writer agent to verify)

- `docs/cli.md` and `AGENTS-VIBEWARDEN.md` were listed in the ADR as required changes in this PR but are absent from the diff. The writer agent is the authority on this — flagged for that agent's attention. The `AGENTS-VIBEWARDEN.md` gap is higher-priority given three retros on docs-runtime drift. See inline comment on `CHANGELOG.md:36`.

---

### What passes cleanly

- Hexagonal purity: `internal/app/ops/image_identity.go` is a pure value object + pure function (modulo `filepath.EvalSymlinks` FS read). No Docker calls, no I/O other than the defined `ports.ImageInspector` interface. Clean.
- `EvalSymlinks` fallback to `filepath.Abs` on failure, with debug log — matches ADR spec exactly.
- Port signature change (`DockerBuildOptions` struct) is local blast radius; all callers updated.
- All in-repo callers of `DockerBuilder.Build` migrated (bundle cmd wires through `BuildService` which was updated).
- Label sort in `buildDockerArgs` is deterministic and correctly placed before contextDir.
- `verifyAppImageIdentity` call is between `checkAppImage` and `checkContainerFreshness` — correct ordering.
- Golden tests for both error variants are byte-exact against the ADR-spec wording.
- All skip conditions (nil inspector, user-set image, compose-builds, `ProjectRootHash` failure, inspector error) are tested with the correct assertions (`fakeCompose.UpCalls == 0` on mismatch).
- No new dependencies — stdlib only (`crypto/sha256`, `encoding/hex`).
- No global variables, no `init()` side effects, no panics outside main.
- Branch name `fix/1219-image-identity-label` matches convention.
- CHANGELOG breaking-change framing is clear and accurate.

tibtof (commented):
Documentation review: docs consistent with code changes after fixes applied in this review.

Issues found and fixed in commit c3dc612:

1. **llms-full.txt — vibew build and vibew dev CLI table rows** were missing the new behavior entirely. Updated both rows: `vibew build` now mentions label stamping; `vibew dev` now documents the identity pre-flight, both error variants, the recovery command (`vibew build && vibew dev`), and the `app.image:` bypass. Added a full "vibew dev: image identity check failed" troubleshooting section with verbatim error text, causes, fixes, and the post-upgrade trap.

2. **docs/troubleshooting.md** had no entry for the new failure mode. Added "Image identity check failed (v0.19.0+)" with both verbatim error variants (byte-exact from the ADR/golden tests), cause, fix, and the `app.image:` skip note.

3. **AGENTS-VIBEWARDEN template and static example** (internal/cli/templates/agents/agents-vibewarden.md.tmpl and docs/examples/AGENTS-VIBEWARDEN.md) — the ADR explicitly listed the template as a doc surface to update, but it was absent from the PR diff. Added an image identity check bullet to Known limitations in both files, including the directory-reuse trap and the post-upgrade one-time block.

Pre-existing correct: CHANGELOG.md entry is specific, covers all four required points (block behavior + both trigger cases, aspirational `vibew dev --rebuild` reference, legacy-image upgrade caveat, `app.image:` escape hatch). decisions/README.md ADR-100 row present. vibewarden.reference.yaml: N/A (no new config fields). Cross-references with #1197, #1199, #1203: no contradictions.

CHANGELOG error text matches the golden-tested variants from `dev_format_test.go`.

Writer review: APPROVED.

tibtof:
@reviewer addressed:

**Finding 1 — aspirational `vibew dev --rebuild`:** Not changing the message. Added a detailed code comment in `internal/app/ops/dev_format.go` (above `formatProjectRootMismatch`) explicitly stating that the recovery command is delivered by #1220 in the same v0.18.3 release cycle, and that the merge-ordering invariant is the release gate. The comment also documents the exact fallback wording to use if #1219 ever needs to ship ahead of #1220. Added a matching release-gate note to the CHANGELOG entry.

**Finding 2 — adapter JSON parse test gap (BLOCKING):** Extracted the JSON→ImageInfo conversion from `Inspect` into an unexported `parseInspectJSON` helper in `image_inspect.go`, then exposed it via `ParseInspectOutputForTest` in `internal/adapters/ops/export_test.go` (matches the existing `ParseDownOutputForTest` / `BuildDockerArgsForTest` pattern). Added two test functions in `image_inspect_test.go`:
- `TestParseInspectOutput_LabelsPopulated`: feeds a real JSON blob with both `org.vibewarden.project-root-hash` and `org.vibewarden.project-root` in `Config.Labels`; asserts both keys land in `ImageInfo.Labels`.
- `TestParseInspectOutput_LabelsAbsent`: table-driven with two sub-cases — `Config.Labels null` and `Config key absent entirely`; asserts `ImageInfo.Labels` is non-nil and empty in both.

**Finding 3 — architecture invariant test (BLOCKING):** Added `test/architecture/image_labels_test.go` with `TestBuildService_Run_AlwaysStampsBothLabelKeys`. Uses a `captureBuilder` fake (implements `ports.DockerBuilder`, records `DockerBuildOptions`), three invocation variants (explicit config + ProjectRoot, nil config using WorkDir as root, pre-resolved ImageTag), and asserts `LabelProjectRootHash` + `LabelProjectRoot` are present with correct format in every captured call. The file docstring explains why this is load-bearing for a future fast-rebuild path in #1220.

Writer's doc commits (c3dc612) are untouched. Commit SHA: ded2935. `make check` clean (golangci-lint 0 issues, all tests green, race detector clean). Test delta: +6 new test functions across `internal/adapters/ops` (+3) and `test/architecture` (+1). Ready for re-review.

tibtof:
**Reviewer Agent: APPROVED**

Re-review of commit `ded2935`. All three previously blocking findings resolved. Line-level confirmations below.

---

**Finding 1 — aspirational `vibew dev --rebuild` (concern):** CONFIRMED RESOLVED.

`internal/app/ops/dev_format.go` lines 29–38: the block comment above `formatProjectRootMismatch` is present and complete. It explicitly names v0.18.3 as the release cycle, states "Both #1219 and #1220 MUST merge before tagging v0.18.3 — this is the merge-ordering invariant for this retro batch", and documents the exact fallback wording (`vibew down && docker rmi <image> && vibew build && vibew dev`) with the instruction to update the golden tests if the split-ship path is ever taken.

`CHANGELOG.md` lines 43–47: "Release gate:" paragraph is present and explicit — names #1220, names v0.18.3, names the fallback command, and cross-references `dev_format.go`.

---

**Finding 2 — adapter JSON parse test gap (was BLOCKING):** CONFIRMED RESOLVED.

`internal/adapters/ops/image_inspect.go`: unexported `parseInspectJSON` helper extracted (lines ~73–117). `Inspect` delegates to it; the `info.Tag = tag` assignment happens in `Inspect` after the call, correctly excluded from the helper. Error wrapping is correct (`fmt.Errorf("docker image inspect: %w", err)` at both call sites).

`internal/adapters/ops/export_test.go`: `ParseInspectOutputForTest` is present, in package `ops` (not `ops_test`), delegates directly to `parseInspectJSON`. Matches the `ParseDownOutputForTest` / `BuildDockerArgsForTest` pattern exactly.

`internal/adapters/ops/image_inspect_test.go`:
- `TestParseInspectOutput_LabelsPopulated` (lines ~190–243): feeds a real JSON blob with both `org.vibewarden.project-root-hash` and `org.vibewarden.project-root` in `Config.Labels`; asserts both keys land in `ImageInfo.Labels` with correct values; spot-checks `Architecture` and `SizeBytes` to confirm full parsing ran.
- `TestParseInspectOutput_LabelsAbsent` (lines ~245–305): table-driven with two sub-cases — `Config.Labels null` and `Config key absent entirely`; asserts `info.Labels` is non-nil and empty in both. This directly guards the `if labels == nil { labels = make(map[string]string) }` branch in `parseInspectJSON`.

---

**Finding 3 — architecture invariant test (was BLOCKING):** CONFIRMED RESOLVED.

`test/architecture/image_labels_test.go`: `TestBuildService_Run_AlwaysStampsBothLabelKeys` is present. Three invocation variants tested: explicit config + ProjectRoot, nil config (WorkDir as root), pre-resolved ImageTag. `captureBuilder` records `DockerBuildOptions` from every `Build` call. Assertions verify `Labels != nil`, `LabelProjectRootHash` present and `sha256:`-prefixed, `LabelProjectRoot` present. The file docstring correctly explains why this is a load-bearing regression guard for #1220's fast-rebuild path.

---

**Commit ordering:** `ded2935` parent is `c3dc612` (writer doc commit). Not amended. Correct.

**Branch:** `fix/1219-image-identity-label`. No force push evidence. Correct.

All four open review threads resolved. Ready to merge once CI is green.

PR #1226, opened 2026-04-29
feat(#1220): vibew dev --rebuild — stop+rmi+rebuild+start in one command

### LINKED ISSUE #1220: feat(retro-0.18.2): vibew dev --rebuild — stop+rmi+rebuild+start in one command

tibtof:
Closes #1220

## Summary

- Adds `vibew dev --rebuild` flag: stops the stack, removes the resolved app image, rebuilds via `vibew build` (which stamps #1219's identity labels on the new image), then starts the stack.
- Adds `vibew dev --rebuild --volumes` to also reset named volumes (Postgres data, LE certs, etc.).
- Flag-conflict validation: `--rebuild` + `--watch` → rejected. `--volumes` without `--rebuild` → rejected.

## Verbose stdout template (pinned in rebuild_test.go)

```
Stopping stack...
Removing image qr-code-blackhole-app:latest
Rebuilding image...
Starting stack...
```

User-set image skip line (also pinned):

```
vibew dev --rebuild: app.image is user-managed; skipping rmi+build, starting stack normally.
```

## #1219 coupling

This is the recovery command for #1219's image-identity mismatch block. Without `--rebuild`, `vibew dev` exits 1 when the app image was built from a different project root. With `--rebuild`, the sequence bypasses the identity check structurally: after `down → rmi → build`, the final `Run` sees a freshly-built image with correct project-root labels and passes naturally — no special-casing needed.

Both #1219 and #1220 must merge before tagging v0.18.3 (release gate satisfied by this PR).

## Files changed

New:
- `internal/adapters/ops/image_remove.go` + test — `ImageRemoveAdapter` shelling out to `docker image rm`. Treats "No such image" as idempotent nil.
- `internal/app/ops/rebuild.go` + test — `DevService.Rebuild` orchestrator (method on existing service, not a new type).
- `internal/cli/cmd/dev_rebuild_test.go` — CLI flag-validation tests.

Modified:
- `internal/ports/ops.go` — `DockerImageRemover` interface appended.
- `internal/app/ops/dev.go` — `WithImageRemover` builder, `imageRemover` field, `DevOptions.Rebuild` / `DevOptions.RebuildVolumes`.
- `internal/app/ops/dev_test.go` — `upCalled` counter added to `fakeCompose` (needed by rebuild tests in same package).
- `internal/adapters/ops/export_test.go` — `BuildImageRmArgsForTest` export helper.
- `internal/cli/cmd/dev.go` — flags wired, rebuild branch.
- `CHANGELOG.md` — Unreleased/Added entry; #1219 release-gate note updated.

## Test plan

- `make check` clean (golangci-lint, gofmt, build, race tests, demo-app).
- All 12 architect-specified test cases covered in `rebuild_test.go` + `dev_rebuild_test.go`.
- `TestRebuild_ForeignLabelBypass` confirms the structural bypass of #1219's identity check.
- `TestRebuild_SubsequentDevPassesIdentityCheck` confirms two-phase behaviour: rebuild then plain dev.


internal/app/ops/rebuild.go
```diff
@@ -0,0 +1,93 @@
...
+//
+//	Stopping stack...
+//	Removing image qr-code-blackhole-app:latest
+//	Rebuilding image...
+//	Starting stack...
+//
+// User-set escape hatch line (pinned by rebuild_test.go):
+//
+//	vibew dev --rebuild: app.image is user-managed; skipping rmi+build, starting stack normally.
+func (s *DevService) Rebuild(
```
tibtof, line 55: **Concrete coupling: `builder *BuildService` should be an interface parameter**

`Rebuild` takes `builder *BuildService` (a concrete application service) while every other optional dependency on `DevService` — `imageChecker`, `imageInspector`, `imageRemover` — is typed as its port interface. This breaks the hexagonal rule that the app layer depends on ports, not on sibling concrete types.

`ports.DockerBuilder` already exists (see `internal/ports/ops.go` line 161). The BuildService owns a `builder ports.DockerBuilder` field. Rebuild only needs the capabilities exposed by that port.

**Why it matters:** If `BuildService` ever gains additional fields (`shellProber`, etc.), callers of `Rebuild` silently inherit those side effects. The identity-label stamping is also tied to `BuildService.Run` — if someone passes a shimmed `BuildService` without `WithShellProber` wired they're fine, but the design doesn't make this explicit.

**Suggested fix:** Change the parameter type to `ports.DockerBuilder` and move the BuildOptions construction into `Rebuild`:

```go
func (s *DevService) Rebuild(
    ctx context.Context,
    cfg *config.Config,
    opts DevOptions,
    builder ports.DockerBuilder,
    out io.Writer,
) error {
    // ...
    if err := builder.Build(ctx, tag, cfg.ResolvedBuildContext(), ports.DockerBuildOptions{
        Labels: computeIdentityLabels(cfg),
    }, out); err != nil {
        return fmt.Errorf("rebuilding image: %w", err)
    }
}
```

Alternatively, keep `*BuildService` but justify it in an ADR comment ("Rebuild delegates to BuildService.Run to guarantee the same label-stamping logic that vibew build uses"). The second option is weaker but defensible for a same-package call.

internal/app/ops/rebuild_test.go
```diff
@@ -0,0 +1,485 @@
...
+		currIdx := strings.Index(out, anchors[i])
+		if prevIdx < 0 || currIdx < 0 {
+			continue // already flagged above
+		}
+		if prevIdx > currIdx {
+			t.Errorf("line %q must appear before %q", anchors[i-1], anchors[i])
+		}
+	}
+
+	// Verify down was called.
```
tibtof, line 107: **Missing `upCalled` assertion in happy path test**

`TestRebuild_HappyPath` verifies Down, Remove, and Build are called, but does not assert `fc.upCalled > 0`. `TestRebuild_BuildFailure` does assert `fc.upCalled == 0` as the negative case. The positive case should mirror that symmetry.

The 'Starting stack...' line is printed just before `s.Run`, which itself calls `compose.Up` — but the line being present in stdout only proves `Rebuild` reached that code path, not that `compose.Up` was actually invoked.

**Fix:** Add before the close of `TestRebuild_HappyPath`:

```go
if fc.upCalled == 0 {
    t.Error("expected compose.Up to be called on the happy path")
}
```

tibtof:

Two issues found, one architectural (blocking), one test gap (minor but required for coverage symmetry).

---

## Issue 1 — Architectural: `builder *BuildService` concrete parameter (BLOCKING)

**File:** `internal/app/ops/rebuild.go` line 50

`Rebuild` takes `builder *BuildService` (a concrete app-layer type) while every other optional dependency on `DevService` uses its port interface: `imageChecker ports.DockerImageChecker`, `imageInspector ports.ImageInspector`, `imageRemover ports.DockerImageRemover`.

`ports.DockerBuilder` already exists. The `rebuildFakeBuilder` in the test file already implements `ports.DockerBuilder` and is compile-asserted as such (`var _ ports.DockerBuilder = (*rebuildFakeBuilder)(nil)`). The `newBuildService` helper in the test is a thin wrapper that would disappear if the parameter were typed as the interface directly.

Two acceptable fixes:
1. Change the parameter to `builder ports.DockerBuilder` and call `builder.Build` directly in `Rebuild`, passing the `BuildOptions{ImageTag: tag}` inline. The label-stamping logic lives in `BuildService.Run`, not in `DockerBuilder.Build` — so the caller in `cli/cmd/dev.go` would need to pass `opsadapter.NewBuildAdapter()` (the raw adapter, not the service). This is cleaner architecturally.
2. Add a comment on the `builder *BuildService` parameter explaining why the concrete type is intentional: "Rebuild delegates to BuildService.Run rather than DockerBuilder.Build so that the identity-label stamping in BuildService.Run is guaranteed to run." This is a weaker justification (label stamping is in BuildService, not in DockerBuilder), but acceptable if the team prefers keeping BuildService encapsulation.

Either fix is fine. The current state — concrete type, no justification comment — is not.

---

## Issue 2 — Test gap: `TestRebuild_HappyPath` missing `upCalled` assertion (minor, required)

**File:** `internal/app/ops/rebuild_test.go` line 111

The happy path asserts Down/Remove/Build are called but not Up. `TestRebuild_BuildFailure` asserts `upCalled == 0` as the negative case — the positive case must mirror this. The 'Starting stack...' line in stdout proves Rebuild reached the Run call, not that compose.Up fired.

Add `fc.upCalled > 0` check at the end of `TestRebuild_HappyPath`.

---

## What is confirmed correct

**#1219 interaction test confirmed.** `TestRebuild_ForeignLabelBypass` (line 469) is load-bearing and correct:
- Constructs a fake inspector returning a foreign project hash.
- Proves plain `Run` blocks (`blockErr == nil` → fatal).
- Proves `Rebuild` with the same config succeeds, with `fb.buildCalls > 0` confirming the build ran.
- Uses a separate `fiAfterBuild` inspector (current hash) for the `DevService` that `Rebuild` calls into, correctly modelling the post-build state.

**Architecture otherwise sound:**
- `DockerImageRemover` interface in `internal/ports/ops.go` — correct location, idempotent contract documented.
- `ImageRemoveAdapter` stderr classification: "No such image" → nil, all others → wrapped error. The distinction between "daemon broken" and "image not there" is maintained — the adapter wraps the exit error plus stderr in the non-nil path so callers have full context.
- `compose.Down` called directly (not `DownService.Run`) — correct; skips TTY confirmation and `printDownSummary` which do not belong on the rebuild path.
- Flag conflict validation fires before any I/O. Both rejections (`--rebuild --watch`, `--volumes` without `--rebuild`) are tested in `dev_rebuild_test.go`.
- `gofmt` clean, build clean, all tests pass.
- Branch is `feat/1220-vibew-dev-rebuild`, one conventional-commit, targets main. No force push, no `--no-verify`.
- CHANGELOG entry is accurate and the release-gate comment in the Unreleased section is updated correctly.

tibtof:
**Documentation review: APPROVE** (fixes committed in a5de632)

**Verdict: APPROVE**

Six doc surfaces were stale or missing the new flags. All fixed directly in this review pass, committed to the PR branch, and pushed clean through `make check`.

---

**Top 3 findings**

1. **llms-full.txt CLI table missing --rebuild rows.** The table at line 1349 described `vibew dev` recovery as aspirational ('once #1220 ships'). No rows for `vibew dev --rebuild` or `--rebuild --volumes` existed. Fixed: two new rows added; aspirational language removed.

2. **Four docs still said 'will automate when it ships'.** `docs/troubleshooting.md` fix block showed `vibew build && vibew dev` with a link to #1220. `llms-full.txt` troubleshooting section and TRAP notes same. Template and static example AGENTS-VIBEWARDEN.md said 'Recovery: `vibew build && vibew dev`'. All replaced with `vibew dev --rebuild`.

3. **No upgrader onramp in getting-started.md.** Users upgrading from ≤ v0.18.2 hit Variant 2 on every project but the 'when to use which mode' table had no row pointing them at `--rebuild`. Fixed: two rows added for the upgrade and mismatch cases.

---

**Files touched** (in commit a5de632):
- `/Users/tibtof/workspace/vibewarden/llms-full.txt`
- `/Users/tibtof/workspace/vibewarden/docs/troubleshooting.md`
- `/Users/tibtof/workspace/vibewarden/docs/getting-started.md`
- `/Users/tibtof/workspace/vibewarden/README.md`
- `/Users/tibtof/workspace/vibewarden/internal/cli/templates/agents/agents-vibewarden.md.tmpl`
- `/Users/tibtof/workspace/vibewarden/docs/examples/AGENTS-VIBEWARDEN.md`

---

**Pre-existing issue (not blocking, not introduced by #1220)**

The golden-tested error message embeds 'VibeWarden v0.19.0' but the CHANGELOG release tag for this batch is v0.18.3. Naming inconsistency originated in #1219. The error text is golden-tested; changing it requires a separate golden-test update. Flag for follow-up if the release is tagged v0.18.3.

tibtof:
@reviewer addressed: rebuild now takes ports.DockerBuilder (port interface — stronger choice taken); label-stamping is done via new BuildLabels helper in image_identity.go so the invariant is preserved without routing through BuildService; upCalled assertion added to TestRebuild_HappyPath; new TestRebuild_BuildLabelsPresent arch-invariant test asserts both identity labels are in DockerBuildOptions on every rebuild; #1219 error message corrected to v0.18.3 with golden test updated. Ready for re-review.

internal/cli/cmd/dev.go
```diff
@@ -50,11 +53,26 @@ To also start Prometheus and Grafana, run 'vibew obs up' after the stack is up.
 Pass --watch to watch vibewarden.yaml for changes and automatically
 regenerate config files and restart the stack (blocks until Ctrl+C).
 
+Pass --rebuild to stop the stack, remove the app image, rebuild via vibew build,
+and start the stack again. This is the recovery path for the image-identity
+mismatch error introduced in v0.19.0. --rebuild and --watch are mutually exclusive.
```
tibtof: **Version string inconsistency: `v0.19.0` should be `v0.18.3`**

The fix commit (7afe80e) corrected `dev_format.go` and its golden test, but this help-text line was not updated in the same pass.

Line 58 reads:
```
mismatch error introduced in v0.19.0. --rebuild and --watch are mutually exclusive.
```

The actual release shipping this feature is v0.18.3. This string is user-visible via `vibew dev --help` and should match the corrected `dev_format.go` wording.

Fix: change `v0.19.0` → `v0.18.3` on this line.

README.md
```diff
@@ -297,6 +297,7 @@ features after the initial setup.
 | `vibew generate` | Regenerate `docker-compose.yml` from config |
 | `vibew build` | Build the Docker image for the app |
 | `vibew dev` | Start local dev environment |
+| `vibew dev --rebuild` | Stop stack, remove app image, rebuild, start — recovery for image-identity mismatch (v0.19.0+) |
```
tibtof: **Version string inconsistency: `v0.19.0+` should be `v0.18.3+`**

This row was added by the writer agent in commit a5de632 using the version string that was correct at the time of writing but has since been corrected in `dev_format.go` by commit 7afe80e.

The table entry reads:
```
| `vibew dev --rebuild` | ... recovery for image-identity mismatch (v0.19.0+) |
```

The feature ships in v0.18.3, not v0.19.0. Fix: change `v0.19.0+` → `v0.18.3+`.

tibtof:

Re-review of commit 7afe80e.

---

## What is fully resolved

**Issue 1 (BLOCKING) — concrete coupling fixed.** `Rebuild` now takes `ports.DockerBuilder` (line 59 of `internal/app/ops/rebuild.go`). The new `BuildLabels` helper in `internal/app/ops/image_identity.go` (lines 78–96) is the single source of truth for label-stamping on both the `BuildService.Run` path and the `Rebuild` path. `BuildService.Run` in `internal/app/ops/build.go` (line 107) delegates to `BuildLabels` — no logic duplication. CLI wires `opsadapter.NewBuildAdapter()` directly (line 132 of `internal/cli/cmd/dev.go`), no `NewBuildService` wrapper on the rebuild path.

**Issue 2 (test gap) — `upCalled` assertion added.** `TestRebuild_HappyPath` lines 129–132 of `internal/app/ops/rebuild_test.go` assert `fc.upCalled > 0` with an explicit symmetry comment pointing at `TestRebuild_BuildFailure`.

**Issue 3 (writer finding) — `dev_format.go` corrected.** Line 55 of `internal/app/ops/dev_format.go` now reads `v0.18.3` and the golden test at line 50 of `dev_format_test.go` matches exactly.

**Architecture confirmation:**
- `DockerImageRemover` interface is in `internal/ports/ops.go` lines 249–259 — correct location, idempotent contract documented.
- `rebuildFakeBuilder` carries a compile assertion `var _ ports.DockerBuilder = (*rebuildFakeBuilder)(nil)` (line 540 of `rebuild_test.go`).
- `fakeRemover` carries a compile assertion `var _ ports.DockerImageRemover = (*fakeRemover)(nil)` (line 556).
- `TestRebuild_BuildLabelsPresent` (lines 135–176) is the arch-invariant guard asserting both label keys are in `DockerBuildOptions.Labels` on every rebuild.
- `TestRebuild_ForeignLabelBypass` (line 682, diff numbering) still passes — structure unchanged.
- Architecture test at `test/architecture/image_labels_test.go` still asserts `BuildService.Run` stamps both keys regardless of tag resolution path.

---

## Remaining finding (BLOCKING-MINOR)

**`v0.19.0` not fully swept — two surfaces missed.**

The fix commit updated `dev_format.go` and its golden test but did not update two other surfaces where `v0.19.0` is user-visible:

1. `internal/cli/cmd/dev.go` line 58 — the `--rebuild` help text paragraph says "mismatch error introduced in v0.19.0". This string is printed by `vibew dev --help` and contradicts the corrected `dev_format.go`. Inline comment posted above.

2. `README.md` line 300 — the command table row says "recovery for image-identity mismatch (v0.19.0+)". Inline comment posted above.

**Note:** `llms-full.txt` line 1638 also contains `v0.19.0` inside the embedded Variant 2 error message replica, but that line was introduced by the already-merged #1219 PR (commit 9331506) and is outside this PR's diff. It is pre-existing technical debt, not a blocker for this PR.

---

## Fix required

In a single commit:
- `internal/cli/cmd/dev.go` line 58: `v0.19.0` → `v0.18.3`
- `README.md` line 300: `(v0.19.0+)` → `(v0.18.3+)`

Issue #1221, opened 2026-04-29
feat(retro-0.18.2): vibew logs [--tail N] [<service>] — local dev-stack logs wrapper


tibtof:
## Retro evidence

From the v0.18.2 retrospective (qr-code-blackhole Node/Express deploy):

> `vibew logs --tail 60` doesn't exist either — I had to fall back to `docker logs`.
> Used raw docker for `docker logs <container>` (because `vibew logs --tail` doesn't exist)

## Constitutional check

This is **local stack ergonomics** (the `vibew dev` compose project), NOT remote orchestration. Does not conflict with the post-#1138 / ADR-086 stance against remote orchestration; that line stays drawn.

## Severity

**HIGH** when paired with #1225 (image-tag debugging starts with logs). Standalone, MEDIUM. Filed as HIGH because the v0.18.2 retro flagged raw `docker logs` as a bypass that vibew should cover.

## Acceptance criteria

- [ ] New CLI: `vibew logs [--tail N] [--follow] [--since <duration>] [<service>]`.
- [ ] **Default behavior** (no positional service arg): show all services interleaved, mirroring `docker compose logs` default.
- [ ] **Positional service arg** (`vibew logs vibewarden`, `vibew logs app`): scopes to that service.
- [ ] `--tail N` defaults to 100. `--follow` / `-f` streams. `--since <duration>` accepts `5m`, `1h`, RFC3339 timestamps.
- [ ] Implementation: shell out to `docker compose -p <project-name> -f <resolved-compose-path> logs <flags>`. Reuse the project-name resolution from `internal/config/config.go::ComposeProjectName` (post-#1199).
- [ ] Failure modes:
  - Stack not running: `Stack is not running. Start with: vibew dev` → exit 1.
  - Unknown service name: `Unknown service '<x>'. Known: vibewarden, app, ...` (list the services from the compose project).
  - Docker not available: reuse `ports.ErrDockerUnavailable` → exit 3.
- [ ] Help text covers all four flags + a one-line tip directing the agent to `vibew logs vibewarden` and `vibew logs app`.
- [ ] Test: integration test that exercises positional service selection, `--tail`, and stack-not-running error path.

## Out of scope

- Streaming logs from a remote host (rejected at retro #9).
- `--grep` / filtering — `docker compose logs ... | grep` is the right shape.
- Log persistence / backfill / file output.

## Source

Retro: `~/notes/vibewarden/retro-0.18.2.md` (2026-04-28)

[...]

Issue #1198, opened 2026-04-28
chore(retro-0.18.1): drop misleading 'upstream unreachable' check from vibew doctor


tibtof:
## Retro evidence

> `Upstream reachable: http://0.0.0.0:3000 — unreachable (app may not be started yet)` — fires before `vibew dev`. From the host laptop, port 3000 isn't bound to anything because the app runs inside Docker behind the sidecar; it's never bound to the host. This warning is technically correct but pedagogically misleading: implies "your app is broken" when really "you haven't run dev yet."

## Problem

The upstream port lives on the docker-compose internal network. From the host shell, it's *never* reachable directly. The probe is asking the wrong question. After the health endpoint fix lands, `_vibewarden/health` is the runtime signal for upstream health.

## Acceptance criteria

- [ ] Remove the upstream-reachable check from `vibew doctor` entirely.
- [ ] `vibew doctor` is now strictly about config + Dockerfile + toolchain — not runtime health.
- [ ] Doctor docs / help text updated to reflect the narrower scope.
- [ ] `_vibewarden/health` (post-companion-fix) is the canonical runtime signal; mention this in doctor docs.

## Scope notes

- **Depends on the upstream-health fix landing first** (companion issue in this retro chain).
- Order matters: do not remove the doctor check until the runtime probe is in place.

## Severity

**MEDIUM.** Confusing, not broken. Fixed by deletion.

## Source

Retro: `~/notes/vibewarden/retro-0.18.1.md` (2026-04-28)

tibtof:
Status: READY_FOR_DEV

No ADR — pure deletion + docs touchup. Decisions captured here.

## Files to modify (≤5)

1. `internal/app/ops/doctor.go` — delete `checkUpstreamReachable` method, the `results = append(...)` call site (line 215), and the `healthChecker ports.HealthChecker` field + constructor parameter. After this change, the only Layer-2 check is `checkTLSCertValid` — `Local Runtime` section header stays valid.
2. `internal/app/ops/doctor_test.go` — delete `TestDoctorService_Run_UpstreamReachable` and `TestDoctorService_Run_UpstreamUnreachable` (lines 555–604). Delete the `reachableHealthChecker()` / `unreachableHealthChecker()` helpers (lines 56–70). Update all 38 `ops.NewDoctorService(fc, pc, hc)` call sites to `ops.NewDoctorService(fc, pc)` and drop the `hc := reachableHealthChecker()` line above each.
[...]

PR #1228, opened 2026-04-29
fix(#1222): doctor skips runtime checks pre-vibew dev (mirror of #1198)

### LINKED ISSUE #1222: fix(retro-0.18.2): vibew doctor fires 3 spurious WARNs pre-vibew dev (generated files, container health, TLS cert)
## Retro evidence

> `vibew doctor` warnings before stack was running ("Generated files not found", "Container health: stack may not be running", "TLS certificate: sidecar not reachable"). These are technically correct but they fired before I ran `vibew dev`, which is exactly when an agent runs doctor. They should be downgraded to INFO or skipped pre-`vibew dev`. A new user reading three WARNs on a green-field project will assume something is wrong.

## Background

Same misleading-warn class as #1198 (which we already fixed by deleting the upstream-reachable check from doctor). Three more siblings live on:

1. **Generated files** — `.vibewarden/generated/docker-compose.yml not found — run 'vibew generate' first`
2. **Container health** — `could not query containers — stack may not be running`
3. **TLS certificate** — `sidecar not reachable — start 'vibew dev'`

All three are correct *post-stack-start* but produce misleading WARNs *pre-stack-start* — exactly when an agent runs doctor as part of the kickoff flow.

## Severity

**HIGH.** Same class as #1198. New users / agents read three WARNs on a green-field project and assume something is wrong.

## Acceptance criteria

- [ ] Detect whether the dev stack is running. Recommended approach: run `docker compose -p <app.name> ps -q` and check non-empty output. (Do not depend on `.vibewarden/generated/` existence — that lies if the user ran compose manually.)
- [ ] **When stack is down:** skip all three checks entirely. They are not displayed at any level (no INFO either — empty section is preferred over noise).
- [ ] **When stack is up:** keep TLS certificate + Generated files checks at WARN level (real signal — TLS broken cert or stale generated state).
- [ ] **Delete** the Container health check entirely. After #1197, `_vibewarden/health` (HTTP query against the running sidecar) is the canonical runtime signal for "is the stack actually serving." The Container health doctor check is redundant with that runtime signal.
- [ ] Update `docs/troubleshooting.md` to reflect the narrower scope (matching the #1198 cleanup pattern).
- [ ] Update `internal/cli/templates/agents/agents-vibewarden.md.tmpl` if it lists doctor's checks.
- [ ] Update `llms-full.txt` doctor checks list.
- [ ] Test: with stack down, doctor stdout does not contain "Generated files", "Container health", or "TLS certificate" rows. With stack up (mock or integration), the surviving two render normally.

## Out of scope

- Same-class checks beyond these three. If a fourth pre-stack check is later flagged, separate issue.
- Re-introducing the upstream-reachable check (#1198) under a stack-state-aware flag — explicitly rejected, `_vibewarden/health` covers it.

## Source

tibtof:
Closes #1222

## Summary

- **Deleted** `checkContainerHealth` entirely — runtime container liveness is covered by `/_vibewarden/health` since #1197. The check produced spurious WARNs pre-stack and was redundant post-stack.
- **Added** `isStackRunning(ctx, runner, composeFile) bool` — package-private helper using `ComposeRunner.PS` (same port, same detection path as `LogsStreamService.Run` from #1221). Errors map to `false` (defensive).
- **Gated** `checkGeneratedFiles` on `stackUp`: silently skipped pre-stack, runs normally post-stack.
- **Gated** `checkTLSCertValid` on `stackUp`: silently skipped pre-stack (no misleading "sidecar not reachable — start vibew dev" WARN before the user has run vibew dev), runs normally post-stack.
- Updated `vibew doctor --help` Long text to reflect the narrower pre-stack scope.
- Updated `docs/troubleshooting.md`: new check table, updated sample outputs (Stack not yet started, All checks pass, Port conflict, JSON), removed Container health from all examples, replaced the Kratos/Postgres Container health symptom sections with vibew-logs-based equivalents.
- Updated `internal/cli/templates/agents/agents-vibewarden.md.tmpl` L117: removed "container health," from the doctor checks sentence.
- Updated `llms-full.txt`: dropped check #6 (Container health), renumbered, split into Static vs Stack-state-gated.
- Added CHANGELOG `[Unreleased] → Changed` entry.

**Before (stack down):**
```
  [OK]    Config file       vibewarden.yaml — valid
  [OK]    Docker daemon     running
  [OK]    Docker Compose    Docker Compose version v2.27.0
  [OK]    Proxy port        port 8443 is available
  [WARN]  Generated files   .vibewarden/generated/docker-compose.yml not found — run 'vibew generate' first
  [WARN]  Container health  no containers found — run 'vibewarden dev' to start the stack
  [WARN]  TLS certificate   sidecar not reachable — start 'vibew dev'
```

**After (stack down):**
```
  [OK]  Config file      vibewarden.yaml — valid
  [OK]  Docker daemon    running
  [OK]  Docker Compose   Docker Compose version v2.27.0
  [OK]  Proxy port       port 8443 is available
  [OK]  ACME email       not using ZeroSSL — email not required
```

Three fewer rows, zero misleading WARNs. Same class of fix as #1198.

## Test plan

- `TestDoctor_StackDown_SkipsGenAndTLS` (new): stack down → Generated files, Container health, TLS certificate all absent; Config + Docker static checks present.
- `TestDoctor_StackUp_AllOK` (new): stack up + generated file → Generated files present; Container health absent (regression guard against re-introducing the deleted check).
- `TestDoctor_StackUp_GeneratedFileMissing_StillWarns` (new): stack up but file absent → [WARN] Generated files row still appears (gate doesn't suppress the real signal).
- `TestDoctor_ContainerHealth_AbsentInJSON` (new): JSON mode, healthy stack → no result has `Name == "Container health"`.
- `TestDoctorService_Run_AllPassing` (updated): removed Container health from wanted strings, added regression guard asserting it is absent.
- `TestDoctorService_ChecksAreIndependent` (updated): dropped Container health + Generated files from wanted strings (both absent when PS fails → stack down).
- `TestDoctorService_Run_GeneratedFileMissing_IsWarn` (updated): switched to `healthyContainersCompose()` so the gated check runs.
- All TLS cert tests (updated): switched to `healthyContainersCompose()` so the gated TLS check runs.
- `TestDoctorService_Run_UnhealthyContainer_IsFail` (deleted): covers the deleted check.
- `TestDoctorService_Run_ContainersHealthy_AllOK` (deleted): covers the deleted check; replaced by `TestDoctor_StackUp_AllOK`.
- `make check` clean — golangci-lint 0 issues, race tests pass.


llms-full.txt
```diff
@@ -1554,32 +1554,33 @@ vibew doctor exit codes:
...
+     13. Dockerfile: non-root USER — final stage must set non-root USER (WARN, non-blocking)
+     14. Dockerfile: multi-stage build — compiled languages (Go) need multi-stage (FAIL if not)
+     15. Dockerfile: <lang> toolchain version — builder FROM tag major.minor must match
          the project manifest go.mod/nvmrc/pyproject.toml (FAIL on mismatch; OFF when
          tag is "latest" or unrecognised, or when no manifest is found)
      All 6 checks are omitted when no Dockerfile is present in the project root.
-  Runtime upstream health is NOT checked by doctor — query /_vibewarden/health
-  after vibew dev is up for live upstream status.
+  Runtime container health is NOT checked by doctor — query /_vibewarden/health
+  after vibew dev is up for live runtime status.
```
tibtof, line 1583: **Doc/code discrepancy**: Dockerfile contract checks (10-15) appear visually indented under the `Stack-state-gated` section header, but they are NOT gated on stack state — `checkDockerfile` is called unconditionally in `runChecks` (`doctor.go:224`), regardless of `stackUp`. See the `Stack-state-gated` label on the line two above this range.

An LLM agent reading this file will incorrectly infer that Dockerfile checks are skipped pre-stack. That is wrong and could cause an agent to misdiagnose a Dockerfile contract failure as a stack-state issue.

**Fix**: Move items 10-15 into a separate third category, after the `Stack-state-gated` block:

```
  Stack-state-gated (skipped silently when vibew dev is not running):
  8. Generated files — .vibewarden/generated/docker-compose.yml present
  9. TLS certificate — live TLS handshake; reports expiry state
  Dockerfile contract (only when a Dockerfile is present in the project root):
  10-15. ...
```

The `docs/troubleshooting.md` and `internal/cli/cmd/doctor.go` Long text are correct (Dockerfile contract is described separately or omitted from the stack-gated list). Only `llms-full.txt` has the wrong grouping.

internal/app/ops/doctor_test.go
```diff
@@ -692,9 +669,10 @@ func TestDoctorService_Run_SectionHeaders(t *testing.T) {
 	if !strings.Contains(out, "Config & Docker") {
 		t.Errorf("expected 'Config & Docker' section header, got:\n%s", out)
 	}
-	if !strings.Contains(out, "Local Runtime") {
-		t.Errorf("expected 'Local Runtime' section header, got:\n%s", out)
-	}
+	// "Local Runtime" only renders when the stack is up and a TLS row is emitted.
```
tibtof: **Misleading comment — will confuse next dev**: The comment says `"Local Runtime" only renders when the stack is up and a TLS row is emitted. With an external provider (doctorConfig sets Provider="external"), the TLS check is skipped inside checkTLSCertValid — so Local Runtime may be absent.`

This is factually wrong. When `stackUp` is true and `Provider == "external"` and `tlsState == nil`, `checkTLSCertValid` hits the `Provider != "self-signed"` short-circuit at `doctor.go:391` and returns an OK row (`"provider is \"external\" — skipping local cert check"`). That row is emitted with `sectionLocalRuntime`, so the header WILL render.

The comment's claim that the section header assertion was dropped because the section "may be absent" is incorrect. The real reason is that the assertion was dropped because `TestDoctor_StackUp_AllOK` was meant to cover it — but `TestDoctor_StackUp_AllOK` does not assert `"Local Runtime"` is present either, leaving a gap.

**Fix**: Either:
(a) Add `if !strings.Contains(out, "Local Runtime") { t.Errorf(...) }` to `TestDoctorService_Run_SectionHeaders` and fix the comment to explain the section always appears when stack is up, OR
(b) Add the assertion to `TestDoctor_StackUp_AllOK` and update this comment to say `"Local Runtime header assertion moved to TestDoctor_StackUp_AllOK."`

Option (b) is preferred since the test name better describes the invariant.

tibtof:

## Summary

The core logic is solid. `checkContainerHealth` is cleanly deleted, `isStackRunning` correctly reuses `ports.ComposeRunner.PS`, the two runtime checks are gated correctly, and the four new test functions cover the required contracts. Three issues need fixing before merge.

---

## Finding 1 — BLOCKING: `llms-full.txt` places Dockerfile contract checks under the wrong heading

**File**: `llms-full.txt`, line 1575 onward.

Checks 10-15 (Dockerfile contract) appear visually indented under the `Stack-state-gated` heading, but `checkDockerfile` is called unconditionally in `runChecks` at `doctor.go:224` — it is NOT gated on `stackUp`. An LLM agent reading `llms-full.txt` will believe Dockerfile checks are skipped pre-stack. That is incorrect.

`docs/troubleshooting.md` and the Long help text are both correct (they separate Dockerfile from the stack-gated list). Only `llms-full.txt` has the wrong grouping.

**Fix**: Break the `Stack-state-gated` block after item 9 and start a separate third category:

```
  Stack-state-gated (skipped silently when vibew dev is not running):
  8. Generated files — .vibewarden/generated/docker-compose.yml present
  9. TLS certificate — live TLS handshake; reports expiry state
  Dockerfile contract checks (only when a Dockerfile is present in the project root):
  10. Dockerfile: alpine base ...
  ...
  All 6 checks are omitted when no Dockerfile is present in the project root.
```

---

## Finding 2 — BLOCKING: Missing CLI Long text regression guard in `internal/cli/cmd/doctor_test.go`

The spec explicitly required a guard asserting `doctor.Long` does not contain `"container health"` (case-insensitive), mirroring the existing `TestDoctorCmd_LongDescription_NoUpstreamReachable` from #1198. The file was not modified in this PR and the guard was not added.

**Fix**: Add to `internal/cli/cmd/doctor_test.go`:

```go
// TestDoctorCmd_LongDescription_NoContainerHealth is a regression guard that
// ensures the deleted Container health check (removed in #1222) is not
// re-added to the help text. Runtime container health is at /_vibewarden/health.
func TestDoctorCmd_LongDescription_NoContainerHealth(t *testing.T) {
	root := cmd.NewRootCmd("test")
	doctorCmd, _, _ := root.Find([]string{"doctor"})

	if strings.Contains(strings.ToLower(doctorCmd.Long), "container health") {
		t.Error("doctor Long description must not mention 'container health' (deleted check, #1222)")
	}
}
```

---

## Finding 3 — NON-BLOCKING: `TestDoctorService_Run_SectionHeaders` drops the `"Local Runtime"` assertion with a factually wrong justification

The comment says `"Local Runtime may be absent"` when `doctorConfig` uses `Provider="external"`. This is incorrect. With `stackUp` true and `Provider="external"`, `checkTLSCertValid` hits the non-self-signed short-circuit and still returns an OK row tagged `sectionLocalRuntime` — so the header always renders. The assertion was dropped with a wrong explanation, and `TestDoctor_StackUp_AllOK` (which was cited as the coverage) does not assert the header either. Net: no test asserts `"Local Runtime"` appears when the stack is up.

Fix: add the assertion to `TestDoctor_StackUp_AllOK` and replace the comment with: `// Local Runtime section header assertion is in TestDoctor_StackUp_AllOK.`

This is non-blocking because the production logic is correct — it is a test coverage gap and a misleading comment, not a bug.

---

## What is correct (no action needed)

- `isStackRunning` uses `ComposeRunner.PS` consistently with #1221. The 5-second sub-context timeout is intentional and better than #1221's unbounded pass-through — the comment could say "adds a bounded timeout" rather than "mirrors", but this is cosmetic.
- `TestDoctorService_Run_UnhealthyContainer_IsFail` and `TestDoctorService_Run_ContainersHealthy_AllOK` are replaced by intentional tombstone comments with #1222 references. Accepted.
- All four required new tests (`TestDoctor_StackDown_SkipsGenAndTLS`, `TestDoctor_StackUp_AllOK`, `TestDoctor_StackUp_GeneratedFileMissing_StillWarns`, `TestDoctor_ContainerHealth_AbsentInJSON`) are present and correct.
- `_vibewarden/health` is referenced correctly in docs without claiming a specific wire format (no false claim of "just ok").
- Architecture is clean: no new port methods, no new adapters, pure package-private helper.
- Branch is `fix/1222-doctor-stack-state-aware-skip`. Branch naming, commit style, PR title all correct.

tibtof:
## Documentation review — APPROVE (with fixes committed)

Two issues found and fixed in commit 91f9cc8 on this branch. All other docs verified correct.

---

### Fixes committed

**1. `docs/examples/AGENTS-VIBEWARDEN.md` — stale doctor descriptions (two spots)**

Lines 102-109 and 266: both still listed "container health" as an always-run check and placed "generated files" alongside it without a stack-gating qualifier. This is the static example file that ships as a reference for AI agents, so the stale content would have taught agents the wrong pre-stack behavior.

Updated to: static checks always run; generated-files and TLS are stack-gated; container health is deleted (use `/_vibewarden/health`). Added canonical health response vocabulary (`"ok"/"failing"` per #1197) so agents get the current wire format.

**2. `llms-full.txt` — Dockerfile checks miscategorized as stack-gated**

Items 10-15 (Dockerfile contract checks) were placed under the "Stack-state-gated" section header, implying they are skipped pre-stack. `doctor.go` line 224 shows they are unconditional (`results = append(results, s.checkDockerfile(...)...)`). Added a separate "Dockerfile contract checks (always run when a Dockerfile is present)" header.

---

### Verified correct — no changes needed

- `vibew doctor --help` Long text: pre-stack scope correct, no Container health, `_vibewarden/health` pointer present.
- `docs/troubleshooting.md`: Container health row removed, Generated files + TLS annotated as stack-gated, `vibew logs` recommended for runtime diagnosis, no stale `"healthy"`/`"unhealthy"` vocabulary.
- `internal/cli/templates/agents/agents-vibewarden.md.tmpl` L117: "container health" dropped, pre/post-stack split visible.
- `llms-full.txt` doctor checks: Container health absent, static vs stack-gated split present.
- CHANGELOG `[Unreleased] → Changed`: entry present, specific, references #1197 for container-health deletion and #1198 for same-class fix.
- No upstream-reachable re-introduction (#1198 regression).
- No claim that doctor probes image-identity (belongs to `vibew dev` per #1219).
- `_vibewarden/health` vocabulary throughout uses post-#1197 terms (`"ok"`/`"failing"`/`"unknown"`); zero stale `"healthy"`/`"unhealthy"` strings found in any doc file.

tibtof:
@reviewer addressed: TestDoctorCmd_LongDescription_NoContainerHealth regression guard added (mirrors #1198's pattern); Local Runtime header assertion restored in TestDoctorService_Run_SectionHeaders. Writer's commit 91f9cc8 already addressed the llms-full.txt grouping. Ready for re-review.

Issue #1146, opened 2026-04-26
fix: vibew bundle STALE warning fires on byte-identical rebuilds (mtime-based, should be content-based)


tibtof:
## Problem

From the retro:

> "STALE" warning fires on bundle even when the resulting image digest is byte-identical. Touching `vibewarden.yaml` after a build invalidates by mtime, not by content. Made me do a redundant rebuild that produced the same digest.

The current staleness walk (ADR-089, #1085) compares source mtimes against `image.tar` mtime. That's a noisy proxy: editing a YAML and saving it bumps mtime even if the diff is whitespace.

## Proposal

When the staleness walk detects mtime-newer sources, **don't immediately warn** — instead:

1. Compute a content hash of the relevant inputs (the merged config, the Dockerfile, the watched source paths).
2. Compare against a hash stored from the previous successful bundle (maybe `.vibewarden/bundle/.input-digest`).
3. Only warn if the content hash actually changed.

If the hash file is missing (first bundle, post-clean), fall back to the mtime warning (current behavior).

## Acceptance criteria

- Bundle a project, `touch vibewarden.yaml` (no content change), bundle again — no STALE warning.
- Bundle a project, change one byte in a watched source file, bundle again — STALE warning fires.
- Hash file lives under `.vibewarden/` and is documented in the bundle README.
- Tests cover both paths.

## Anchored by (policy reference)

- CLAUDE.md §Architecture principles → Artifact policy (see #1137 / PR #1152 — same trust-poisoning middle-ground pattern as deploy.sh).

## Source

- `~/notes/vibewarden/retro-0.17.0.md`, retro ugly #1.

tibtof:
## PM Spec — fix(#1146): bundle staleness uses content-hash, not mtime

### Problem statement

`vibew bundle` emits a STALE warning whenever source file mtimes are newer than `image.tar`, even when the rebuild would produce a byte-identical image. Touching `vibewarden.yaml` without changing content bumps mtime and fires the warning — a false positive that erodes trust in the signal (qr-dali retro §Ugly #1; CLAUDE.md §Artifact policy, trust-poisoning middle-ground pattern).

### User story

As a vibe coder, I want the STALE warning to fire only when my source inputs have actually changed content, so that I can trust it and skip redundant rebuilds.

### Context: what the current code does

[...]

PR #1229, opened 2026-04-29
fix(#1223): bundle no longer fires STALE on fresh build (self-induced .gitignore mutation)

### LINKED ISSUE #1223: fix(retro-0.18.2): vibew bundle STALE false positive immediately after fresh build
## Retro evidence

> `vibew bundle` STALE warning: "1 source files changed since image was built". Suggestion was clear (`--allow-stale` or rebuild). But I had just built the image seconds before — something (probably config regeneration) was touching a file that bumps the mtime. Felt like a false positive; rebuilding fixed it.

## Mechanism (suspected)

Bundle's freshness check (ADR-089 §Refinement, #1146) compares input mtimes against the image build time. The bundle pipeline writes/regenerates `.vibewarden/generated/docker-compose.yml`, the merged `vibewarden.yaml`, possibly `.env`. If the freshness check walks `<projectRoot>` and includes vibew-generated files, **bundle's own write triggers the warn on a fresh build**. Self-induced staleness.

## Severity

**MEDIUM.** Annoying false positive; defeats the freshness signal value. Workaround documented (`--allow-stale` or rebuild).

## Investigation step (architect please reproduce first)

Before designing the fix:
1. Reproduce locally — fresh project, `vibew build && vibew bundle`, no edits between.
2. Dump the watched-input set actually used by the freshness check.
3. Identify which file's mtime is newer-than-image after `vibew build`.
4. Document the culprit in the status comment so the dev fix is data-driven, not assumption-driven.

The retro's "1 source files changed" suggests the warn already knows the count — verify whether the offending path is currently printed; if not, that's a separate UX nit to fix in the same PR.

## Acceptance criteria

- [ ] Identify and document the offending file(s) in the architect's status comment.
- [ ] Switch freshness comparison from mtime-of-input to **content-hash digest** of input. Already noted as an ADR-089 refinement and partially landed via #1146; verify and complete the migration if not fully active in this code path.
- [ ] **Prune the watched input set** to user-source only. Deny-list at minimum: `.vibewarden/`, `.git/`, `node_modules/`, `vendor/`, `dist/`, `bin/`, `build/`, `.next/`, `target/`. Honor `.gitignore` and `.dockerignore` where present. The watched set should reflect "what goes into the Docker build context," not "every file under projectRoot."
- [ ] Warn message **prints the offending path(s)** when stale: `1 source file changed since image was built: src/index.ts (rebuild or pass --allow-stale)`.
- [ ] Test: `vibew build && vibew bundle` on a fresh project produces NO stale warning. Touching a file in the watched set after build → warn fires with the correct path. Touching `.vibewarden/generated/foo.yml` after build → warn does NOT fire (excluded path).
- [ ] No ADR (existing ADR-089 governs freshness; this is implementation refinement).

## Out of scope

- Replacing the freshness check entirely with a different signal (e.g. Docker image labels, content-addressed image tags).
- Multi-arch / manifest-list freshness — single-arch only here; #1200 governs arch.

## Source

tibtof:
Closes #1223

## Root cause

`WriteInputDigest` called `ensureGitIgnored` which opened the user's `.gitignore` for append on every bundle run, bumping its mtime even when the line was already present. Go's `O_APPEND|O_WRONLY` updates mtime on open. The bumped mtime was then captured by `FileSystemStalenessWalker.NewestMTime` on any subsequent bundle whose digest path fell back to mtime (first run, wiped `.vibewarden/`, corrupt digest, schema-version mismatch), producing a false STALE verdict.

Architect's comment confirming root cause and design: https://github.com/VibeWarden/vibewarden/issues/1223

## Before / after

**Before (second bundle after a clean build, no edits):**
```
  Freshness:    STALE — 1 source files changed since image was built
```

**After:**
```
  Freshness:    FRESH
```

**After (when files actually change):**
```
  Freshness:    STALE — source files changed since image was built:
                  - main.go (modified)
                  - newfile.txt (added)
```

## Summary

- **Drop `.gitignore` mutation.** `ensureGitIgnored`, `readGitIgnoreLines`, and `inputDigestGitIgnoreLine` are removed. `WriteInputDigest` now calls `ensureVibewardenGitignoreFile` which writes `<projectRoot>/.vibewarden/.gitignore` containing `*\n` (idempotent — no touch if content already matches).
- **Digest schema v1 → v2.** The new schema adds a `"files": [{path, sha256}]` array per-file. v1 files are treated as missing on read (first-run baseline FRESH, not mtime fallback). No mtime path exists after this PR.
- **Drop `FreshnessModeTime` / mtime fallback end-to-end.** No prior digest → `Mode=FreshnessModeBaseline`, `Stale=false`. The `FreshnessModeTime` constant is kept for API compatibility but will never be set by `CheckImageHealth`.
- **Render up to 5 changed paths** (removed → added → modified, alphabetical within each bucket). `(and N more)` suffix when truncated.
- **Add `EvalSymlinks`** to the digest walker; symlinks that escape `projectRoot` are skipped.
- **Add `bin` and `.next`** to `hardIgnoreDirs`.

## Digest schema v1 → v2 migration

First `vibew bundle` after upgrade: existing v1 file is treated as missing → `Mode=baseline` → FRESH. Digest file is rewritten in v2 format. Second bundle: digest comparison is active. No false positive at any step.

## Test plan

All 6 architect-required tests added and passing:

- `TestDigest_GitignoreNotMutatedByWriteInputDigest` — regression: `.gitignore` mtime and content unchanged after `WriteInputDigest` (FAILS on pre-fix code)
- `TestDigest_SelfInducedStalenessRegression` — end-to-end: two consecutive bundle cycles with no edits both report FRESH
- `TestDigest_StalePathsListedInRender` — rendered block contains `added`/`removed`/`modified` labels
- `TestDigest_ChangedPathsCappedAt5` — 12 modified files → 5 shown + `(and 7 more)`
- `TestDigest_VibewardenGitignoreFileWritten` — `.vibewarden/.gitignore` written with `*\n`, idempotent on second call
- `TestDigest_SymlinkOutsideRootSkipped` — symlink escaping project root not included in digest

`make check` (golangci-lint + build + `go test -race ./...` + demo-app) passes.

### DIFF
--- a/internal/app/bundle/input_digest.go
+++ b/internal/app/bundle/input_digest.go
@@ -1,7 +1,6 @@
 package bundle
 
 import (
-	"bufio"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
@@ -18,36 +17,69 @@ import (
 const inputDigestFileName = ".vibewarden/.input-digest"
 
 // inputDigestSchemaVersion is the schema version written to and required when
-// reading the digest file. A file carrying a different version is treated as
-// corrupt and triggers a mtime fallback.
-const inputDigestSchemaVersion = 1
-
-// inputDigestGitIgnoreLine is the line appended to the project .gitignore on
-// first write of the digest file. File-specific entry (not the whole
-// .vibewarden/ dir) so users retain visibility of any future state files.
-const inputDigestGitIgnoreLine = ".vibewarden/.input-digest"
+// reading the digest file. Files with a different version are treated as
+// missing (first-run baseline), never as an mtime fallback.
+//
+// History:
+//
+//	v1 — single "digest" field, no per-file hashes. Dropped in #1223.
+//	v2 — per-file SHA-256 in "files" array; supports changed-path rendering.
+const inputDigestSchemaVersion = 2
 
 // InputDigest is the JSON structure stored at .vibewarden/.input-digest.
 //
 // It is machine-only state: never commit, never share across machines.
 // schema_version gates future format changes. digest is the SHA-256 content
-// hash over the sorted input set. inputs is the sorted list of forward-slash
-// relative paths that contributed to the digest (diagnostic only — not
-// re-verified on read).
+// hash over the sorted input set. files is the sorted list of {path, sha256}
+// entries that contributed to the digest — used to render the changed-path
+// list in the freshness warning.
 type InputDigest struct {
-	SchemaVersion int      `json:"schema_version"`
-	Digest        string   `json:"digest"`
-	Inputs        []string `json:"inputs"`
+	SchemaVersion int             `json:"schema_version"`
+	Digest        string          `json:"digest"`
+	Files         []InputFileHash `json:"files"`
 }
+
+// InputFileHash is one entry in InputDigest.Files. Path is a forward-slash
+// relative path from the project root. Hash is "sha256:<64 hex digits>".
+type InputFileHash struct {
+	Path string `json:"path"`
+	Hash string `json:"hash"`
+}
+
+// ChangedPathKind classifies the type of change detected between two digests.
+type ChangedPathKind string
+
+const (
+	// ChangedPathAdded means the file was not present in the prior digest.
+	ChangedPathAdded ChangedPathKind = "added"
+	// ChangedPathRemoved means the file was present in the prior digest but is
+	// gone in the current digest.
+	ChangedPathRemoved ChangedPathKind = "removed"
+	// ChangedPathModified means the file exists in both digests with different
+	// content hashes.
+	ChangedPathModified ChangedPathKind = "modified"
+)
+
+// ChangedPath records a single file change between two digest snapshots.
+type ChangedPath struct {
+	// Path is the project-root-relative forward-slash path of the changed file.
+	Path string
+	// Kind classifies the change.
+	Kind ChangedPathKind
+}
+
+// maxChangedPathsRendered is the maximum number of changed paths printed in
+// the freshness block. Paths beyond this limit are summarised as "(and N more)".
+const maxChangedPathsRendered = 5
 
 // readInputDigest reads and validates the digest file at
 // <projectRoot>/.vibewarden/.input-digest.
 //
 // Returns (digest, true) when the file exists, parses as valid JSON, has
-// schema_version == 1, and carries a "sha256:<64-hex>" digest value.
-// Returns ("", false) in all other cases (missing, corrupt, wrong version,
-// malformed digest) — callers fall back to the mtime walker. Failures are
-// logged at debug level only; they must not surface as errors.
+// schema_version == 2, and carries a "sha256:<64-hex>" digest value.
+// Returns (InputDigest{}, false) in all other cases (missing, corrupt, wrong
+// version, malformed digest) — callers treat this as first-run baseline
+// (FRESH, no prior comparison). Failures are logged at debug level only.
 func readInputDigest(projectRoot string) (InputDigest, bool) {
 	path := filepath.Join(projectRoot, inputDigestFileName)
 	data, err := os.ReadFile(path) //nolint:gosec // path derived from project root
@@ -65,7 +97,8 @@ func readInputDigest(projectRoot string) (InputDigest, bool) {
 	}
 
 	if d.SchemaVersion != inputDigestSchemaVersion {
-		slog.Debug("input-digest: unsupported schema version", "path", path, "version", d.SchemaVersion)
+		slog.Debug("input-digest: unsupported schema version — treating as missing (first-run baseline)",
+			"path", path, "version", d.SchemaVersion, "supported", inputDigestSchemaVersion)
 		return InputDigest{}, false
 	}
 
@@ -80,10 +113,9 @@ func readInputDigest(projectRoot string) (InputDigest, bool) {
 // writeInputDigest writes the digest to
 // <projectRoot>/.vibewarden/.input-digest.
 //
-// A write failure is logged at warn level but does not fail the caller — the
-// next bundle run will fall back to mtime, which is the documented degraded
-// behaviour. Callers are responsible for updating .gitignore before calling
-// this function so that .gitignore participates in the digest consistently.
+// A write failure is logged at warn level but does not fail the caller. The
+// next bundle run will treat the missing file as a first-run baseline (FRESH),
+// which is the correct degraded behaviour.
 func writeInputDigest(projectRoot string, d InputDigest) {
 	vibewardenDir := filepath.Join(projectRoot, ".vibewarden")
 	if err := os.MkdirAll(vibewardenDir, 0o750); err != nil {
@@ -106,7 +138,8 @@ func writeInputDigest(projectRoot string, d InputDigest) {
 
 // computeInputDigest walks projectRoot using the same ignore rules as the
 // FileSystemStalenessWalker and computes a SHA-256 content digest over the
-// sorted set of non-ignored files.
+// sorted set of non-ignored files. The returned InputDigest carries a per-file
+// hash for each walked file (schema v2).
 //
 // Algorithm (deterministic, platform-independent):
 //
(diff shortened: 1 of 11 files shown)


internal/app/bundle/input_digest.go
```diff
@@ -170,25 +213,38 @@ func computeInputDigest(projectRoot string) (InputDigest, error) {
...
+		if d.Type()&os.ModeSymlink != 0 {
+			resolved, err := filepath.EvalSymlinks(path)
+			if err != nil {
+				slog.Debug("input-digest: cannot resolve symlink, skipping", "path", path, "err", err)
+				return nil
+			}
+			// Skip if resolved target escapes project root.
+			if !strings.HasPrefix(resolved, absRoot) {
+				slog.Debug("input-digest: symlink escapes project root, skipping",
+					"path", path, "resolved", resolved)
```
tibtof, line 240: **Bug: symlink escape-check uses HasPrefix without path separator — path traversal false positive**

The current check is:
```go
if !strings.HasPrefix(resolved, absRoot) {
```
This fires correctly when `absRoot = "/tmp/proj"` and `resolved = "/tmp/other/file"`, but it silently admits a symlink when `absRoot = "/tmp/proj"` and `resolved = "/tmp/proj-other/secret.txt"` — `proj-other` starts with `proj`.

Fix: anchor the check to a path separator boundary:
```go
if resolved != absRoot && !strings.HasPrefix(resolved, absRoot+string(os.PathSeparator)) {
    // resolved escapes project root — skip
}
```

The existing `TestDigest_SymlinkOutsideRootSkipped` uses two independent `t.TempDir()` calls, which on Linux produce sibling paths like `/tmp/TestXxx.../001` and `/tmp/TestXxx.../002` — those don't share a prefix, so the test passes despite the bug. On a project named e.g. `proj` next to a directory named `proj-secret`, a symlink pointing into `proj-secret` would bypass the check.

internal/app/bundle/input_digest_test.go
```diff
@@ -597,6 +575,325 @@ func TestCheckImageHealth_NoWalkerNoVerdict(t *testing.T) {
...
+	walker := bundleapp.NewFileSystemStalenessWalker(root)
+	verdict := runHealthCheck(t, root, inspector, walker)
+
+	// Digest must match (symlink was skipped, not included in the watched set).
+	// Because we have a prior digest (from the svc.WriteInputDigest call above),
+	// Mode must be digest. The new walk result must equal the prior digest.
+	if verdict.Stale {
+		t.Errorf("symlink outside root should not cause STALE, got Stale=true")
+	}
+	if verdict.Mode != bundleapp.FreshnessModeDigest {
```
tibtof, line 892: **Test gap: `TestDigest_SymlinkOutsideRootSkipped` does not cover the HasPrefix boundary bug**

The test creates `root` and `outside` via two separate `t.TempDir()` calls. On Linux these produce sibling directories under the same parent (e.g. `/tmp/TestXxx.../001` and `/tmp/TestXxx.../002`). Their paths do not share a prefix, so the existing `strings.HasPrefix(resolved, absRoot)` check works correctly in the test even though it is wrong in general.

Add a case where the outside directory's name begins with the project root's name — e.g., create a sub-temp dir manually using `filepath.Join(t.TempDir(), "proj")` as root and `filepath.Join(parentTmp, "proj-secret")` as the outside directory — to make the test fail on the buggy implementation and pass only after the fix described in the companion comment on `input_digest.go` line 227.

tibtof:

Two issues must be fixed before this merges. Everything else is solid.

---

## Must fix (2)

### 1. Symlink escape check: HasPrefix without separator anchor — security bug

`internal/app/bundle/input_digest.go`, line 227.

```go
if !strings.HasPrefix(resolved, absRoot) {
```

If `absRoot = "/home/user/proj"` and a symlink resolves to `/home/user/proj-secret/credentials`, the check admits the file because `proj-secret` starts with `proj`. The fix is to anchor the prefix at a path separator:

```go
if resolved != absRoot && !strings.HasPrefix(resolved, absRoot+string(os.PathSeparator)) {
    // skip: escapes project root
}
```

### 2. Symlink test does not cover the prefix-boundary case

`internal/app/bundle/input_digest_test.go`, `TestDigest_SymlinkOutsideRootSkipped`.

The test creates root and outside via two independent `t.TempDir()` calls. On all platforms these get sibling paths that share no prefix (e.g. `/tmp/TestXxx/001` vs `/tmp/TestXxx/002`), so the buggy `HasPrefix` still passes the test. Add a sub-case where the outside path name is a prefix-extension of the root path name (e.g. `/tmp/parent/proj` vs `/tmp/parent/proj-secret`) so that the test fails on the broken implementation and passes only after the fix.

---

## Non-blocking observations (no action required)

**Stale algorithm comment in `computeInputDigest`** (context lines, cannot attach inline): the godoc says the rolled-up hash feeds raw file content into `h.Write`. Since v2, `buildDigest` hashes the per-file SHA-256 hex strings, not raw content. The comment describes the v1 algorithm. Fixing it would be good hygiene — clean up whenever the file is next touched.

**Dead variable in `TestDigest_SymlinkOutsideRootSkipped`**: `digestBefore` is assigned then blanked with `_ = digestBefore`. Remove or use it.

**`stalenessWalker` field on `Service` is now write-only** (set by `WithStalenessWalker`, never read post-#1223). This is intentional per the deprecation comment — fine to leave until a follow-up cleanup PR.

---

## What was verified

- `ensureGitIgnored`, `readGitIgnoreLines`, `inputDigestGitIgnoreLine` fully removed — root cause eliminated.
- `ensureVibewardenGitignoreFile`: read-first idempotent write, correct error handling, correct path.
- Schema v2 bump + v1 → baseline migration path: `readInputDigest` rejects mismatched `schema_version` with a debug log and returns `(InputDigest{}, false)` — caller gets baseline FRESH, no STALE, no panic. Correct.
- `diffDigests` ordering (removed → added → modified, alphabetical within buckets), cap-at-5, `ChangedCount` preserved separately. All correct.
- `renderFreshnessLine` path-list guard (`Stale && !AllowStale && Mode==Digest && len(ChangedPaths)>0`): correct — AllowStale suppresses the list.
- `FreshnessModeBaseline` label: `FRESH (no prior baseline; digest written for next run)`. Correct.
- `.vibewarden/.gitignore` excluded from digest walk via `hardIgnoreDirs['.vibewarden']` before any descent. Correct.
- `bin` and `.next` added to `hardIgnoreDirs` with matching test cases. Correct.
- All 6 required tests present and structurally sound.
- Branch: `fix/1223-bundle-stale-fp`. Commit message: conventional. No force push.

tibtof:
**Documentation review — APPROVE (fixes applied in commit 16ac80c)**

Three drift issues found and fixed directly on this branch:

**1. ADR-089 refinement section contradicts the #1223 implementation (three points)**

The 2026-04-23 (#1146) refinement text in `decisions/adr-089-bundle-image-health-tag-scoping-freshness-arch.md` said:
- JSON example shows `schema_version: 1` — implementation writes `2`
- 'bundle writer MUST append to the project's .gitignore' — the fix writes `.vibewarden/.gitignore` instead of mutating the user's `.gitignore`
- 'schema_version != 1 → fall back to mtime' — FreshnessModeTime was removed end-to-end; v1 is treated as missing → FRESH baseline

Fixed by appending a 2026-04-28 (#1223) refinement note that corrects all three, plus documents hardIgnoreDirs additions (bin, .next) and changed-path rendering.

**2. troubleshooting.md: no section for 'vibew bundle reports STALE'**

Added a complete section: symptom block with path-list format, cause, --build and --allow-stale fixes, what triggers STALE (file list including bin/.next additions), first-upgrade baseline behaviour, and .vibewarden/.gitignore generated file note.

**3. llms-full.txt: image health block shows FRESH only**

Extended with STALE format example showing changed-path lines, v2 digest schema reference, and upgrade baseline note.

**Surfaces with no issues:**
- CHANGELOG entry: correct placement, specific root cause, v1→v2 migration path documented
- `internal/cli/templates/agents/agents-vibewarden.md.tmpl`: no freshness or gitignore mutation language; --allow-stale / --build flags correct
- `docs/examples/AGENTS-VIBEWARDEN.md`: no freshness references; no change needed
- README: vibew bundle is reference-only; no detailed freshness section to update

tibtof:
@reviewer addressed: symlink prefix-extension escape fixed in input_digest.go (added explicit `resolved != absRoot && !strings.HasPrefix(resolved, absRoot + os.PathSeparator)` check); new test case covers `/proj` vs `/proj-secret`. v1 godoc updated to v2. dead test var removed. Ready for re-review.

PR #1235, opened 2026-04-30
fix(#1217): bundle deploy commands use dotfile-safe tar pipe instead of buggy scp glob

### LINKED ISSUE #1217: fix(retro-0.18.3): bundle stdout/README/prompts switch from buggy scp glob to dotfile-safe tar pipe
## Smoke catch
v0.18.2 release smoke caught: when running the documented deploy contract from `bundle/README.md`/`prompt-template --deploy`/`vibew bundle` stdout, the scp wildcard does not always expand:
```
$ scp -r /tmp/vw-smoke-0.18.2/.vibewarden/bundle/* root@host:/opt/vw-smoke/
scp: stat local '/tmp/vw-smoke-0.18.2/.vibewarden/bundle/*': No such file or directory
```
The workaround `tar -czf - -C bundle . | ssh host 'tar -xzf - -C /opt/<app>/'` works but is not what we tell agents to run.

## Cause
Globs depend on the shell's expansion behavior + whether dotfiles are included. The bundle directory contains `.env`, `.credentials`, `.env.template` which a non-dotglob shell will silently skip. Different invocation contexts (Bash tool inside agents, zsh with NULL_GLOB, etc.) yield different behavior.

## Fix options
- (a) Document the glob form AND a no-glob form: `scp -r .vibewarden/bundle root@host:/opt/<app>-bundle && ssh root@host 'mv /opt/<app>-bundle/* /opt/<app>-bundle/.* /opt/<app>/ 2>/dev/null; rmdir /opt/<app>-bundle'`. Ugly.
- (b) Switch to `rsync -av --delete .vibewarden/bundle/ user@host:/opt/<app>/`. Cleaner, includes dotfiles, requires rsync on both ends.
- (c) Tar-pipe: `tar -C .vibewarden/bundle -czf - . | ssh user@host 'tar -C /opt/<app> -xzf -'`. POSIX, no rsync needed, includes dotfiles.

My lean: **(c)** for the canonical bundle README + prompt-template + bundle stdout. POSIX, robust to shell behavior, includes dotfiles deterministically.

## Severity
**MEDIUM** — does not block v0.18.2 release (the deploy contract works with workaround). But if we ship and an agent runs the literal documented commands, they'll hit the same friction the v0.18.1 retro flagged.

## Acceptance
- [ ] Bundle README, `prompts/deploy.tmpl`, `vibew bundle` stdout, `docs/agent-kickoff.md` and `llms-full.txt` Step 7 all use the same robust transfer mechanism.
- [ ] Forensic test (#1204's `TestBundle_Readme_AlignsWithDeployTmpl`) updated to match.

tibtof:
Closes #1217

## Problem

`scp -r .vibewarden/bundle/*` silently drops dotfiles (`.env`, `.credentials`, `.env.template`) because POSIX shells exclude dotfiles from `*` glob expansion unless `dotglob` is set. Three retros called this out: v0.18.1, v0.18.2, v0.18.3. The v0.18.3 deploy shipped a "successful" stack that was missing `.env` and required manual recovery.

## Before → After (the transfer line only)

**Before (broken):**
```
ssh user@host 'mkdir -p /opt/<app>'
scp -r .vibewarden/bundle/* user@host:/opt/<app>/
ssh user@host "cd /opt/<app> && docker load -i image.tar && docker compose up -d"
curl -fsSL https://<domain>/_vibewarden/health
```

**After:**
```
ssh user@host 'mkdir -p /opt/<app>'
tar -czf - -C .vibewarden/bundle . | ssh user@host 'tar -xzf - -C /opt/<app>/'
ssh user@host "cd /opt/<app> && docker load -i image.tar && docker compose up -d"
curl -fsSL https://<domain>/_vibewarden/health
```

Why tar pipe:
- `.` operand (not `*`) includes dotfiles unconditionally — no glob expansion.
- Single ssh connection (faster than `scp -r`).
- `-z` gzip compression.
- POSIX baseline: `tar` + `ssh` already required per `docs/guide/bundle-to-vps.md`.
- `absOut` is quoted with `%q` in `bundle.go` so paths with spaces survive.

## Surfaces changed

All four literal-command surfaces flip to tar pipe. Prose-only `scp` mentions (docs, agent templates) are left alone per architect's verified list.

| Surface | File |
|---------|------|
| `vibew bundle` stdout | `internal/cli/cmd/bundle.go:346` |
| Bundle README.md | `internal/app/bundle/bundle_extras.go:335` |
| `vibew prompt-template --deploy` | `internal/cli/templates/prompts/deploy.tmpl:66` |
| llms-full.txt §Agent Kickoff Prompt | `llms-full.txt:1211` |
| Golden file | `internal/app/promptkickoff/testdata/deploy.golden:66` |

## Forensic regression guard

`TestBundle_Readme_AlignsWithDeployTmpl` extended:
- `extractDeployCommands` regex updated to capture `tar -czf` lines (previously only matched `ssh`/`scp`/`docker`/`curl`).
- Positive check: `tar -czf - -C` and `tar -xzf - -C` required in bundle README, deploy.tmpl, and llms-full.txt.
- Negative check: `scp -r .vibewarden/bundle/*`, `bash deploy.sh`, `./deploy.sh` forbidden across all surfaces.

If any surface drifts back to the glob form, the test fails fast with the surface name and offending substring.

## Test plan

- `make check` passes (golangci-lint + race tests).
- `vibew bundle --skip-image` stdout shows `tar -czf - -C` and does NOT show `scp -r`.
- `.vibewarden/bundle/README.md` fenced deploy block shows tar pipe.
- `vibew prompt-template --deploy --name foo --describe "x" --domain demo.example.com` Step 7 shows tar pipe.


internal/app/bundle/bundle_extras_test.go
```diff
@@ -919,6 +921,44 @@ func TestBundle_Readme_AlignsWithDeployTmpl(t *testing.T) {
...
+	// Forensic alignment: load all four surfaces and assert required/forbidden
+	// patterns across each one (#1217). Failures here mean a surface drifted
+	// back to the dotfile-eating scp glob or a banned artifact command.
+	llmsFullPath := filepath.Join(repoRoot, "llms-full.txt")
+	llmsFullBytes, err := os.ReadFile(llmsFullPath) //nolint:gosec // path derived from source tree
+	if err != nil {
+		t.Fatalf("reading llms-full.txt: %v", err)
+	}
+
+	surfaces := map[string]string{
```
tibtof, line 937: **Missing surface in forensic test**: The comment on this line says "load all four surfaces" but the `surfaces` map only has three entries: `bundle README`, `deploy.tmpl`, and `llms-full.txt`. The fourth surface — bundle stdout (the `runBundle` print block in `internal/cli/cmd/bundle.go`) — is absent from this guard.

Why it matters: the stdout block is the most user-visible output and is generated by different code (`fmt.Fprintf` in `bundle.go`) than the other three. If `bundle.go`'s tar line is ever reverted or mutated, the forensic test will not catch it — only `TestBundle_Stdout_PrintsLiteralDeployCommands` does, which does not run forbidden-pattern checks.

How to fix: capture the bundle stdout as a string inside this test (the test already calls `svc.Bundle()` for the README — wire a fake `io.Writer` to `runBundle` or call the cobra command with a buffer, similar to what `TestBundle_Stdout_PrintsLiteralDeployCommands` does) and add it as a fourth entry:
```go
"bundle stdout": stdoutBuf.String(),
```
Or alternatively, add the forbidden-pattern assertions directly to `TestBundle_Stdout_PrintsLiteralDeployCommands` as a `wantAbsent` table column (simpler, and that test already exercises the stdout path end-to-end).

internal/cli/cmd/bundle_test.go
```diff
@@ -503,8 +503,8 @@ func TestBundle_Stdout_PrintsLiteralDeployCommands(t *testing.T) {
 			wantCmds: []string{
 				"Next: deploy",
 				"ssh user@host 'mkdir -p /opt/myapp'",
-				"scp -r",
-				"user@host:/opt/myapp/",
+				"tar -czf - -C",
```
tibtof, line 507: **Missing test for `%q` path-with-spaces behavior** (re `TestBundle_Stdout_PrintsLiteralDeployCommands`): The architect's design explicitly required: *"Spaces in paths: `%q` should produce `tar -czf - -C "/path with spaces" .` — verify the test covers this."* No such test case exists in `TestBundle_Stdout_PrintsLiteralDeployCommands` or anywhere else in the diff.

Why it matters: `%q` is a deliberate choice to handle the space case — if a future maintainer changes the format verb back to `%s`, paths with spaces will silently break the tar invocation. Without a test exercising a path that contains a space, there is no guard.

How to fix: add a third table entry to `TestBundle_Stdout_PrintsLiteralDeployCommands` using an output dir whose absolute path contains a space, then assert stdout contains the opening double-quote:
```go
{
    name: "output path with spaces is double-quoted",
    yaml:  "name: myapp\ntls:\n  domain: myapp.example.com\nserver:\n  port: 8080\nupstream:\n  port: 3000\n",
    wantCmds: []string{
        // %q wraps the path in double-quotes; assert the quote is present.
        "tar -czf - -C \"",
    },
    spacedOutDir: true, // use filepath.Join(base, "my project", "out")
},
```
(The test runner would need to construct the project inside a path with a space, e.g. `filepath.Join(base, "my project")`.)

tibtof:

Two required items from the architect's design are incomplete. Everything else in the diff is correct — tar form, `%q` quoting, regex ordering, golden file, CHANGELOG entry, all four code-emitted surfaces updated, forbidden-pattern assertions, and negative checks for `bash deploy.sh` / `./deploy.sh`.

---

### Finding 1 — Forensic test covers three surfaces, not four (blocking)

`internal/app/bundle/bundle_extras_test.go` line 934: the comment says "load all four surfaces" but the `surfaces` map has only three entries: `bundle README`, `deploy.tmpl`, `llms-full.txt`. Bundle stdout (`runBundle`'s `fmt.Fprintf` output in `bundle.go`) is the fourth surface specified by the architect and is absent from the required/forbidden guard.

The stdout surface is guarded separately by `TestBundle_Stdout_PrintsLiteralDeployCommands`, but that test has no `wantAbsent` for `scp -r .vibewarden/bundle/*`, `bash deploy.sh`, or `./deploy.sh` — so the forensic contract is incomplete. Fix by either adding the stdout string to the `surfaces` map or adding the forbidden-pattern assertions to the `wantAbsent` column in the stdout test.

### Finding 2 — No test for `%q` path-with-spaces quoting (blocking)

`internal/cli/cmd/bundle_test.go`: the architect's design explicitly required: *"Spaces in paths: `%q` should produce `tar -czf - -C "/path with spaces" .` — verify the test covers this."* No test case exercises a path containing a space or asserts that the output path is wrapped in double-quotes. The `%q` verb is the only guard against space-induced shell word-splitting; without a test, a future `%s` regression goes undetected. Add a third table entry in `TestBundle_Stdout_PrintsLiteralDeployCommands` using an output dir whose absolute path contains a space and assert `"tar -czf - -C \""` is present in stdout.

### Finding 3 — Regex ordering is correct (informational, no action needed)

The updated `extractDeployCommands` regex in `bundle_extras_test.go` places `tar -czf ` before `ssh ` in the alternation. Verified by local execution: the tar-pipe line `tar -czf - -C .vibewarden/bundle . | ssh user@host 'tar -xzf - -C /opt/myapp/'` is captured as one unit (four total matches, no splits, no duplicates). This is correct.

---

Two inline comments posted on the diff. Please fix Findings 1 and 2, then re-request review.

tibtof (commented):
Documentation review: docs consistent with code changes after writer agent fixes.

Seven prose-only surfaces still referenced `scp` as the bundle transfer mechanism after the code surfaces switched to tar pipe. An agent reading any of those surfaces would have reverted to the dotfile-eating glob form. Fixed in commit 4ad5aff on this branch:

- `docs/guide/bundle-to-vps.md` — step 2 now shows the tar pipe command with dotfile rationale; prerequisites updated from `scp` to `tar`+`ssh`; intro chain description updated.
- `docs/deploy-reference.md` — opening sentence updated.
- `docs/agent-kickoff.md` — deploy flavor description updated.
- `llms.txt` — `vibew bundle` command description and `--deploy` flavor description both updated.
- `llms-full.txt` — README output-layout description (line 1452), agent-chain sentence (line 1537), and Step 7 prose summary (line 1915) all updated.
- `internal/cli/templates/agents/agents-vibewarden.md.tmpl` — step 4, CLI reference table row, agent-chain sentence all updated.
- `docs/examples/AGENTS-VIBEWARDEN.md` — step 4 and agent-chain sentence updated.

CHANGELOG entry: names all four code surfaces, cites three retros, states dotfile-safety rationale. Passes.

make check passes. Forensic alignment test passes. Verdict: APPROVE.

tibtof:
@reviewer addressed: forbidden patterns added to wantAbsent in all cases of TestBundle_Stdout_PrintsLiteralDeployCommands (approach b — stdout is in cli/cmd layer, not the domain-layer bundle_extras_test.go; forensic comment updated from 'four surfaces' to 'three static surfaces' with cross-reference to the stdout guard). Path-with-spaces case added as third table entry using filepath.Join(base, "my project", "out"); asserts stdout contains `tar -czf - -C "` (the %q-produced opening double-quote). Writer's doc commits (4ad5aff) unchanged. make check clean. Ready for re-review.

Issue #1224, opened 2026-04-29
docs(retro-0.18.2): macOS system curl (LibreSSL) fails handshake on dev cert — add troubleshooting note


tibtof:
## Retro evidence

From the v0.18.2 retrospective (qr-code-blackhole Node/Express deploy):

> Used Python for HTTPS testing because system curl on macOS uses LibreSSL which couldn't handshake the self-signed dev cert. Workaround: `python3 -c "import ssl, urllib...."`. A note in docs ("on macOS, prefer `curl --insecure` from Homebrew curl, or `python3 -m ...`") would save users this.

> `LibreSSL/3.3.6: error:06FFF064: ... bad decrypt` from system curl

## Diagnosis

macOS ships system curl linked against LibreSSL, which is stricter than OpenSSL on certain self-signed cert profiles (Caddy local CA's ECC intermediate is the likely culprit). Not vibew's bug; document the workaround.

## Severity

**LOW.** One paragraph of docs, one line of `vibew doctor` output on darwin.

## Acceptance criteria

- [ ] Add a section to `docs/troubleshooting.md` titled "macOS: system curl fails handshake on dev cert" with both alternatives:
  - **Homebrew curl:** `brew install curl` → `/opt/homebrew/opt/curl/bin/curl --insecure https://localhost:8443/`
  - **Python:** one-liner using `ssl._create_unverified_context()` + `urllib.request`.
- [ ] `vibew doctor` on darwin: append a single advisory line (NOT a WARN) at the bottom of stdout when running on macOS, e.g.:
  ```
  Note (macOS): system curl uses LibreSSL which may fail handshake on the local dev cert.
                Use Homebrew curl with --insecure or python3 ssl module for HTTPS testing.
                See: docs/troubleshooting.md
  ```
- [ ] Do not surface this on Linux/Windows — darwin-only message.
- [ ] No ADR (pure docs + cosmetic stdout line).

## Out of scope

- **Fixing the cert chain so LibreSSL accepts it.** Worth investigating later — possibly a Caddy upstream issue or a local CA extension flag — but out of scope for this retro pipeline. File as a separate follow-up if you want to dig in.
- Recommending users disable macOS system curl. Not our call.

## Source

Retro: `~/notes/vibewarden/retro-0.18.2.md` (2026-04-28)

tibtof:
## Architect handoff — READY_FOR_DEV

No ADR (per spec: pure docs + cosmetic stdout line on darwin only).

### Files to change

[...]

PR #1239, opened 2026-04-30
feat(#1233): vibew probe [--env name] — Go-stdlib HTTPS probe of /_vibewarden/health

### LINKED ISSUE #1233: feat(retro-0.18.3): vibew probe [--env <name>] — Go-stdlib HTTPS probe of /_vibewarden/health
## Retro evidence

From the v0.18.3 retrospective (qr-code-brutalist Node deploy):

> Used Python for HTTPS testing because system curl on macOS uses LibreSSL which couldn't handshake the self-signed dev cert. Workaround: `python3 -c "import ssl, urllib.request..."`. A `vibew curl <path>` or `vibew probe` helper that uses Go's TLS stack would erase a real friction point.

#1224 added a `vibew doctor` advisory pointing at the workaround, but the friction remains: agents still have to context-switch to a Python one-liner to verify the dev stack.

## Fix

New CLI: `vibew probe [--env <name>]`.

- **Default (no flag):** GETs `https://localhost:<server.port>/_vibewarden/health` using Go's stdlib HTTP client (with `InsecureSkipVerify` for the dev cert). Bypasses LibreSSL by construction. Prints the parsed JSON + a one-line summary.
- **`--env <name>`:** Reads `vibewarden.<name>.yaml` (e.g. `--env production` reads `vibewarden.production.yaml`), resolves the production domain via merged config, probes that. Same Go stdlib path. Replaces the curl line in bundle stdout for users who want a working probe everywhere.

## Acceptance criteria

- [ ] `vibew probe` (no args) exits 0 when the local stack returns 200 with `components.upstream:"ok"`. Exits 1 otherwise. Stack-not-running detected the same way `vibew logs` does (post-#1221).
- [ ] `vibew probe --env <name>` reads `vibewarden.<name>.yaml`, resolves `tls.domain`, probes `https://<domain>/_vibewarden/health`. Real cert verification (no `InsecureSkipVerify`) for non-dev envs.
- [ ] **Generalizable `--env` mechanism.** The flag wiring should be reusable for future verbs (`vibew status --env prod`, `vibew validate --env prod`, etc). Architect designs the env-resolver shared helper.
- [ ] Output format:
  ```
  $ vibew probe
  https://localhost:8443/_vibewarden/health
    status:              ok
    version:             0.18.4
    components.sidecar:  ok
    components.upstream: ok

  OK — dev stack healthy.
  ```
- [ ] On `components.upstream:"unknown"` (boot-gap window), wait + retry up to 10s before reporting `degraded`. Match #1197's wire-format design.

## Out of scope

- Generic curl wrapper (`vibew curl <path>`) — explicitly rejected at triage (curl re-implementation, scope creep).
- `--no-tls` dev-mode bypass — explicitly rejected (weakens dev/prod parity).
- Fixing the LibreSSL cert-chain compatibility upstream — separate follow-up.

## Severity

**MEDIUM.** Workaround documented in #1224 but adds friction every dev session on macOS.

## Source

Retro: `~/notes/vibewarden/retro-0.18.3.md` (2026-04-29)

tibtof:
Closes #1233

## Summary

- Adds `vibew probe [--env <name>]` CLI verb (ADR-102)
- New `HealthProber` port + `HealthDocument` value object + sentinel errors in `internal/ports/health_prober.go`
- `HTTPProber` adapter in `internal/adapters/health/prober.go` — `NewLocalhostProber` (InsecureSkipVerify) and `NewStrictProber` (full chain), both pure stdlib
- Env-resolver package `internal/app/env/` — canonical `Resolver` / `FileResolver` / `Resolved` for all future `--env`-flag verbs; delegates deep-merge to existing `bundleapp.LoadMergedConfig`
- Probe service `internal/app/probe/` — boot-gap retry (10s / 1s) at app layer, injectable sleep for test isolation
- Render package with six golden-pinned output branches (dev-ok, env-ok, boot-gap-exhausted, refused, non-200, malformed)
- Root wired: `NewProbeCmd()` added to `cmd/root.go`

## Resolves the macOS LibreSSL friction (#1224 complement)

The #1224 advisory documented a workaround (`python3 -c "import ssl..."`). `vibew probe` eliminates the need for it: Go's TLS stack handshakes the dev self-signed cert without complaint.

## Dev-ok output (pinned by golden test)

```
https://localhost:8443/_vibewarden/health
  status:               ok
  version:              0.18.4
  components.sidecar:   ok
  components.upstream:  ok

OK — dev stack healthy.
```

## Env-ok output (pinned by golden test)

```
https://app.example.com/_vibewarden/health
  status:               ok
  version:              0.18.4
  components.sidecar:   ok
  components.upstream:  ok

OK — production stack healthy.
```

## Test plan

- [ ] `go test -race ./internal/app/probe/... ./internal/adapters/health/... ./internal/app/env/... ./internal/cli/cmd/...` — all pass
- [ ] `golangci-lint run ./...` — 0 issues
- [ ] `gofmt -l .` — empty
- [ ] All 6 golden branches verified (dev-ok, env-ok, boot-gap-exhausted, refused, non-200, malformed)
- [ ] `make check` passes (pre-push hook confirmed on push)

## ADR

ADR-102: `decisions/adr-102-vibew-probe-go-stdlib-health-probe-and-env-resolver.md`

### DIFF
--- a/internal/adapters/health/prober_test.go
+++ b/internal/adapters/health/prober_test.go
@@ -2,6 +2,7 @@ package health
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -185,6 +186,41 @@ func TestNewLocalhostProber_InsecureSkipVerify(t *testing.T) {
 	}
 	if doc.Status != "ok" {
 		t.Errorf("status = %q, want %q", doc.Status, "ok")
+	}
+}
+
+func TestNewStrictProber_RejectsSelfSignedCert(t *testing.T) {
+	// httptest.NewTLSServer presents a self-signed certificate. NewStrictProber
+	// uses the stdlib default TLS verification (no InsecureSkipVerify), so the
+	// handshake must fail with an x509 unknown-authority error. This is the
+	// symmetric counterpart to TestNewLocalhostProber_InsecureSkipVerify and
+	// validates the --env safety property: a production probe against an invalid
+	// cert is rejected, not silently accepted.
+	srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write(healthBody("ok", "0.18.4", map[string]string{
+			"sidecar":  "ok",
+			"upstream": "ok",
+		}))
+	}))
+	defer srv.Close()
+
+	prober := NewStrictProber(3 * time.Second)
+	_, err := prober.Probe(context.Background(), srv.URL)
+	if err == nil {
+		t.Fatal("NewStrictProber should reject a self-signed certificate, but got nil error")
+	}
+
+	// The error must be (or wrap) an x509.UnknownAuthorityError, proving that
+	// strict TLS verification is active. Accept either errors.As match or a
+	// substring check to remain robust across Go stdlib versions.
+	var unknownAuth x509.UnknownAuthorityError
+	msg := err.Error()
+	if !errors.As(err, &unknownAuth) &&
+		!strings.Contains(msg, "certificate signed by unknown authority") &&
+		!strings.Contains(msg, "x509") {
+		t.Errorf("expected x509 TLS rejection error, got: %v", err)
 	}
 }
 
--- a/internal/adapters/health/prober_test.go
+++ b/internal/adapters/health/prober_test.go
@@ -1,0 +1,237 @@
+package health
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vibewarden/vibewarden/internal/ports"
+)
+
+// healthBody is a helper to produce a valid /_vibewarden/health JSON response.
+func healthBody(status, version string, components map[string]string) []byte {
+	b, _ := json.Marshal(map[string]any{
+		"status":     status,
+		"version":    version,
+		"components": components,
+	})
+	return b
+}
+
+func TestHTTPProber_Probe_200_ValidBody(t *testing.T) {
+	srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write(healthBody("ok", "0.18.4", map[string]string{
+			"sidecar":  "ok",
+			"upstream": "ok",
+		}))
+	}))
+	defer srv.Close()
+
+	// Use the test server's own client (trusts its cert).
+	prober := &HTTPProber{client: srv.Client()}
+
+	doc, err := prober.Probe(context.Background(), srv.URL)
+	if err != nil {
+		t.Fatalf("expected no error, got: %v", err)
+	}
+	if doc.Status != "ok" {
+		t.Errorf("status = %q, want %q", doc.Status, "ok")
+	}
+	if doc.Version != "0.18.4" {
+		t.Errorf("version = %q, want %q", doc.Version, "0.18.4")
+	}
+	if doc.Components["upstream"] != "ok" {
+		t.Errorf("components.upstream = %q, want %q", doc.Components["upstream"], "ok")
+	}
+}
+
+func TestHTTPProber_Probe_200_MalformedBody(t *testing.T) {
+	tests := []struct {
+		name string
+		body string
+	}{
+		{"not json", "this is not json"},
+		{"missing status", `{"version":"0.18.4","components":{"sidecar":"ok"}}`},
+		{"missing components", `{"status":"ok","version":"0.18.4"}`},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+				w.WriteHeader(http.StatusOK)
+				_, _ = w.Write([]byte(tt.body))
+			}))
+			defer srv.Close()
+
+			prober := &HTTPProber{client: srv.Client()}
+			_, err := prober.Probe(context.Background(), srv.URL)
+			if err == nil {
+				t.Fatal("expected error, got nil")
+			}
+			if !errors.Is(err, ports.ErrProbeMalformed) {
+				t.Errorf("expected ErrProbeMalformed in chain, got: %v", err)
+			}
+		})
+	}
+}
+
+func TestHTTPProber_Probe_Non200(t *testing.T) {
+	tests := []struct {
+		name       string
+		statusCode int
+		body       string
+	}{
+		{"502 Bad Gateway", http.StatusBadGateway, "Bad Gateway"},
+		{"503 Service Unavailable", http.StatusServiceUnavailable, "Service Unavailable"},
+		{"404 Not Found", http.StatusNotFound, "not found"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+				w.WriteHeader(tt.statusCode)
+				_, _ = w.Write([]byte(tt.body))
+			}))
+			defer srv.Close()
+
+			prober := &HTTPProber{client: srv.Client()}
+			_, err := prober.Probe(context.Background(), srv.URL)
+			if err == nil {
+				t.Fatal("expected error, got nil")
+			}
+			if !errors.Is(err, ports.ErrProbeNon200) {
+				t.Errorf("expected ErrProbeNon200 in chain, got: %v", err)
+			}
+			var ne *ports.ProbeNon200Error
+			if !errors.As(err, &ne) {
+				t.Fatalf("expected *ProbeNon200Error, got: %T", err)
+			}
+			if ne.StatusCode != tt.statusCode {
+				t.Errorf("StatusCode = %d, want %d", ne.StatusCode, tt.statusCode)
+			}
+			if !strings.Contains(ne.Body, tt.body) {
+				t.Errorf("body snippet %q does not contain %q", ne.Body, tt.body)
+			}
+		})
+	}
+}
+
+func TestHTTPProber_Probe_ConnectionRefused(t *testing.T) {
+	// Find a port that is definitely not listening.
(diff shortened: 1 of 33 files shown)


internal/cli/cmd/probe.go
```diff
@@ -0,0 +1,175 @@
...
+	svc := probeapp.NewService(prober)
+
+	opts := probeapp.DefaultOptions(targetURL, envName)
+	result, probeErr := svc.Run(cmd.Context(), opts)
+	probeapp.Render(w, result, probeErr)
+
+	return exitCode(result, probeErr)
+}
+
+// exitCode maps a probe result + error to a cobra-compatible return value.
```
tibtof, line 142: **Bug: silentError does not suppress cobra's error output**

`silentError.Error()` returns `""` but `SilenceErrors` is never set to `true` on `NewProbeCmd()`. Cobra's `Execute()` calls `cmd.PrintErr("Error: ", err.Error())` for any non-nil error from `RunE`, producing a bare `"Error: "` line (with empty message) to stderr on top of the already-rendered output. This redundant error line is exactly what the silentError pattern is supposed to prevent.

Fix: add `SilenceErrors: true` to the command struct:

```go
cmd := &cobra.Command{
    Use:           "probe [--env <name>]",
    SilenceErrors: true,
    // ...
}
```

This is the standard pattern for commands that own their own error rendering. Without it every exit-1 path emits `"Error: "` to stderr in addition to the rendered output block.

internal/cli/cmd/probe.go
```diff
@@ -0,0 +1,175 @@
...
+//
+// Exit codes:
+//   - 0: components.upstream == "ok" (stack is fully healthy)
+//   - 1: any other state (refused, malformed, non-200, boot-gap exhausted, failing)
+func NewProbeCmd() *cobra.Command {
+	var envName string
+
+	cmd := &cobra.Command{
+		Use:   "probe [--env <name>]",
+		Short: "Probe the sidecar's /_vibewarden/health endpoint",
```
tibtof: **Missing: `cobra.NoArgs` not set**

The review spec requires `cobra.NoArgs` on `vibew probe` (it takes no positional args). Without it, `vibew probe some-accidental-arg` silently succeeds (or produces a confusing error) rather than giving a clean `"unknown command"`-style rejection.

Fix: add `Args: cobra.NoArgs` to the command struct alongside `Use` and `Short`:

```go
cmd := &cobra.Command{
    Use:   "probe [--env <name>]",
    Args:  cobra.NoArgs,
    Short: "Probe the sidecar's /_vibewarden/health endpoint",
    // ...
}
```

internal/adapters/health/prober_test.go
```diff
@@ -0,0 +1,237 @@
...
+	_, err := prober.Probe(ctx, srv.URL)
+	if err == nil {
+		t.Fatal("expected error from cancelled context, got nil")
+	}
+	// The error should not be ErrProbeRefused (it's a cancellation).
+	if errors.Is(err, ports.ErrProbeRefused) {
+		t.Errorf("cancelled context should not produce ErrProbeRefused")
+	}
+	_ = fmt.Sprintf("%v", err) // suppress unused import warning
+}
```
tibtof, line 273: **Missing test: NewStrictProber must reject a self-signed cert**

The review spec explicitly requires: "TLS verification asymmetry test: explicit assertion that the env-target prober rejects an invalid cert (httptest with a bad cert)."

This test is missing from `prober_test.go`. `TestNewLocalhostProber_InsecureSkipVerify` proves the insecure path succeeds, but there is no symmetric test proving `NewStrictProber` *fails* against `httptest.NewTLSServer`'s self-signed cert.

Add a test like:

```go
func TestNewStrictProber_RejectsUntrustedCert(t *testing.T) {
    srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
        w.WriteHeader(http.StatusOK)
    }))
    defer srv.Close()

    // NewStrictProber uses full chain verification — must reject the self-signed cert.
    prober := NewStrictProber(2 * time.Second)
    _, err := prober.Probe(context.Background(), srv.URL)
    if err == nil {
        t.Fatal("expected TLS error against self-signed cert, got nil")
    }
    // Should NOT be ErrProbeRefused — the connection was made, TLS failed.
    if errors.Is(err, ports.ErrProbeRefused) {
        t.Error("TLS failure should not masquerade as connection refused")
    }
}
```

Without this test, the TLS asymmetry that the entire --env safety property rests on is not verified in code.

internal/app/env/resolver_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+// minimalBase returns a minimal vibewarden.yaml content suitable for tests.
+func minimalBase(port int) string {
+	return "server:\n  port: " + itoa(port) + "\n"
+}
+
+// itoa converts an int to a string. We avoid importing strconv to keep helpers minimal.
+func itoa(i int) string {
+	if i == 8443 {
+		return "8443"
```
tibtof, line 28: **Fragile hand-rolled itoa that silently returns wrong values**

The `itoa` helper (lines 1108–1113) only handles two hard-coded values (8443 → "8443", everything else → "9443"). A test using any other port would silently get "9443" back, making the YAML written to the temp dir incorrect and the test unreliable. The comment says "we avoid importing strconv to keep helpers minimal" but this is worse than the problem it avoids.

Replace with `fmt.Sprintf("%d", i)` or `strconv.Itoa(i)`:

```go
func minimalBase(port int) string {
    return "server:\n  port: " + strconv.Itoa(port) + "\n"
}
```

There's no reason to avoid `strconv` here — it's stdlib.

internal/cli/cmd/probe_test.go
```diff
@@ -0,0 +1,221 @@
...
+	root := &cobra.Command{Use: "vibew"}
+	root.AddCommand(cmd.NewProbeCmd())
+	return root
+}
+
+// execProbe runs "vibew probe [args...]" and returns the captured stdout,
+// stderr and the cobra error.
+func execProbe(t *testing.T, args []string) (string, string, error) {
+	t.Helper()
+	root := newProbeRoot()
```
tibtof, line 39: **TestProbeCmd_DefaultPath_OK has zero assertions**

This test (lines 40–42 relative to probe_test.go) explicitly discards all output and errors: `_ = stdout` with a comment saying "connection refused is expected; test verifies no panic". A test that only verifies "no panic" provides near-zero value — panics are caught by the Go test runner regardless. The entire test is equivalent to a smoke test with no smoke.

The review spec requires the CLI test to assert both output content and exit code. At minimum this test should:
1. Assert the rendered output contains the target URL.
2. Assert the returned error is non-nil (expected connection refused).
3. Verify that stderr does NOT contain `"Error: "` (which would confirm the silentError fix above is working).

If the test can't control the network path cleanly, it should be refactored to inject the prober (or use `--env` path with an httptest server as done in `TestProbeCmd_EnvPath_OK`).

tibtof:

Four issues require fixes before merge. None are architecture violations — the hexagonal structure is clean, the boot-gap retry logic is correct, and the TLS asymmetry is properly implemented in production code. The blocking issues are in the test layer and CLI wiring.

---

**Issue 1 (BLOCKING): silentError pattern incomplete — cobra will still emit `Error: ` to stderr**

`silentError.Error()` returns `""` but `SilenceErrors: true` is never set on `NewProbeCmd()`. Cobra's `Execute()` prints `"Error: " + err.Error()` to stderr for any non-nil `RunE` return. Every exit-1 path will emit a bare `"Error: "` line to stderr on top of the already-rendered output block. The entire purpose of the silentError type is defeated. Fix: add `SilenceErrors: true` to the `cobra.Command` struct in `NewProbeCmd()`. Inline comment posted at line 139 of `internal/cli/cmd/probe.go`.

**Issue 2 (BLOCKING): Missing `cobra.NoArgs` — probe silently accepts positional args**

The review spec requires `cobra.NoArgs` on `vibew probe`. Without it `vibew probe some-accidental-arg` is accepted rather than rejected. Add `Args: cobra.NoArgs` to the command struct. Inline comment posted at line 38 of `internal/cli/cmd/probe.go`.

**Issue 3 (BLOCKING): No test asserting `NewStrictProber` rejects a self-signed cert**

The review spec explicitly requires: "explicit assertion that the env-target prober rejects an invalid cert (httptest with a bad cert)." `TestNewLocalhostProber_InsecureSkipVerify` proves the insecure path succeeds but the symmetric test — that `NewStrictProber` fails against the same cert — does not exist. The TLS asymmetry that the entire `--env` safety property rests on is unverified in code. The template for the missing test is in the inline comment at line 237 of `internal/adapters/health/prober_test.go`.

**Issue 4 (BLOCKING): `TestProbeCmd_DefaultPath_OK` has zero assertions**

This test discards all output (`_ = stdout`) and only verifies "no panic". The review spec requires CLI tests to assert both output content and exit code. At minimum: verify the target URL appears in rendered output and the returned error is non-nil (connection refused is expected). Inline comment posted at line 39 of `internal/cli/cmd/probe_test.go`.

---

**Non-blocking findings (fix now to avoid follow-up):**

- `itoa` helper in `internal/app/env/resolver_test.go` is silently broken for any port other than 8443 or 9443. Use `strconv.Itoa` or `fmt.Sprintf`. Inline comment at line 28.
- `PerProbe` field on `Options` is documented as "not currently used" — dead field. Consider removing it or wiring it as a per-probe context deadline before merge to avoid the field becoming permanent dead weight.
- No architecture invariant test asserting `internal/adapters/health/prober.go` does not import `internal/app/probe/`. The review spec calls this out. Low-priority but worth adding.

---

**What is correct and does NOT need changes:**

- Hexagonal structure: ports in `internal/ports/`, adapter in `internal/adapters/health/`, app service in `internal/app/probe/`, env resolver in `internal/app/env/`. Clean.
- Boot-gap retry: the `deadline := time.Now().Add(opts.BootGapWait); for time.Now().Before(deadline)` pattern with sleep-before-probe yields exactly N retries within the budget window. No off-by-one.
- TLS asymmetry in production code: `NewLocalhostProber` sets `InsecureSkipVerify: true`, `NewStrictProber` uses plain `&http.Client{Timeout: timeout}` (stdlib default = full chain verification). No `--insecure` flag. Correct.
- `bundleapp.LoadMergedConfig` delegation in `FileResolver`: no duplicated deep-merge logic.
- `ErrBootGapExhausted` sentinel returned with last `Result` so renderer can write the degraded block before exit.
- All 6 golden branches present and byte-matched.
- No new external dependencies — pure stdlib.
- CHANGELOG and `decisions/README.md` entries are accurate.
- `//nolint:gosec` annotation with clear justification comment on the `InsecureSkipVerify` line.

tibtof:
@reviewer addressed: SilenceErrors+SilenceUsage on probe cmd; NewStrictProber rejection test added against httptest.NewTLSServer; cobra.NoArgs added with rejection test; TestProbeCmd_DefaultPath_OK now asserts rendered output. Writer's doc commits 58b6c70 + f0a1624 unchanged. Ready for re-review.

PR #1250, opened 2026-04-30
feat(#1245): vibew bundle --print-deploy --host --user --path (ad-hoc deploy override)

### LINKED ISSUE #1245: feat(retro-0.18.4 follow-up): vibew bundle --print-deploy --host --user --path (ad-hoc companion to #1244)
## Retro evidence (Codex v0.18.4 follow-up feedback)

> Add a `vibew bundle --print-deploy --host ... --user ... --path ...`. Not a deployment tool, just command rendering: it prints the exact tar/ssh/docker commands with substitutions. That keeps ownership with the operator but reduces agent mistakes.

## Companion to #1244

#1244 adds a persistent `deploy.host` field in `vibewarden.production.yaml` for version-controlled SSH targets. This issue adds an **ad-hoc per-invocation override** for one-off deploys, multi-host CI, or environments where mutating config isn't desired.

Both share the same substitution code path under the hood; the flag overrides the config when both are present.

## Acceptance criteria

- [ ] New flag set on `vibew bundle`: `--print-deploy --host <host> --user <user> --path <remote-path>`. All three flags must be supplied together when `--print-deploy` is set; missing any → exit 1 with clear error.
- [ ] When `--print-deploy` is set: bundle still runs as today, but the "Next: deploy" stdout block uses `<user>@<host>` and `<remote-path>` verbatim (no `<your-ssh-user>@<your-ssh-host>` placeholder, no hint paragraph).
- [ ] When `--print-deploy` is NOT set:
  - If `cfg.Deploy.Host` is set in the merged production config: substitute as #1244 specifies.
  - Else: emit the bracketed placeholder + hint as #1244 specifies.
- [ ] When BOTH are set: flag wins. Document this in `vibew bundle --help`.
- [ ] Test cases:
  - `vibew bundle --print-deploy --host h --user u --path /opt/x` → stdout has `ssh u@h 'mkdir -p /opt/x'` etc.
  - `vibew bundle --print-deploy --host h --user u` (missing --path) → exit 1 with `--print-deploy requires --host, --user, --path`.
  - `vibew bundle --print-deploy --host h --user u --path /opt/x` overrides `cfg.Deploy.Host = root@otherhost` → flag wins, host `h` used.

## Severity

**HIGH.** Closely paired with #1244; ships in the same release.

## Source

Retro: `~/notes/vibewarden/retro-0.18.4.md` (2026-04-30) — Codex follow-up feedback recommendation #4.

tibtof:
Closes #1245

## Summary

- Adds `--print-deploy` (bool), `--host`, `--user`, `--path` (strings) flags to `vibew bundle`
- When `--print-deploy` is set with all three sub-flags, the "Next: deploy" stdout block substitutes `<user>@<host>` and `--path` for the remote directory, overriding `deploy.host` from `vibewarden.production.yaml`
- Bundle README is unaffected (Option A: README is a versioned artifact, flag is stdout-only by design)
- Flag combination validation runs before any config load or bundle work — both invalid forms exit 1 with a clear error

## Before / after

**Ad-hoc invocation (no config, flag wins):**
```
vibew bundle --skip-image --print-deploy --host 1.2.3.4 --user deploy --path /opt/myapp

Next: deploy
    ssh deploy@1.2.3.4 'mkdir -p /opt/myapp'
    tar -czf - -C "/path/to/.vibewarden/bundle" . | ssh deploy@1.2.3.4 'tar -xzf - -C /opt/myapp/'
    ssh deploy@1.2.3.4 "cd /opt/myapp && docker compose up -d"
    curl -fsSL https://myapp.example.com/_vibewarden/health
```

**Config + flag override (flag wins):**
```yaml
# vibewarden.production.yaml
deploy:
  host: root@configured.example
```
```
vibew bundle --skip-image --print-deploy --host 1.2.3.4 --user deploy --path /opt/myapp
# stdout: deploy@1.2.3.4, NOT root@configured.example
```

**README unaffected (Option A):**
- README.md on disk still uses `root@configured.example` (from config) or the bracketed placeholder (when no config). The flag never touches the README.

**Validation error examples:**
```
vibew bundle --print-deploy --host h --user u
# Error: --print-deploy requires --host, --user, --path (missing: --path)

vibew bundle --host h --user u --path /opt/x
# Error: --host/--user/--path require --print-deploy
```

## Test plan

- `TestValidatePrintDeployFlags` (9 table cases, `package cmd` internal test) — unit tests the pure validation function covering all accept/reject combinations
- `TestBundle_PrintDeploy_FlagValidationErrors` (7 cases) — drives invalid combos through cobra and asserts exit 1 with correct error text
- `TestBundle_PrintDeploy_StdoutSubstitution` (3 cases) — all-three-substituted, flag-overrides-config, README-ignores-flag
- `TestBundle_PrintDeploy_PathSubstitutesRemoteDir` — confirms `--path` replaces `/opt/<appname>` in all three SSH lines; healthcheck curl unaffected
- `TestBundle_PrintDeploy_HintParagraphSuppressed` — hint paragraph absent when `--print-deploy` is set
- `TestBundle_Stdout_PrintsLiteralDeployCommands` extended — `"--print-deploy"` added to `wantAbsent` in all four existing cases (sanity: flag name never leaks into output)
- `make check` clean (gofmt, golangci-lint, build, `go test -race ./...`, demo-app) — verified by pre-push hook

### DIFF
--- a/internal/cli/cmd/bundle.go
+++ b/internal/cli/cmd/bundle.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
@@ -46,6 +47,12 @@ const multiSiteErrorMessage = "multi-site bundle is post-v1; see #1169 for the N
 //   - --allow-stale        suppress the stale warning; bundle proceeds regardless
 //   - --target-platform    target deployment platform (default: linux/amd64)
 //
+// New flags (#1245):
+//   - --print-deploy        substitute --host/--user/--path into the stdout "Next: deploy" block
+//   - --host                SSH host for the stdout deploy block (requires --print-deploy)
+//   - --user                SSH user for the stdout deploy block (requires --print-deploy)
+//   - --path                remote deploy path for the stdout deploy block (requires --print-deploy)
+//
 // Exit codes:
 //   - 0: success (including warnings)
 //   - 1: generic / config failure
@@ -64,6 +71,10 @@ func NewBundleCmd() *cobra.Command {
 		build          bool
 		allowStale     bool
 		targetPlatform string
+		printDeploy    bool
+		deployHost     string
+		deployUser     string
+		deployPath     string
 	)
 
 	cmd := &cobra.Command{
@@ -110,9 +121,22 @@ Examples:
   vibew bundle --output build/deploy
   vibew bundle --skip-image
   vibew bundle --image ghcr.io/acme/myapp:v1.2.3
-  vibew bundle --overwrite`,
+  vibew bundle --overwrite
+  vibew bundle --print-deploy --host 1.2.3.4 --user root --path /opt/myapp   # ad-hoc deploy block (no config mutation)
+
+Deploy block precedence ("Next: deploy" stdout):
+  1. --print-deploy --host <h> --user <u> --path <p>  (ad-hoc; affects stdout only)
+  2. deploy.host in vibewarden.production.yaml         (persistent; affects stdout AND bundle README)
+  3. bracketed placeholder + hint paragraph            (default; nothing configured)
+
+The bundle README always reflects #2 or #3 — --print-deploy is stdout-only by design,
+since the README is a versioned artifact that ships with the bundle.
+
+--print-deploy overrides any deploy.host in vibewarden.production.yaml for the
+printed "Next: deploy" block. All three sub-flags (--host, --user, --path) must be
+supplied together. Paths with spaces in --path are not supported.`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			code, err := runBundle(cmd, outputDir, imageTag, targetPlatform, overwrite, skipImage, build, allowStale)
+			code, err := runBundle(cmd, outputDir, imageTag, targetPlatform, overwrite, skipImage, build, allowStale, printDeploy, deployHost, deployUser, deployPath)
 			if err != nil {
 				// Set the process exit code for semantic exit codes (2, 3) while
 				// still surfacing the error message via cobra.
@@ -137,6 +161,11 @@ Examples:
 	cmd.Flags().BoolVar(&allowStale, "allow-stale", false, "suppress the stale-image warning (bundle proceeds regardless of freshness)")
 	cmd.Flags().StringVar(&targetPlatform, "target-platform", "",
 		"expected deployment platform, e.g. linux/arm64 (default: deploy.target_platform from vibewarden.production.yaml, or linux/amd64)")
+	cmd.Flags().BoolVar(&printDeploy, "print-deploy", false,
+		"substitute --host/--user/--path into the printed 'Next: deploy' block (overrides deploy.host from production.yaml)")
+	cmd.Flags().StringVar(&deployHost, "host", "", "SSH host to substitute into the printed deploy block (requires --print-deploy)")
+	cmd.Flags().StringVar(&deployUser, "user", "", "SSH user to substitute into the printed deploy block (requires --print-deploy)")
+	cmd.Flags().StringVar(&deployPath, "path", "", "remote deploy path to substitute into the printed deploy block, e.g. /opt/myapp (requires --print-deploy; paths with spaces are not supported)")
 
 	return cmd
 }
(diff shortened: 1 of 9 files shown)


internal/cli/cmd/bundle.go
```diff
@@ -159,6 +188,12 @@ func runBundle(cmd *cobra.Command, outputDir, imageTag, targetPlatform string, o
 		outputDir = defaultBundleOutputDir
 	}
 
+	// Validate --print-deploy flag combination BEFORE any config load or bundle
+	// work. Both invalid forms exit 1 with a clear error message.
+	if err := validatePrintDeployFlags(printDeploy, deployHost, deployUser, deployPath); err != nil {
```
tibtof: **Spec violation: validation runs AFTER `requireScaffolding()`, not before**

The issue spec is explicit: *"Validation order: validate flag combination FIRST, before any config load, before any bundle work."* The current ordering is:

```
requireScaffolding()          // 1st — bundle precondition check
outputDir default             // 2nd
validatePrintDeployFlags()    // 3rd  ← should be 1st
loadAndResolve()              // 4th
```

`requireScaffolding()` is bundle setup — it checks whether the command can operate at all. The spec says to validate flags before any bundle work. Under the current code, a user running `vibew bundle --host h --user u` in a non-vibewarden directory sees:

> vibewarden not initialized in this directory

instead of:

> --host/--user/--path require --print-deploy

The UX impact is minor but the ordering contradicts the spec. Fix: move `validatePrintDeployFlags` above `requireScaffolding()`:

```go
func runBundle(cmd *cobra.Command, ...) (int, error) {
    // Validate flags first — pure check, no I/O, no side effects.
    if err := validatePrintDeployFlags(printDeploy, deployHost, deployUser, deployPath); err != nil {
        return 1, err
    }
    if err := requireScaffolding(); err != nil {
        return 1, err
    }
    ...
}
```

`validatePrintDeployFlags` is pure (no I/O, reads only its arguments) — there is no reason it cannot run first.

internal/cli/cmd/bundle_test.go
```diff
@@ -645,6 +650,325 @@ func TestBundle_Stdout_PrintsLiteralDeployCommands(t *testing.T) {
...
+		{
+			name:        "--path without --print-deploy",
+			args:        []string{"bundle", "--skip-image", "--path", "/opt/x"},
+			errContains: []string{"require --print-deploy"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Validation runs before scaffolding checks (requireScaffolding comes
```
tibtof: **Misleading comment — contradicts the code it describes**

The comment says `"Validation runs before scaffolding checks"` on the first line, then immediately admits the opposite in the parenthetical: `"(requireScaffolding comes first in runBundle...)"`.

These two clauses directly contradict each other. The code runs scaffolding first, then validation. The test is correct (it uses `setupBundleProject` so scaffolding passes and validation error is observed), but the leading clause will mislead any reader.

Fix: remove the contradictory leading clause and keep the accurate parenthetical:

```go
// Use a temp dir with scaffolding so the test can observe the validation
// error, not the scaffolding error (requireScaffolding runs before
// validatePrintDeployFlags in runBundle).
```

Or, if the ordering is fixed to put validation first (see bundle.go comment), update to:

```go
// Validation runs before scaffolding checks — validatePrintDeployFlags is
// called before requireScaffolding. Use a bare temp dir here to confirm
// the validation error surfaces immediately without scaffolding.
```

tibtof:

Two issues found. No architecture violations, no test coverage gaps, no license issues.

---

## Finding 1 — Spec violation: validation runs after `requireScaffolding()` (MUST FIX)

**File:** `internal/cli/cmd/bundle.go`, line 193

The issue spec says: *"Validation order: validate flag combination FIRST, before any config load, before any bundle work."*

Actual order in `runBundle`:
1. `requireScaffolding()` — checks .vibewarden/ marker exists
2. outputDir default
3. `validatePrintDeployFlags()` ← spec says this must be 1st
4. `loadAndResolve()`

`requireScaffolding()` is bundle precondition work. A user running `vibew bundle --host h --user u` in a non-vibewarden directory gets the scaffolding error instead of the flag validation error. The spec requires the opposite priority.

Fix: move `validatePrintDeployFlags` above `requireScaffolding()`. The function is pure (no I/O, reads only its arguments), so there is no barrier to running it first.

---

## Finding 2 — Misleading test comment (MUST FIX, follows from Finding 1)

**File:** `internal/cli/cmd/bundle_test.go`, line 702

The comment reads: *"Validation runs before scaffolding checks (requireScaffolding comes first in runBundle, but validatePrintDeployFlags is called right after the outputDir default)."*

The leading clause and the parenthetical directly contradict each other. The parenthetical correctly describes the code; the leading clause does not. Any reader — human or agent — will be confused. Fix the comment to match whichever ordering is chosen after Finding 1 is resolved.

---

## What passed

- **Validation purity:** `validatePrintDeployFlags` is fully pure — no I/O, no globals, no side effects. Table-driven, all 9 cases correct.
- **Substitution correctness:** `remotePath` replaces `/opt/<appName>` in all three SSH `Fprintf` lines. `%q` quoting on the LOCAL `absOut` path (post-#1217) is preserved on line 411. The `--path` value itself is correctly not `%q`-quoted, matching the existing unquoted `/opt/<app>` form.
- **README-unaffected guarantee:** The flag values never reach `bundleapp.BundleOptions` or `bundle_extras.go`. `renderBundleReadme` uses `opts.Config.Deploy.Host` exclusively. `TestBundle_PrintDeploy_StdoutSubstitution` / flag-overrides-config case asserts `readmeContains: ["root@configured.example"]` and `readmeNotContains: ["alice@h.example", "/custom/foo"]`. `TestBundle_Readme_AlignsWithDeployTmpl` (forensic drift guard) is unaffected — it exercises the bundle service directly without any CLI flags.
- **Test count:** 21 new cases (9 internal + 7 + 3 + 1 + 1 external). All test names follow project conventions.
- **Architecture:** Change is CLI-layer only. No domain, port, or adapter changes. No new external dependencies.
- **Help text:** All four flags documented. Precedence note clear. Examples block present.
- **CHANGELOG:** Entry present under `[Unreleased] / Added`, specific, references #1245.
- **Branch name:** `feat/1245-bundle-print-deploy-flag` — correct.
- **No mocking frameworks, no panics, no swallowed errors.**

tibtof:
@reviewer addressed: validatePrintDeployFlags moved to first line of runBundle; misleading test comment corrected to match new ordering. Writer's 5-file doc commit unchanged. Ready for re-review.

PR #1251, opened 2026-04-30
feat(#1243): vibew probe retries TLS handshake errors during ACME issuance (--env only, 30s)

### LINKED ISSUE #1243: feat(retro-0.18.4): vibew probe retries TLS handshake errors during ACME issuance (--env only, 30s budget)
## Retro evidence

From the v0.18.4 retrospective (qr-code-azulejos Node deploy, Codex agent):

> First production TLS probe failed with `tls: internal error`. That is technically true but not helpful. A retry worked after Let's Encrypt finished.
> The real state was "ACME certificate issuance in progress." The logs made that clear, but the probe did not.

## Diagnosis

Caddy returns a TLS handshake-level error while it's negotiating an ACME certificate. From the Go TLS client's perspective the wire-level error is opaque — `remote error: tls: internal error`, `tls: handshake failure`, etc. — with no signal that this is transient. Without privileged access to Caddy's admin API, the probe can't distinguish "ACME in progress" (transient, retry) from "broken cert chain" (permanent, fail).

The pattern is reliable enough in practice: TLS handshake error immediately after `docker compose up -d` is almost always ACME-in-progress. After a few seconds Caddy completes the issuance and the handshake succeeds.

## Severity

**HIGH.** Not actively wrong (the TLS error is real), but unactionable for the most common case (probe right after deploy).

## Acceptance criteria

- [ ] When `--env <name>` is set AND the underlying error is a recognized TLS handshake error (substring match on `tls: internal error`, `tls: handshake failure`, `bad certificate`, `tls: protocol version not supported`), retry every 2s for up to 30s.
- [ ] **Progress feedback** every probe iteration:
  ```
  Waiting for ACME issuance... (TLS handshake failed; retrying 30s)
  Waiting for ACME issuance... (15s elapsed)
  Waiting for ACME issuance... (25s elapsed)
  ```
- [ ] On retry exhaustion (30s), exit 1 with:
  ```
  ERROR: TLS handshake failed for 30s.

  Likely ACME (Let's Encrypt) issuance still in progress. Check:
    ssh <host> docker compose logs vibewarden | grep -i acme
  If the cert hasn't been issued yet, retry `vibew probe --env <name>`
  in another minute.
  ```
- [ ] **Default mode (no `--env`)** does NOT retry on TLS errors — localhost dev cert is from Caddy's local CA, immediate. TLS handshake error there is a real bug (cert chain misconfigured), should fail fast.
- [ ] Pin both branches in golden tests: TLS-retry-success, TLS-retry-exhausted.

## Out of scope

- Probing Caddy's admin API directly to read ACME state (rejected at triage — adds privileged port surface).
- Generic HTTP retry beyond TLS errors (rejected — would mask real misconfigurations).

## Companion

Pairs with the connection-refused-message fix (companion issue) — both improve `vibew probe --env <name>` UX during the deploy → verify window.

## Source

Retro: `~/notes/vibewarden/retro-0.18.4.md` (2026-04-30)

tibtof:
Closes #1243

## Summary

- **`ports.ErrTLSHandshake`** sentinel added. Adapter classifies 4 TLS handshake error substrings into it via `isTLSHandshakeError` (same pattern as `isConnectionRefused`).
- **`probe.ErrTLSRetryExhausted`** sentinel added. Service returns it when the 30s budget is exhausted.
- **TLS retry loop** in `Service.Run`: sibling (sequential) to the boot-gap loop, not nested. Only engages when `ErrTLSHandshake` AND `EnvName != ""`. Default mode fails fast — localhost dev cert from Caddy's CA; a TLS error there is a real config bug.
- **`Options.ProgressWriter io.Writer`**: new field; CLI wires to `cmd.ErrOrStderr()`. Progress goes to stderr; final rendered block goes to stdout (machine-parseable).
- **`renderTLSRetryExhausted`**: new render branch with actionable `docker compose logs vibewarden | grep -i acme` message.

**Before (#1243 scenario):** `vibew probe --env production` right after `docker compose up -d` returns `tls: internal error` with no guidance. User doesn't know whether to wait or debug.

**After:** Probe retries every 2s for 30s with progress on stderr:
```
Waiting for ACME issuance... (TLS handshake failed; retrying 30s)
Waiting for ACME issuance... (2s elapsed)
Waiting for ACME issuance... (4s elapsed)
...
```
If cert issues within 30s: falls through to boot-gap check, exits 0 on healthy. If still failing after 30s: exits 1 with:
```
ERROR: TLS handshake failed for 30s.

Likely ACME (Let's Encrypt) issuance still in progress. Check:
  ssh <host> docker compose logs vibewarden | grep -i acme
If the cert hasn't been issued yet, retry `vibew probe --env production`
in another minute.
```

## Test plan

- `TestIsTLSHandshakeError_Substrings` — unit test for all 4 substrings + false cases
- `TestHTTPProber_Probe_TLSHandshakeError` — adapter integration test: raw TCP alert servers for `handshake_failure` (40), `bad_certificate` (42), `internal_error` (80); TLS version mismatch for `protocol version not supported`
- `TestService_TLSRetry_DefaultMode_NoRetry` — TLS error + no EnvName → fail fast, no progress
- `TestService_TLSRetry_EnvMode_SuccessAfterDelay` — TLS error 3x then success → retries, correct progress lines
- `TestService_TLSRetry_EnvMode_Exhausted` — TLS error indefinitely → `ErrTLSRetryExhausted`
- `TestService_TLSRetry_EnvMode_ErrorClassChange` — TLS error then `ErrProbeRefused` → exits loop with new error
- `TestRender_TLSRetryExhausted` — golden test pinned at `tls_retry_exhausted.golden`
- `TestProbeCmd_ProgressWriter_WiredToStderr` — verifies progress does not appear on stdout
- `make check` clean (gofmt + golangci-lint + go test -race ./...)


internal/app/probe/render.go
```diff
@@ -152,6 +155,21 @@ func renderGenericError(w io.Writer, result Result, err error) {
...
 
+// renderTLSRetryExhausted writes the TLS-retry-exhausted error block. The
+// message points the user at docker compose logs vibewarden to check ACME
+// issuance progress.
+func renderTLSRetryExhausted(w io.Writer, result Result) {
+	envName := result.EnvName
+	if envName == "" {
+		envName = "<env>"
+	}
+	fmt.Fprintf(w, "ERROR: TLS handshake failed for 30s.\n")
```
tibtof: **Spec violation — hardcoded `30s` in render and sentinel**

`renderTLSRetryExhausted` prints `"ERROR: TLS handshake failed for 30s."` (line 166) and `ErrTLSRetryExhausted`'s message in `service.go` also bakes in `"30s"`. The architect's spec says: *"The `30s` substring is parameterised by `TLSRetryWait` (rendered as integer seconds)"*.

Right now, if any future test or prod override sets `TLSRetryWait` to e.g. 60s, the rendered output still says 30s — an outright lie to the user.

Fix: add `TLSRetryBudget time.Duration` (or just `TLSRetryWait`) to `Result`, populate it in `Run` before calling `runTLSRetry`, and use `int(result.TLSRetryBudget.Seconds())` in `renderTLSRetryExhausted`. Update the golden test — it stays at `30s` for the default fixture, so the golden file content doesn't change. Also update `ErrTLSRetryExhausted.Error()` to drop the hardcoded duration (the sentinel message doesn't need the number; the render block is the user-facing surface).

internal/cli/cmd/probe_test.go
```diff
@@ -230,3 +230,58 @@ func hostFromAddr(t *testing.T, addr string) string {
...
+
+	// Execute "vibew probe --env staging". The probe will fail with a TLS error
+	// (self-signed cert or connection refused). We only assert that the --env
+	// flag is accepted and that stdout does not contain progress-style output.
+	stdout, _, _ := execProbe(t, []string{"--env", "staging"})
+
+	// Progress messages are written to stderr, not stdout. Confirm stdout does
+	// not contain the "Waiting for ACME" prefix — that would indicate the
+	// ProgressWriter was incorrectly wired to stdout.
+	if strings.Contains(stdout, "Waiting for ACME") {
```
tibtof: **CLI test does not verify that stderr receives progress messages**

The test name is `TestProbeCmd_ProgressWriter_WiredToStderr` and the docstring claims it verifies the ProgressWriter is wired to `cmd.ErrOrStderr()`. But the domain `127.0.0.1` (no server running) yields a connection-refused error — `ErrProbeRefused` — which exits before the TLS retry loop ever fires. As a result, `opts.ProgressWriter` is never written to during this test. The assertion (`stdout does not contain "Waiting for ACME"`) is trivially true and does not validate the wiring.

The spec requires: *"CLI: ProgressWriter wired to stderr"*.

**Why it matters**: if someone accidentally changed `cmd.ErrOrStderr()` to `cmd.OutOrStdout()`, this test would still pass.

**Fix**: Use the `tlsAlertServer` helper (already in `prober_test.go`) or an `httptest.Server` sending a fatal TLS alert to guarantee `ErrTLSHandshake` is returned, triggering the retry loop. Assert `stderr` (the second return value of `execProbe`) contains `"Waiting for ACME issuance..."`. Example sketch:

```go
srvURL, closeSrv := tlsAlertServer(t, 40) // handshake_failure
defer closeSrv()
// write a vibewarden.staging.yaml pointing at srvURL host/port
// execProbe(t, ["--env", "staging"]) → stderr should contain "Waiting for ACME"
_, stderr, _ := execProbe(t, []string{"--env", "staging"})
if !strings.Contains(stderr, "Waiting for ACME issuance...") {
    t.Error("expected progress on stderr, got none")
}
```

internal/app/probe/service_test.go
```diff
@@ -233,3 +236,157 @@ func TestService_ResultURL_PreservedOnError(t *testing.T) {
...
+// and the prober returns ErrTLSHandshake indefinitely, the service returns
+// ErrTLSRetryExhausted after the retry budget is exhausted.
+func TestService_TLSRetry_EnvMode_Exhausted(t *testing.T) {
+	prober := &fakeProber{responses: []fakeResponse{{err: tlsErr()}}}
+	svc := probe.NewService(prober).WithSleep(noSleep)
+
+	_, err := svc.Run(context.Background(), probe.Options{
+		URL:          "https://example.com/_vibewarden/health",
+		EnvName:      "production",
+		TLSRetryWait: 200 * time.Millisecond, // short budget for fast test
```
tibtof, line 349: **Exhausted test: busy-wait on real wall clock, not true sleep injection**

`TLSRetryWait: 200 * time.Millisecond` with `WithSleep(noSleep)` means the deadline in `runTLSRetry` is still driven by `time.Now()`. `noSleep` removes the `s.sleep(poll)` wait but the loop still spins on real CPU for 200ms of wall time before `!time.Now().Before(deadline)` fires. That is not sleep injection — it is a busy-wait.

The spec says sleep injection makes 30s→~0s in tests. The fix here only moves from 30s to 200ms; the loop still burns real time.

**Why it matters**: on a CI runner under load this can take much longer than 200ms per test, and the tight loop hammers the fake prober millions of times per run.

**Preferred fix**: cap the exhausted scenario with a counter rather than a real deadline:

```go
// Option A: use a counter-bounded fakeProber (N TLS errors → then panic so test
// terminates fast on regression), plus TLSRetryWait large enough that the real
// clock never fires, and a clock-injection seam (e.g. Service.now func() time.Time).

// Option B (simplest): add a nowFn func() time.Time seam to Service parallel to
// the sleep seam; test injects a fake clock that advances by TLSRetryWait+1ms
// on the first call after N iterations.
```

This is the lowest-severity of the three findings — 200ms is fast enough not to be a CI problem today — but it is a known-bad pattern that accumulates into a slow test suite.

tibtof:

Reviewed commit `6a0886e` (branch `feat/1243-probe-tls-retry-acme`). Overall the implementation is solid on the critical structural requirements. Three issues found; two require fixes before approval.

---

### What passes

**Sequential vs nested** — The two retry loops are correctly sequential. `Run()` calls `runTLSRetry` as a sibling, not inside the boot-gap loop. On TLS-retry success, the returned `doc` falls through to the boot-gap check cleanly. The godoc on `Run()` spells this out precisely.

**Error-class-change handling** — `TestService_TLSRetry_EnvMode_ErrorClassChange` correctly verifies the loop exits with `ErrProbeRefused` after two TLS errors followed by a connection-refused. `runTLSRetry` returns the new error immediately before the deadline check. This is the most important correctness property of the feature and it is right.

**Sentinel distinctness** — `ErrTLSHandshake` in `ports/`, `ErrTLSRetryExhausted` in `app/probe/`. No cross-contamination. `errors.Is` chain preserved via double-`%w` wrap (valid in Go 1.26). All existing sentinels untouched.

**Adapter classification ordering** — DNS → connection refused → TLS → generic. Correct. No substring overlap between `isConnectionRefused` and `isTLSHandshakeError`.

**CLI wiring** — `opts.ProgressWriter = cmd.ErrOrStderr()` is on the `--env` path only; dev path leaves `ProgressWriter` nil (→ `io.Discard`). `--env` flag still wired correctly post-#1233.

**Golden file** — `tls_retry_exhausted.golden` matches the spec's exhausted block verbatim (literal `<host>`, literal `<name>` replaced with `production`, literal `30s`).

**CHANGELOG** — references `#1243`, describes the feature accurately, cites the v0.18.4 retro Codex finding. Specific and correct.

---

### Finding 1 — MUST FIX: `renderTLSRetryExhausted` hardcodes `"30s"` instead of reading `TLSRetryWait`

See inline comment on `internal/app/probe/render.go` line 166.

The architect's spec is explicit: *"The `30s` substring is parameterised by `TLSRetryWait` (rendered as integer seconds)"*. Both the render function and `ErrTLSRetryExhausted.Error()` bake in `"30s"`. The `Result` struct does not carry the duration, so the renderer has no way to substitute it.

Fix: add `TLSRetryBudget time.Duration` to `Result`; populate it in `Run` (set before calling `runTLSRetry` so it is populated even on exhaustion); render using `int(result.TLSRetryBudget.Seconds())`. The golden file content does not change (default is 30s). Optionally clean up the `ErrTLSRetryExhausted` message to drop the hardcoded number — the user-facing surface is the render block, not the sentinel `Error()` string.

---

### Finding 2 — MUST FIX: CLI stderr-wiring test is a no-op

See inline comment on `internal/cli/cmd/probe_test.go` line 262.

`TestProbeCmd_ProgressWriter_WiredToStderr` uses `domain: 127.0.0.1` with no server → `ErrProbeRefused` → TLS retry loop never fires → `ProgressWriter` is never written to. The assertion (stdout lacks `"Waiting for ACME"`) is trivially true regardless of whether the wiring is correct. This test cannot catch a regression.

Fix: set up a TLS-alert server (alert code 40 / `handshake_failure`) as the probe target, point the staging config at it, run the probe, and assert `stderr` contains `"Waiting for ACME issuance..."`. The `tlsAlertServer` helper already exists in `internal/adapters/health/prober_test.go` but is not exported — either export it (to `internal/testhelpers/`) or duplicate the 15-line raw-TCP listener inline in the CLI test.

---

### Finding 3 — ADVISORY: Exhausted test busy-waits on real wall clock

See inline comment on `internal/app/probe/service_test.go` line 349.

`TLSRetryWait: 200ms + WithSleep(noSleep)` still burns 200ms of real CPU time. The existing boot-gap exhausted test (`TestService_AllRetriesUnknown_ReturnsBootGapExhausted`) has the same pattern. Consistent, but both accumulate into flakiness on heavily loaded CI. Consider adding a `Service.now func() time.Time` seam alongside `sleep` so the deadline can be driven by a fake clock. This is not a merge blocker for this PR, but flag it for a follow-up before the exhausted test count grows further.

tibtof:
@reviewer addressed: TLSRetryBudget added to Result; render and sentinel parameterized by Options.TLSRetryWait. ProgressWriter test now uses a raw TLS-alert server, asserts stderr contains progress AND stdout does not. Writer's 8-file doc commit unchanged. Ready for re-review.

Advisory (finding 3 — busy-wait on real wall clock in TestService_TLSRetry_EnvMode_Exhausted): left for follow-up; not addressed in this commit.

PR #1252, opened 2026-04-30
feat(#1246): vibew doctor --preflight <env> — pre-deploy validation

### LINKED ISSUE #1246: feat(retro-0.18.4 follow-up): vibew doctor --preflight <env> — pre-deploy validation (DNS, server.port, arch, TLS email)
## Retro evidence (Codex v0.18.4 follow-up feedback)

> Add a `vibew doctor --preflight production`. Not deployment, just validation before bundle/deploy:
> - DNS for `tls.domain` points somewhere plausible
> - production `server.port` is 443
> - target platform is set
> - image exists and arch matches
> - warns if TLS email is absent, but not fatal for Let's Encrypt

## Diagnosis

`vibew doctor` today validates static config (post-#1198, #1222). `vibew probe` validates runtime (post-#1233). There's no pre-deploy validation step that catches the most common production-deploy mistakes BEFORE the user runs `vibew bundle` and ssh's the bundle to a host.

A `--preflight production` mode runs a focused set of checks that only matter when the next step is "deploy to a real host":

- DNS resolution for `tls.domain` to ANY IP (doesn't need to be the right host; just "DNS is configured at all").
- `server.port` is 443 in the resolved production config (a common mis-merge has it staying at 8443).
- `deploy.target_platform` is set (post-#1200 default is linux/amd64; preflight asserts it's not empty).
- App image exists locally AND its arch matches the target platform (image-inspect post-#1219 already does this for `vibew bundle`; here we do it before bundle).
- TLS email warning (not fatal for Let's Encrypt — they accept anonymous registration — but a missing email surface during certificate issuance failures is harder to debug; warn, don't block).

## Severity

**MEDIUM.** New surface, meaningful payoff for the "agent runs bundle without thinking" path. Not blocking for v0.18.5 if scope grows.

## Acceptance criteria

- [ ] `vibew doctor --preflight production` (or `--preflight <env>` matching #1233's env-resolver pattern) runs the following checks IN ADDITION to the existing static checks:
  - **DNS resolution.** Resolve `<merged.tls.domain>` via Go's stdlib resolver. PASS if any A/AAAA record returns. WARN if NXDOMAIN (DNS not configured yet — common during initial setup). FAIL if the lookup itself errors (transient network — surface).
  - **Production server.port.** Read merged `server.port`. PASS if 443. WARN if anything else (likely a forgotten override).
  - **deploy.target_platform.** PASS if set. FAIL if empty (post-#1200 default writes it; this catches users who deleted it).
  - **App image existence + arch match.** Reuse #1219's image-inspect path: app image with `<project>-app:latest` (or user-set `app.image`) exists locally; its `Architecture` field matches `deploy.target_platform`. FAIL on mismatch. WARN if the image doesn't exist locally (might be a registry image — can't verify locally).
  - **TLS email warning.** Read merged `tls.email` (or wherever the ACME email lives). WARN if empty + `tls.provider: letsencrypt`. Not fatal.
- [ ] Reuse the env-resolver from #1233 (`internal/app/env`).
- [ ] Output format mirrors existing `vibew doctor` rows.
- [ ] When `--preflight` is set: STATIC config + Dockerfile checks still run, then the preflight section runs after.

## Out of scope

- Probing the remote host (that's `vibew probe --env <name>` post-#1233).
- DNS-correctness checks (does the A record point at the right IP) — too brittle for many setups (Cloudflare, multi-region).
- Live SSH validation (would require `--host` config; explicitly NOT a deployment tool).

## Source

Retro: `~/notes/vibewarden/retro-0.18.4.md` (2026-04-30) — Codex follow-up feedback recommendation #7.

tibtof:
Closes #1246

## Summary

- New `--preflight <env>` flag on `vibew doctor`. When set it reads `vibewarden.<env>.yaml` via the shared env-resolver (ADR-102), merges it with `vibewarden.yaml`, and appends a **Preflight: \<env\>** section with five checks:
  1. **DNS resolves** — `tls.domain` resolves to at least one IP; NXDOMAIN or lookup failure → WARN (doctor stays runnable offline)
  2. **Production port** — `server.port == 443`; other value → WARN
  3. **Target platform** — `deploy.target_platform` is non-empty; empty → FAIL
  4. **App image arch** — inspects the local image and compares OS/arch against `deploy.target_platform`; mismatch → FAIL; image not found or Docker unavailable → WARN
  5. **TLS email** — `tls.email` set when provider is `letsencrypt`; empty → WARN
- Missing override file (`vibewarden.<env>.yaml`) exits immediately with a clear message — no checks run.
- Static Config & Docker checks still run first; preflight section appends afterward.
- No new external dependencies. Reuses `ports.ImageInspector` and `app/env.FileResolver`.

## Sample output — all 5 outcomes

```
VibeWarden Doctor
─────────────────────────────────────────

  Config & Docker
  [OK]    Config file              vibewarden.yaml — valid
  [OK]    Docker daemon            running
  [OK]    Docker Compose           Docker Compose version v2.35.1
  [OK]    Proxy port               port 8443 is available
  [OK]    ACME email               not using ZeroSSL — email not required

  Preflight: production
  [OK]    DNS resolves             demo.example.com → 178.104.159.3
  [WARN]  Production port          server.port = 8443 (expected 443 for production)
  [OK]    Target platform          linux/amd64
  [OK]    App image arch           linux/amd64 matches deploy.target_platform
  [WARN]  TLS email                empty (Let's Encrypt accepts anonymous, but warnings won't reach you. Set tls.email in vibewarden.<env>.yaml)
```

## Files changed (7)

**New (5):**
- `/internal/ports/dns_resolver.go` — `DNSResolver` port
- `/internal/adapters/ops/dns_resolver.go` — `net.Resolver`-backed adapter with 3s timeout
- `/internal/adapters/ops/dns_resolver_test.go`
- `/internal/app/ops/preflight.go` — orchestrator + 5 helpers
- `/internal/app/ops/preflight_test.go` — table-driven tests per helper + integration tests

**Modified (2):**
- `/internal/app/ops/doctor.go` — `WithPreflight` optional builder; `runChecks` appends preflight section
- `/internal/cli/cmd/doctor.go` — `--preflight` flag; env-resolver wiring
- `/internal/cli/cmd/doctor_test.go` — flag registration + CLI error-path tests

## Test plan

- [ ] `make check` passes locally (pre-push hook confirmed)
- [ ] `go test -race ./internal/app/ops/... ./internal/adapters/ops/... ./internal/cli/cmd/...` all green
- [ ] `TestCheckDNSResolves_PASS_WARN_FAIL` — 7 cases (single IP, 4 IPs truncated, 3 IPs exact, NXDOMAIN error, empty addrs, generic error, empty domain)
- [ ] `TestCheckProductionPort_PASS_WARN` — 4 cases (443, 8443, 0, 9000)
- [ ] `TestCheckTargetPlatform_PASS_FAIL` — 3 cases (linux/amd64, linux/arm64, empty)
- [ ] `TestCheckImageArch_PASS_WARN_FAIL` — 7 cases (match, mismatch, not found, Docker unavailable, generic error, no image, project-name fallback)
- [ ] `TestCheckTLSEmail_PASS_WARN` — 4 cases (LE no email, LE with email, self-signed, external)
- [ ] `TestDoctorService_Run_PreflightSection_AppendedAfterStatic` — section ordering
- [ ] `TestDoctorService_Run_PreflightDisabled_NoSection` — regression guard
- [ ] `TestDoctorService_Run_PreflightFAIL_AffectsExitCode` — exit code wiring
- [ ] `TestDoctorCmd_PreflightFlag_MissingEnvFile_ErrorsClearly` — CLI error path
- [ ] `TestDoctorCmd_PreflightFlag_AbsentByDefault` — no section without flag


internal/cli/cmd/doctor.go
```diff
@@ -68,28 +70,57 @@ tls.skip_rate_limit_check: true in vibewarden.yaml) to suppress this check.
...
+			// replaces the base config for ALL checks, including the static ones,
+			// so that the doctor report reflects what will be deployed. On any
+			// resolver error we exit immediately with a clear message — no checks
+			// are run (mirrors vibew probe --env behaviour per ADR-102).
+			var cfg *config.Config
+			if preflightEnv != "" {
+				resolver := envapp.NewFileResolver(workDir)
+				resolved, resolveErr := resolver.Resolve(preflightEnv)
+				if resolveErr != nil {
+					return fmt.Errorf("config file not found: vibewarden.%s.yaml", preflightEnv)
```
tibtof: **Architecture violation (minor): error from `resolver.Resolve` is swallowed — misleading message when `vibewarden.yaml` is also missing.**

Line 105 discards `resolveErr` entirely and returns a hardcoded message that always blames the override file:

```go
return fmt.Errorf("config file not found: vibewarden.%s.yaml", preflightEnv)
```

When `vibewarden.yaml` itself is missing, `FileResolver.Resolve` returns `ErrBaseConfigMissing`, but this code reports `vibewarden.production.yaml not found` — pointing the user at the wrong file.

Per ADR-102 (and the issue spec at step 3): `ErrOverrideConfigMissing` → blame override file, `ErrBaseConfigMissing` → blame base file.

Fix:
```go
if resolveErr != nil {
    if errors.Is(resolveErr, envapp.ErrOverrideConfigMissing) {
        return fmt.Errorf("config file not found: vibewarden.%s.yaml", preflightEnv)
    }
    return fmt.Errorf("config error: %w", resolveErr)
}
```

This also preserves the root cause in the error chain ("errors wrapped with context" — CLAUDE.md code-style rule) instead of hiding it.

internal/app/ops/preflight.go
```diff
@@ -0,0 +1,242 @@
...
+		default:
+			return CheckResult{
+				Name:     "App image arch",
+				Severity: SeverityWarn,
+				Detail:   fmt.Sprintf("could not inspect image %q: %v", tag, err),
+			}
+		}
+	}
+
+	imgPlatform := info.Platform()
```
tibtof, line 212: **Bug: `Platform()` returns `"/amd64"` (leading slash) when `info.OS` is empty — arch comparison silently fails.**

`ports.ImageInfo.Platform()` only guards the all-empty case:
```go
if i.OS == "" && i.Architecture == "" {
    return ""
}
return i.OS + "/" + i.Architecture
```

When an older Docker daemon omits the `Os` field (common with images built without explicit `--platform`), `info.OS` is `""` but `info.Architecture` is `"amd64"`. `Platform()` then returns `"/amd64"`, which never equals `"linux/amd64"`. The check incorrectly FAILs with:

`image arch /amd64 does not match deploy.target_platform linux/amd64 — rebuild with --platform=linux/amd64`

This contradicts the spec ("What happens if … `info.OS` is empty (older Docker)? Defensive: treat empty OS as "linux" or warn").

Fix in `checkImageArch` (after successful `Inspect`):
```go
imgPlatform := info.Platform()
if imgPlatform == "" || (info.OS == "" && info.Architecture != "") {
    return CheckResult{
        Name:     "App image arch",
        Severity: SeverityWarn,
        Detail:   fmt.Sprintf("image %q has no OS metadata (older Docker?) — skipping arch check", tag),
    }
}
```

Alternatively, default `info.OS` to `"linux"` before calling `Platform()` since that is Docker's implicit OS for all Linux images, which is the stated intent.

A test case is also missing: `ImageInfo{OS: "", Architecture: "amd64"}` with `targetPlatform: "linux/amd64"` must produce WARN, not FAIL.

internal/app/ops/preflight_test.go
```diff
@@ -0,0 +1,662 @@
...
+			resolverAddrs: []string{"1.2.3.4", "5.6.7.8", "9.10.11.12", "13.14.15.16"},
+			wantSeverity:  ops.SeverityOK,
+			wantSubstr:    "...",
+		},
+		{
+			name:          "OK — exactly 3 addresses no ellipsis",
+			domain:        "demo.example.com",
+			resolverAddrs: []string{"1.2.3.4", "5.6.7.8", "9.10.11.12"},
+			wantSeverity:  ops.SeverityOK,
+			wantSubstr:    "demo.example.com",
```
tibtof, line 76: **Test gap: "exactly 3 addresses" case does not assert absence of ellipsis.**

The test case at this line asserts `wantSubstr: "demo.example.com"`, which will also match when `"..."` is present (the 4-address truncated case). The test name says "no ellipsis" but nothing verifies it.

This is the exact pin the review spec called for:

> DNS truncation to 3 IPs: pin in test.

The truncation boundary is the critical invariant — if someone changes the constant from 3 to 4, this test still passes.

Fix: add a negative assertion in the test loop (or add a dedicated field):
```go
// After verifying the positive badge/substring...
if strings.Contains(out, "...") {
    t.Errorf("exactly-3 case must not contain ellipsis in output:\n%s", out)
}
```
Since the test drives through `svc.Run` which renders the full report, the substring check on `out` is sufficient — just add the negative check.

tibtof:

3 issues found — 2 blocking, 1 minor.

---

**Finding 1 (BLOCKING): image arch comparison produces wrong result for empty-OS images**

`internal/app/ops/preflight.go` line 195

`ports.ImageInfo.Platform()` only guards the both-empty case. When an older Docker daemon omits the `Os` JSON field, `info.OS == ""` but `info.Architecture == "amd64"`. `Platform()` returns `"/amd64"`, which never equals `"linux/amd64"`. The check fires a spurious FAIL instead of the specified WARN. The spec explicitly calls this out: "treat empty OS as 'linux' or warn". Neither is implemented.

Fix: after a successful `Inspect`, guard for empty OS before calling `Platform()` — either default it to `"linux"` (Docker's universal implicit OS) or return WARN with a "no OS metadata" message. Also add a test case: `ImageInfo{OS: "", Architecture: "amd64"}` against `targetPlatform: "linux/amd64"` must produce WARN.

---

**Finding 2 (BLOCKING): resolver error silently swallowed — misleading message when `vibewarden.yaml` is missing**

`internal/cli/cmd/doctor.go` line 105

The catch-all `if resolveErr != nil` branch discards the underlying error and always reports `vibewarden.<env>.yaml not found`. When `vibewarden.yaml` itself is absent, `FileResolver.Resolve` returns `ErrBaseConfigMissing`, but the user sees the wrong filename. This also violates the CLAUDE.md code-style rule ("errors wrapped with context: `fmt.Errorf("doing X: %w", err)`").

Fix: distinguish `ErrOverrideConfigMissing` vs `ErrBaseConfigMissing` (or any other error) and preserve the original error in the chain for the non-override-missing case.

---

**Finding 3 (minor): "exactly 3 addresses" test case does not assert absence of ellipsis**

`internal/app/ops/preflight_test.go` line 75

Test name says "no ellipsis" but `wantSubstr: "demo.example.com"` is satisfied even when `"..."` is present in output. If the truncation threshold changes from 3 to 4, the test still passes. Add a `strings.Contains(out, "...")  → t.Errorf` negative assertion in that table case.

---

**What is clean:**
- Hexagonal structure is correct: port in `internal/ports/dns_resolver.go`, adapter in `internal/adapters/ops/dns_resolver.go`, logic in `internal/app/ops/preflight.go`. No layering violations.
- DNS adapter correctly applies `context.WithTimeout(ctx, 3*time.Second)` and defers cancel — no leak.
- Image-inspector reuse is correct: `opsadapter.NewImageInspectAdapter()` from `internal/adapters/ops/image_inspect.go` — no duplicate code path.
- All 5 check helpers have godoc. All exported symbols in new files have godoc.
- Error wrapping with `%w` is used consistently everywhere except Finding 2.
- No global state, no `init()`, no `panic` outside main.
- No new dependencies — no license review needed.
- 35 test cases present and table-driven. DNS timeout, section ordering, exit-code wiring, and CLI flag registration all tested.
- `TestDoctorCmd_PreflightFlag_MissingEnvFile_ErrorsClearly` tests the right path but will pass even with the wrong error message because it only checks that `err != nil` and the string contains `vibewarden.production.yaml` — which the broken code does produce. Once Finding 2 is fixed, re-verify the test still covers the `ErrBaseConfigMissing` branch.

tibtof:
**Writer-agent review: APPROVE** (cannot self-approve via API — flagging for merge)

Documentation review: docs consistent with code changes after writer-agent fixes.

## Findings (all fixed in commit `61963ea`)

**Finding 1: `llms-full.txt` — CLI table row and flags section missing `--preflight`**
The `vibew doctor` row had no mention of `--preflight`. The flags block listed only `--skip-le-preflight`. The five preflight checks were entirely absent from the doctor checks block. Fixed: added a `vibew doctor --preflight <env>` table row, enumerated all four flags, and added a `Preflight checks` block with P1–P5 and ADR-100/ADR-102 back-references.

**Finding 2: `docs/troubleshooting.md` — no `--preflight` flag entry and no preflight section**
The flags table had three rows; `--preflight` was missing. The checks section had no Layer 3. No troubleshooting guidance existed for the five new checks or the "env file not found" error. Fixed: added flag row, Layer 3 checks table, and a full `Pre-deploy preflight` section with per-check failure causes, fix instructions, and the exact error message an agent would grep for.

**Finding 3: deploy template, golden, agents template, static example, agent-kickoff.md — `--preflight` absent from all agent-facing surfaces**
The deploy template had no mention of the preflight gate between Step 6 and Step 7. The agents template CLI table had a bare `vibew doctor | Diagnose issues` row with no `--preflight` variant. All agent-facing surfaces were silent on the new command. Fixed across all five surfaces.

All tests pass (pre-push hook ran full suite). Docs are now consistent with the code changes.

tibtof:
@reviewer addressed: empty-OS image WARN handling added with new test case; resolveErr properly wrapped with branched messages for ErrBaseConfigMissing vs ErrOverrideConfigMissing; ellipsis-absence pinned in 3-IP test case. Writer's 8-file doc commit unchanged. Ready for re-review.

PR #1317, opened 2026-05-05
fix(#1267): authui — reject external return_to URLs

tibtof:
Closes #1267

## Summary

- **Server-side** (`internal/adapters/authui/handler.go`): `returnToQuery` now delegates to `isSafeReturnTo` before accepting a `return_to` value. Only relative paths starting with `/` pass; `//evil.com`, `/\evil.com`, absolute URLs with any scheme (`https:`, `javascript:`, `data:`, etc.), values whose `url.Parse` result has a non-empty `Scheme` or `Host`, and values containing CRLF characters are all rejected. Rejected values produce an empty string (no `return_to` query fragment on inter-page links).

- **Client-side** (`templates/login.html`, `templates/registration.html`): A `safeReturnTo()` JS function is added before `returnTo` is assigned. It mirrors the server-side rules using the browser's `URL()` constructor: rejects non-`/`-prefixed values, `//`/`/\` prefixes, and any value whose parsed `origin` differs from `window.location.origin`. Falls back to `/` on any invalid input.

## Test plan

- `TestIsSafeReturnTo` — 24 table-driven cases covering valid paths, all external URL forms, protocol-relative, backslash trick, CRLF injection, schemeless values without leading slash, and empty string.
- `TestReturnToQuery_ServerSide` — 9 cases exercising the full `returnToQuery` function via `httptest.NewRequest`.
- `TestReturnToQuery_ExternalURLDroppedFromPage` — integration-style test: starts a real `Handler`, GETs login and registration pages with 4 malicious `return_to` payloads each, asserts the payload does not appear anywhere in the rendered HTML.
- All existing `handler_test.go` tests continue to pass.
- `make check` passes (gofmt, golangci-lint, go build, go test -race).

### DIFF
--- a/internal/adapters/authui/handler.go
+++ b/internal/adapters/authui/handler.go
@@ -281,7 +281,7 @@ func returnToQuery(r *http.Request) string {
 	if !isSafeReturnTo(v) {
 		return ""
 	}
-	return "?return_to=" + v
+	return "?return_to=" + url.QueryEscape(v)
 }
 
 // isSafeReturnTo reports whether v is safe to use as a return_to redirect
--- a/internal/adapters/authui/handler.go
+++ b/internal/adapters/authui/handler.go
@@ -24,6 +24,8 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"net/url"
+	"strings"
 	"time"
 )
 
@@ -264,11 +266,55 @@ func applyDefaults(cfg *AuthUIConfig) {
 
 // returnToQuery extracts the return_to query parameter from r and returns
 // a ready-to-append query string fragment like "?return_to=%2Fdashboard",
-// or an empty string when the parameter is absent or empty.
+// or an empty string when the parameter is absent, empty, or not a safe
+// same-origin relative path.
+//
+// Security: only values that are relative paths (no scheme, no host) are
+// accepted. Values starting with "//" or "/\" are rejected even though
+// url.Parse may classify them as scheme-less, because browsers interpret
+// "//evil.com" as a protocol-relative URL pointing to an external host.
+// "javascript:" URIs and any other value with a non-empty Scheme or Host
+// are also rejected. This prevents open-redirect attacks where an attacker
+// supplies an absolute external URL as return_to.
 func returnToQuery(r *http.Request) string {
 	v := r.URL.Query().Get("return_to")
-	if v == "" {
+	if !isSafeReturnTo(v) {
 		return ""
 	}
 	return "?return_to=" + v
+}
+
+// isSafeReturnTo reports whether v is safe to use as a return_to redirect
+// target. It accepts only relative paths that start with "/" and are not
+// protocol-relative ("//…") or Windows-style path-separator tricks ("/\…").
+// Any value with a URL scheme (e.g. "https:", "javascript:") or a host
+// component is rejected. An empty string is rejected (callers handle the
+// absence of return_to separately).
+func isSafeReturnTo(v string) bool {
+	if v == "" {
+		return false
+	}
+	// Reject protocol-relative URLs ("//evil.com") and backslash variants
+	// ("/\evil.com") before url.Parse, because some browsers treat these as
+	// absolute URLs pointing to external hosts.
+	if strings.HasPrefix(v, "//") || strings.HasPrefix(v, "/\\") {
+		return false
+	}
+	// Reject anything that doesn't start with "/" — relative paths without a
+	// leading slash could be resolved relative to the current page in ways that
+	// surprise callers.
+	if !strings.HasPrefix(v, "/") {
+		return false
+	}
+	// Reject values containing newlines or carriage returns (HTTP header
+	// injection / log injection defence).
+	if strings.ContainsAny(v, "\r\n") {
+		return false
+	}
+	parsed, err := url.Parse(v)
+	if err != nil {
+		return false
+	}
+	// After url.Parse a safe relative path must have an empty Scheme and Host.
+	return parsed.Scheme == "" && parsed.Host == ""
 }
--- a/internal/adapters/authui/return_to_test.go
+++ b/internal/adapters/authui/return_to_test.go
@@ -1,0 +1,210 @@
+package authui
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// TestIsSafeReturnTo verifies that isSafeReturnTo accepts only relative,
+// same-origin paths and rejects all forms of open-redirect payload.
+func TestIsSafeReturnTo(t *testing.T) {
+	tests := []struct {
+		name string
+		v    string
+		want bool
+	}{
+		// Valid relative paths.
+		{name: "root path", v: "/", want: true},
+		{name: "dashboard path", v: "/dashboard", want: true},
+		{name: "nested path", v: "/admin/users", want: true},
+		{name: "path with query", v: "/path?q=1", want: true},
+		{name: "path with fragment", v: "/path#section", want: true},
+		{name: "path with encoded chars", v: "/path%20with%20spaces", want: true},
+
+		// Absolute external URLs — must be rejected.
+		{name: "https external", v: "https://evil.com", want: false},
+		{name: "http external", v: "http://evil.com", want: false},
+		{name: "ftp scheme", v: "ftp://evil.com/path", want: false},
+		{name: "javascript scheme", v: "javascript:alert(1)", want: false},
+		{name: "javascript uppercase", v: "JAVASCRIPT:alert(1)", want: false},
+		{name: "data URI", v: "data:text/html,<h1>evil</h1>", want: false},
+		{name: "vbscript scheme", v: "vbscript:msgbox(1)", want: false},
+
+		// Protocol-relative URLs — browsers treat these as absolute.
+		{name: "protocol-relative double slash", v: "//evil.com", want: false},
+		{name: "protocol-relative with path", v: "//evil.com/path", want: false},
+
+		// Backslash tricks — some browsers interpret /\ as //.
+		{name: "backslash trick", v: "/\\evil.com", want: false},
+		{name: "backslash with path", v: "/\\evil.com/path", want: false},
+
+		// HTTP response-splitting / log-injection variants.
+		{name: "newline injection", v: "/path\nhttp://evil.com", want: false},
+		{name: "carriage return injection", v: "/path\rhttp://evil.com", want: false},
+		{name: "crlf injection", v: "/path\r\nSet-Cookie: x=y", want: false},
+
+		// Values without leading slash.
+		{name: "no leading slash", v: "evil.com", want: false},
+		{name: "relative without slash", v: "dashboard", want: false},
+
+		// Empty value.
+		{name: "empty string", v: "", want: false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := isSafeReturnTo(tt.v)
+			if got != tt.want {
+				t.Errorf("isSafeReturnTo(%q) = %v, want %v", tt.v, got, tt.want)
+			}
+		})
+	}
+}
+
+// TestReturnToQuery_ServerSide verifies that returnToQuery propagates safe
+// return_to values and drops unsafe ones.
+func TestReturnToQuery_ServerSide(t *testing.T) {
+	tests := []struct {
+		name      string
+		returnTo  string
+		wantQuery string
+	}{
+		{
+			name:      "safe path propagated",
+			returnTo:  "/dashboard",
+			wantQuery: "?return_to=/dashboard",
+		},
+		{
+			name:      "safe root path propagated",
+			returnTo:  "/",
+			wantQuery: "?return_to=/",
+		},
+		{
+			name:      "safe path with query params propagated",
+			returnTo:  "/search?q=hello",
+			wantQuery: "?return_to=/search?q=hello",
+		},
+		{
+			name:      "external https URL rejected",
+			returnTo:  "https://evil.com",
+			wantQuery: "",
+		},
+		{
+			name:      "protocol-relative URL rejected",
+			returnTo:  "//evil.com",
+			wantQuery: "",
+		},
+		{
+			name:      "backslash trick rejected",
+			returnTo:  "/\\evil.com",
+			wantQuery: "",
+		},
+		{
+			name:      "javascript scheme rejected",
+			returnTo:  "javascript:alert(1)",
+			wantQuery: "",
+		},
+		{
+			name:      "newline injection rejected",
+			returnTo:  "/path\nhttp://evil.com",
+			wantQuery: "",
+		},
+		{
+			name:      "empty return_to yields empty query",
+			returnTo:  "",
+			wantQuery: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Build request by setting RawQuery directly to avoid httptest.NewRequest
+			// panicking on control characters (newlines) in the URL string.
+			r := httptest.NewRequest(http.MethodGet, "/_vibewarden/login", nil)
+			if tt.returnTo != "" {
+				// Use url.Values to encode the value exactly as a browser would
+				// send it, then override with the raw value so control-char cases
+				// actually reach the validation logic unencoded.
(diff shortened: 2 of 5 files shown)


tibtof (commented):
Documentation review: one required change before merge.

**Missing CHANGELOG entry** (`CHANGELOG.md`, `[Unreleased]` section)

Security fixes must have a changelog entry. The `[Unreleased]` section is currently empty. Required addition:

```markdown
## [Unreleased]

### Security

- **fix(#1267): authui — reject external return_to URLs.** Server-side `isSafeReturnTo` in `handler.go` now rejects any `return_to` value that is not a relative same-origin path (rejects `//`, `/\`, absolute URLs with any scheme, CRLF-injected values, and values without a leading `/`). Client-side `safeReturnTo()` in `login.html` and `registration.html` mirrors these rules via the browser `URL()` constructor. Prevents open-redirect attacks where an attacker supplies an absolute external URL as `return_to`. (Closes #1267)
```

**All other surfaces checked — no drift found:**

- `llms-full.txt` §auth.ui (lines 415–425): documents built-in UI config fields only; no `return_to` claim present.
- `docs/production-hardening.md` line 323–324: references Kratos-level `selfservice.allowed_return_urls`, separate from VibeWarden's auth UI handler, accurate and unaffected.
- `docs/configuration.md` §auth.ui: config field table, no `return_to` behavior claim.
- `AGENTS-VIBEWARDEN.md`: no `return_to` mention.
- `vibewarden.reference.yaml`: no `return_to` field documented.
- `docs/getting-started.md`: no `return_to` mention.
- `docs/examples/`: no `return_to` mention.
- `README.md`: no `return_to` behavior claim.

PR body is adequate for a security fix: both layers (server-side Go + client-side JS) described, attack vectors listed, test plan references specific test function names and case counts.

WRITER VERDICT: **changes required** — add the `### Security` entry under `[Unreleased]` in `CHANGELOG.md` then this is clear to approve.

internal/adapters/authui/handler.go
```diff
@@ -264,11 +266,55 @@ func applyDefaults(cfg *AuthUIConfig) {
...
+// "javascript:" URIs and any other value with a non-empty Scheme or Host
+// are also rejected. This prevents open-redirect attacks where an attacker
+// supplies an absolute external URL as return_to.
 func returnToQuery(r *http.Request) string {
 	v := r.URL.Query().Get("return_to")
-	if v == "" {
+	if !isSafeReturnTo(v) {
 		return ""
 	}
 	return "?return_to=" + v
```
tibtof: **Bug: `returnToQuery` inserts the value unencoded — query parameter injection and malformed links**

`returnToQuery` returns `"?return_to=" + v` without calling `url.QueryEscape(v)`. When `v` is a validated but raw string like `/dashboard?tab=settings` or `/path&other=1`, the resulting query string is malformed or injects extra parameters:

```go
// Current — broken:
return "?return_to=" + v

// Fix:
return "?return_to=" + url.QueryEscape(v)
```

Concrete impact:

- `v = "/path&other=1"` → `?return_to=/path&other=1` — `other=1` becomes a separate query param. `html/template` encodes `&` to `&amp;` in the href, but the browser decodes `&amp;` back to `&` before following the link, so the next page receives `?return_to=/path&other=1`. `url.Query().Get("return_to")` then returns `"/path"` only — the `other=1` part is silently dropped and `isSafeReturnTo` is called on a truncated path. No open-redirect bypass here, but data loss on every return_to that contains `&`.
- `v = "/path?q=1"` → `?return_to=/path?q=1` — technically parseable (RFC 3986 permits `?` in query values) but non-idiomatic and breaks strict parsers.

**Fix**: wrap with `url.QueryEscape` in `returnToQuery`. Note that `url.QueryEscape` encodes `/` as `%2F`; if preserving slashes is desired use `url.PathEscape` only on the query value portion, or use `url.Values{"return_to": {v}}.Encode()` which encodes `&` and `?` correctly while leaving `/` as-is via `url.Values.Encode` (it uses `url.QueryEscape` internally which DOES encode `/` — so use `strings.NewReplacer("%2F", "/")` if needed, or simply accept the encoding). The simplest correct form:

```go
return "?return_to=" + url.QueryEscape(v)
```

The test suite has no case for `v = "/path&other=1"` or `v = "/path?q=1"` going through `returnToQuery` — the existing `TestReturnToQuery_ServerSide` case `"safe path with query params propagated"` checks `wantQuery: "?return_to=/search?q=hello"` which is the incorrect (unencoded) output, cementing the bug in the test expectation.

internal/adapters/authui/return_to_test.go
```diff
@@ -0,0 +1,210 @@
...
+				// send it, then override with the raw value so control-char cases
+				// actually reach the validation logic unencoded.
+				q := r.URL.Query()
+				q.Set("return_to", tt.returnTo)
+				r.URL.RawQuery = q.Encode()
+				// For payloads containing control characters the encoded form
+				// would be percent-escaped and never reach isSafeReturnTo's
+				// newline check via Query().Get(). Simulate an attacker who
+				// bypasses encoding by setting RawQuery directly.
+				if tt.returnTo == "/path\nhttp://evil.com" {
```
tibtof: **Test bug: hardcoded special case inside table-driven runner breaks the table-driven pattern**

The newline injection test case has a special-cased branch inside the loop body:

```go
if tt.returnTo == "/path\nhttp://evil.com" {
    r.URL.RawQuery = "return_to=/path%0ahttp://evil.com"
}
```

This is exactly what table-driven tests are supposed to avoid. The `rawQuery` setup should be a field in the test struct, not a string-matched branch inside the runner. Add a `rawQuery string` field to the struct:

```go
tests := []struct {
    name      string
    returnTo  string
    rawQuery  string   // optional: if non-empty, override RawQuery directly
    wantQuery string
}{
    ...
    {
        name:      "newline injection rejected",
        returnTo:  "/path\nhttp://evil.com",
        rawQuery:  "return_to=/path%0ahttp://evil.com",
        wantQuery: "",
    },
}
```

Then in the runner:
```go
if tt.rawQuery != "" {
    r.URL.RawQuery = tt.rawQuery
} else if tt.returnTo != "" {
    q := r.URL.Query()
    q.Set("return_to", tt.returnTo)
    r.URL.RawQuery = q.Encode()
}
```

The current code would silently fail to test any future newline case added to the table since the branch only matches by exact string value.

internal/adapters/authui/return_to_test.go
```diff
@@ -0,0 +1,210 @@
...
+				defer h.Stop(context.TODO()) //nolint:errcheck
+
+				resp, err := http.Get("http://" + h.Addr() + pg.path + "?return_to=" + payload)
+				if err != nil {
+					t.Fatalf("GET: %v", err)
+				}
+				defer resp.Body.Close() //nolint:errcheck
+
+				var buf [32768]byte
+				n, _ := resp.Body.Read(buf[:])
```
tibtof: **Bug: integration test reads only the first 32 KB of the response body**

```go
var buf [32768]byte
n, _ := resp.Body.Read(buf[:])
body := string(buf[:n])
```

`Read` is not guaranteed to fill the buffer — it may return fewer bytes than available, and for large pages it will miss content that appears after the first 32 KB. A login or registration page with inline CSS, HTML, and embedded JavaScript can easily exceed 32 KB. If a payload appears in the second half of the page, the test will not catch it.

Use `io.ReadAll` instead:

```go
import "io"

bodyBytes, err := io.ReadAll(resp.Body)
if err != nil {
    t.Fatalf("reading body: %v", err)
}
body := string(bodyBytes)
```

This also silently swallows the `Read` error via `n, _ =` — another reason to use `io.ReadAll` which surfaces errors cleanly.

internal/adapters/authui/return_to_test.go
```diff
@@ -0,0 +1,210 @@
...
+				if contains(body, payload) {
+					t.Errorf("page %s leaked external return_to %q into body", pg.path, payload)
+				}
+			})
+		}
+	}
+}
+
+// contains is a substring check to avoid importing strings in this file.
+func contains(s, substr string) bool {
```
tibtof: **Unnecessary reimplementation of `strings.Contains`**

```go
// contains is a substring check to avoid importing strings in this file.
func contains(s, substr string) bool {
    return len(substr) > 0 && len(s) >= len(substr) && func() bool {
        for i := 0; i <= len(s)-len(substr); i++ {
            if s[i:i+len(substr)] == substr {
                return true
            }
        }
        return false
    }()
}
```

The comment says "to avoid importing strings in this file" but:

1. `strings` is stdlib — there is no import cost concern.
2. Test files routinely import `strings`; there is no project rule against it.
3. This implementation does a naive O(n*m) byte-by-byte scan without any of the optimizations in `strings.Contains` (which uses Rabin-Karp for large inputs).
4. The inline IIFE pattern is harder to read than a named helper.

Replace with:

```go
import "strings"

// In the assertion:
if strings.Contains(body, payload) {
    t.Errorf(...)
}
```

Delete the `contains` function entirely.

tibtof:

CI is still pending (Build & Test, Lint, Integration Tests, Trivy all queued) — cannot approve while checks are running.

Beyond CI, three bugs require fixes before this can merge:

**1. returnToQuery does not encode the value (handler.go:284) — correctness bug**
`return "?return_to=" + v` inserts the raw validated string without `url.QueryEscape`. When `v` contains `&` (e.g. `/dashboard?tab=settings&modal=open`), html/template HTML-encodes `&` to `&amp;` in the href; the browser then decodes `&amp;` back to `&` before navigation, causing the next page to parse `return_to` as only the prefix before the `&`. The return destination is silently truncated. Also: the `TestReturnToQuery_ServerSide` case `"safe path with query params propagated"` asserts `wantQuery: "?return_to=/search?q=hello"` which is the wrong expected value — it cements the bug in the test. Fix: `return "?return_to=" + url.QueryEscape(v)`.

**2. Integration test truncates the response body at 32 KB (return_to_test.go:187–189)**
`var buf [32768]byte; n, _ := resp.Body.Read(buf[:])` does a single `Read` call which may return fewer bytes than the full page. On a sufficiently large template, a leaked payload at the end of the page body would go undetected. The swallowed error (via `n, _`) is also a code-quality violation. Fix: use `io.ReadAll(resp.Body)`.

**3. Hardcoded string match inside table-driven runner (return_to_test.go:137)**
`if tt.returnTo == "/path\nhttp://evil.com"` is a special case hardwired into the loop body, which defeats the purpose of the table-driven pattern. Any future newline variant added to the table would be silently skipped. Fix: add a `rawQuery string` field to the test struct and use it in the runner.

**4. Unnecessary reimplementation of strings.Contains (return_to_test.go:201–210)**
The `contains` helper re-implements stdlib `strings.Contains` with a naive O(n*m) scan and a misleading "to avoid importing strings" comment. Delete it; import `strings` and call `strings.Contains` directly.

---

**What is correct and should not change:**

- The layered validation logic in `isSafeReturnTo` is solid: prefix checks before `url.Parse`, CRLF guard, Scheme+Host emptiness check. No bypass vectors found for the primary open-redirect threat.
- Percent-encoded `%2F%2F` is correctly handled: `url.Query().Get` decodes it to `//evil.com`, which is then caught by the prefix check in `isSafeReturnTo`.
- The client-side `safeReturnTo` in both templates correctly mirrors the server-side rules using the browser `URL()` constructor with origin comparison. Server is the authority; client is defense-in-depth. Parity is acceptable.
- Architecture is correct: security helper lives in adapters/ alongside the handler it serves. No domain layer import violations. No new dependencies.
- `isSafeReturnTo` is unexported (lowercase) — correct since it is an internal helper; no godoc comment needed.
- 24-case `TestIsSafeReturnTo` coverage is comprehensive for the primary attack surface.

Three of the four issues above (#2–4) are in the test file only. Issue #1 is a functional correctness bug in the production code path that also causes an incorrect test expectation. All four are required fixes before approval.

tibtof:
Follow-up commit 9cd8a6c addresses all 5 review points. Summary for re-reviewers:

**Reviewer issue #1 — `returnToQuery` production bug (handler.go)**
`return "?return_to=" + url.QueryEscape(v)` — the raw value is now percent-encoded before insertion into the query string. A safe path like `/dashboard?tab=settings&modal=open` becomes `?return_to=%2Fdashboard%3Ftab%3Dsettings%26modal%3Dopen`, eliminating the `html/template &amp;` truncation on the receiving page. The `net/url` import was already present (used by `url.Parse` in `isSafeReturnTo`).

**Reviewer issue #2 — 32KB read truncation (return_to_test.go)**
`var buf [32768]byte; n, _ := resp.Body.Read(buf[:])` replaced with `io.ReadAll(resp.Body)` — reads the full response and surfaces any read error explicitly.

**Reviewer issue #3 — hardcoded string match in table loop (return_to_test.go)**
Added `rawQuery string` field to the test struct. The runner now uses `if tt.rawQuery != ""` to set `r.URL.RawQuery` verbatim. The newline injection case sets `rawQuery: "return_to=/path%0ahttp://evil.com"` in the table entry — no more match on `tt.returnTo == "/path\nhttp://evil.com"` inside the loop.

**Reviewer issue #4 — `contains()` reimplementing stdlib (return_to_test.go)**
Local `contains()` function deleted. `strings.Contains` used directly. `strings` import added.

**Writer issue — missing CHANGELOG entry**
`### Security` subsection added under `## [Unreleased]` with the required entry describing the `return_to` validation on both server-side (`isSafeReturnTo`) and client-side (`safeReturnTo()`).

`make check` (gofmt + golangci-lint + go build + go test -race) passes on the updated branch. Pre-push hook confirmed clean.

tibtof:
Documentation review round 2: CHANGELOG entry verified, no regressions.

**CHANGELOG entry** (commit 9cd8a6c) — correct:
- Under `[Unreleased]` > `### Security`
- Format: `- **fix(#1267): ...**` matches pattern
- Both validation layers named: server-side `isSafeReturnTo`, client-side `safeReturnTo()`
- Rejected cases documented: external URLs, protocol-relative URLs, backslash variants, CRLF-injected paths
- Affected routes listed: `/auth/login`, `/auth/registration`
- Concise form vs round-1 verbose template — accurate and complete at changelog abstraction level, acceptable

**Other surfaces** — no regressions. Files changed on this PR: `CHANGELOG.md`, `internal/adapters/authui/handler.go`, `internal/adapters/authui/return_to_test.go`, `internal/adapters/authui/templates/login.html`, `internal/adapters/authui/templates/registration.html`. No user-facing doc surface was modified.

**url.QueryEscape impact** — no doc update needed. The change is an internal implementation detail; no user-facing doc shows the literal output of `returnToQuery`. All previously confirmed-clean surfaces remain clean.

**PR body** — adequate. Round-2 commit message documents the reviewer and writer fixes.

VERDICT: APPROVE

tibtof:
**Reviewer Agent: APPROVED** (round 2)

All 4 round-1 issues verified resolved in commit `9cd8a6c`. CI fully green.

**Round-1 items — verified:**

1. **Production fix — `url.QueryEscape`**: `handler.go` line ~284 now returns `"?return_to=" + url.QueryEscape(v)`. `net/url` import present. `wantQuery` expectations in `TestReturnToQuery_ServerSide` updated to escaped form (`%2F`, `%2Fdashboard`, `%2Fsearch%3Fq%3Dhello`). Confirmed.

2. **Test fix — `io.ReadAll`**: `TestReturnToQuery_ExternalURLDroppedFromPage` uses `io.ReadAll(resp.Body)` with error surfaced via `t.Fatalf`. No 32KB buffer read. Confirmed.

3. **Test fix — `rawQuery` field**: Test struct has `rawQuery string` field. Runner branches on `if tt.rawQuery != ""`. No hardcoded string match on the newline payload path. Newline injection case supplies `rawQuery: "return_to=/path%0ahttp://evil.com"`. Confirmed.

4. **Test fix — `contains()` deleted**: Local helper gone. `strings.Contains` used directly. `strings` import present. Confirmed.

5. **CHANGELOG**: `### Security` subsection present under `[Unreleased]` with the `fix(#1267)` entry. Confirmed.

6. **Scope**: diff touches only `CHANGELOG.md`, `internal/adapters/authui/handler.go`, `internal/adapters/authui/return_to_test.go`, and the two HTML templates. No unintended changes.

**CI rollup on `9cd8a6c`**: Build & Test (SUCCESS), Lint (SUCCESS), Integration Tests (SUCCESS), CodeQL (SUCCESS), Semgrep SAST (SUCCESS), Trivy Vulnerability Scan (SUCCESS), Go Vulnerability Check (SUCCESS). All checks green.

VERDICT: APPROVE

Issue #1301, opened 2026-05-05
audit(architecture): env Resolver (ADR-102) bypassed by bundle and validate — sibling implementations live on


tibtof:
## Finding

ADR-102 introduced `internal/app/env/resolver.go` as the canonical entry point for `--env <name>` config resolution. Today only two consumers route through it:

- `internal/cli/cmd/probe.go:126` — uses `envapp.NewFileResolver`
- `internal/cli/cmd/doctor.go:102` — uses `envapp.NewFileResolver`

The other consumers ADR-102 explicitly listed — `bundle` and `validate` — still roll their own logic:

- `internal/cli/cmd/bundle.go:499-556` — `prodConfigPathForEnv` (file existence probe), `readProdTLSDomain` (bespoke `yaml.Unmarshal` of `tls.domain` only), `readProdDeployHost` (bespoke `yaml.Unmarshal` of `deploy.host` only).
- `internal/cli/cmd/validate.go:189` — direct `bundleapp.LoadMergedConfig(absCheck, prodPath)` call. `discoverProdOverride` at `validate.go:257-273` re-implements file-existence resolution.

The Resolver also exposes `Resolved.OverridePath` and `Resolved.Cfg` as a clean merged-config result; today bundle.go gets the merged config from `bundle.Service` internally, then *separately* re-reads two specific fields (`tls.domain`, `deploy.host`) directly from the production override file because the merged cfg in the calling stack is base-only. That is the symptom the env Resolver was designed to remove.

## Why it matters

ADR-102 §"Reusability for future verbs" calls out bundle and validate by name as upcoming consumers. The bespoke readers in bundle.go are also fragile: they will silently miss a third field (e.g. `tls.acme_email` already lives in production overrides today) the next time the contract grows, and the reader-per-field pattern duplicates `os.ReadFile` + `yaml.Unmarshal` plumbing per field. The architect comment on PR #1244 explicitly accepted these as YAGNI for two fields, but with #1245 already adding deploy.host as the third surface that crosses the boundary, the threshold has been hit.

## Suggested fix

[...]

PR #1318, opened 2026-05-05
fix(#1269): reject env names containing path-traversal sequences

### LINKED ISSUE #1269: audit(security): env resolver path traversal — crafted --env name reads arbitrary YAML files outside project root
## Finding

**Severity**: high
**OWASP Category**: A01:2021 Broken Access Control / A05:2021 Security Misconfiguration

### Description

`internal/app/env/resolver.go` builds the override config path via string concatenation followed by `filepath.Join`:

```go
overridePath := filepath.Join(root, "vibewarden."+name+".yaml")
```

When `name` contains `../` sequences (e.g. `"../../../etc/secret"`), `filepath.Join` resolves the resulting dotted prefix `"vibewarden.../../../etc/secret.yaml"` by treating `"vibewarden.."` as a directory name and then processing the remaining `../../` segments. The result escapes the project root.

**Concrete example:**

```
root    = /home/user/project
name    = "../../../etc/secret"
joined  = /home/user/project/vibewarden.../../../etc/secret.yaml
cleaned = /home/user/etc/secret.yaml   ← outside root
```

Three or more `../` segments in `name` escape the project root. The file is then loaded by `bundleapp.LoadMergedConfig`, which unmarshals it as a VibeWarden config. Any error is printed to the terminal (leaking partial content via error messages).

### Affected CLI verbs

Both commands that call `env.FileResolver.Resolve`:
- `vibew probe --env <name>`
- `vibew doctor` (when run after `vibew dev` is running, via its own resolver call)

### Reproduction

```bash
# Assuming project root is three levels deep, e.g. /home/user/work/myapp
vibew probe --env '../../../etc/passwd'
# Attempts to load /etc/passwd.yaml — if absent, error message leaks that the file was sought.
# If a .yaml-suffixed file exists outside the root, its content is parsed and config errors
# may disclose portions of its content.
```

### Why it matters

An attacker with local shell access who can invoke `vibew` CLI commands could read YAML-parseable files outside the project directory. More importantly, if an attacker controls the `--env` argument (e.g. via a CI script that passes user-supplied input), they can exfiltrate files.

### Suggested fix

Validate `name` in `FileResolver.Resolve` before path construction. Reject names that contain `/`, `\`, or `.`:

```go
if strings.ContainsAny(name, "/\\.") {
    return Resolved{}, fmt.Errorf("env name %q contains disallowed characters", name)
}
```

Or, after joining, verify the result is inside `root`:

```go
if !strings.HasPrefix(overridePath+string(os.PathSeparator), filepath.Clean(root)+string(os.PathSeparator)) {
    return Resolved{}, fmt.Errorf("env name %q escapes project root", name)
}
```

### Source

- `internal/app/env/resolver.go` — `FileResolver.Resolve` line 113
- `internal/cli/cmd/probe.go` — passes `--env` arg to resolver without validation (line 127)
- `internal/cli/cmd/doctor.go` — also calls resolver without validation (line 102)

tibtof:
Closes #1269

## Summary

- `FileResolver.Resolve` now validates the env name against an allowlist regexp (`^[a-zA-Z0-9_-]+$`) before building any filesystem path. Names containing `/`, `\\`, `..`, `.`, NUL bytes, spaces, or any character outside `[a-zA-Z0-9_-]` are rejected immediately with the new `ErrInvalidEnvName` sentinel.
- Defense-in-depth: after constructing the override path, a `filepath.Rel` check verifies the resolved path stays inside the project root. This catches edge cases the allowlist alone cannot see (e.g. future symlink scenarios).
- Callers (`vibew probe --env`, `vibew doctor --preflight`) pick up the fix for free — they both go through `FileResolver.Resolve`.
- `ErrInvalidEnvName` is distinct from `ErrOverrideConfigMissing` / `ErrBaseConfigMissing` so callers can give precise error messages.

Scope: validation only, no structural changes. #1301 (bundle/validate callers) is a separate follow-up as noted in the issue.

## Test plan

- `TestValidateEnvName_RejectsTraversalNames` — table-driven, 11 bad names (`../etc/passwd`, `../../foo`, `foo/bar`, `..`, `.`, `.hidden`, `foo\bar`, NUL byte, space, `@`, `:`), all must return `ErrInvalidEnvName`.
- `TestValidateEnvName_AcceptsLegitimateNames` — 6 well-formed names (`prod`, `staging-eu`, `local_dev`, `StageUS`, `env01`, `My-Env_123`) must pass through.
- `TestProbeCmd_EnvPathTraversal` — end-to-end CLI integration test: drives `vibew probe --env <bad-name>` for 9 traversal inputs from a clean tmpdir; each must return a non-zero exit code without touching any file outside the project root.


internal/app/env/resolver.go
```diff
@@ -111,6 +131,16 @@ func (r *FileResolver) Resolve(name string) (Resolved, error) {
 	}
 
 	overridePath := filepath.Join(root, "vibewarden."+name+".yaml")
+
+	// Defense-in-depth: verify the resolved override path has not escaped the
+	// project root. filepath.Rel returns a path starting with ".." when the
+	// target is outside the base. This guard catches any edge cases that slip
```
tibtof: **Security gap — symlink escape not blocked by `filepath.Rel`**

The `filepath.Rel` check on line 137 operates on the _constructed_ path `filepath.Join(root, "vibewarden."+name+".yaml")`. Because `name` is guaranteed to be `[a-zA-Z0-9_-]+` by this point, that constructed path is always inside `root` — so the `Rel` check is vacuously true and adds no real defense for this case.

The actual escape vector it was meant to guard is a **symlink**: if the user has a file named `vibewarden.prod.yaml` that is a symlink to `/etc/passwd` (or any path outside the project root), the allowlist passes (`prod` is valid), `filepath.Rel` sees `vibewarden.prod.yaml` (inside root), and then `bundleapp.LoadMergedConfig` reads through the symlink to the target.

`filepath.EvalSymlinks` is needed after the path is constructed and before the `Rel` check:

```go
overridePath := filepath.Join(root, "vibewarden."+name+".yaml")

// Resolve symlinks before the containment check so a symlink pointing
// outside the project root is caught here rather than followed silently.
realPath, err := filepath.EvalSymlinks(overridePath)
if err != nil && !os.IsNotExist(err) {
    return Resolved{}, fmt.Errorf("resolving override path: %w", err)
}
if err == nil { // file exists — check containment of the real path
    rel, relErr := filepath.Rel(root, realPath)
    if relErr != nil || strings.HasPrefix(rel, "..") {
        return Resolved{}, fmt.Errorf("%w: resolved path escapes project root", ErrInvalidEnvName)
    }
}
```

When the file does not yet exist (`os.IsNotExist`), `EvalSymlinks` errors and we fall through to the existing `os.Stat` → `ErrOverrideConfigMissing` path — no behaviour change for the missing-file case.

Why it matters: this is a security fix for a path-traversal audit finding. The allowlist correctly prevents `../` and friends in the _name_, but a pre-existing symlink with a legitimate name is the classic second-order traversal. The codebase already calls `filepath.EvalSymlinks` in `internal/app/bundle/input_digest.go:174,228` and `internal/app/ops/image_identity.go:61` for exactly this reason.

tibtof (commented):
Documentation review — VERDICT: REQUEST CHANGES — 3 issues found, 1 blocking.

---

## Issue 1 (BLOCKING): CHANGELOG.md missing `### Security` entry

`## [Unreleased]` is empty. This is a security fix (path-traversal rejection); it requires a `### Security` entry per the pattern established in #1317.

**Required addition** immediately after `## [Unreleased]`:

```markdown
### Security

- **fix(#1269): `vibew probe --env` and `vibew doctor --preflight` now reject env names containing path-traversal sequences.** `FileResolver.Resolve` validates the name against the allowlist `^[a-zA-Z0-9_-]+$` before constructing any filesystem path. Names containing `/`, `\`, `..`, `.`, NUL bytes, spaces, or any other character outside the allowlist are rejected with `ErrInvalidEnvName`. Defense-in-depth `filepath.Rel` check catches any edge case (e.g. symlinks) that the allowlist cannot see. Callers `vibew probe --env` and `vibew doctor --preflight` both go through `FileResolver.Resolve` and pick up the fix without code changes.
```

---

## Issue 2 (non-blocking, recommended): `--env` flag help text does not document the character allowlist

`probe.go` line 85:
```
"environment name; loads vibewarden.<name>.yaml and probes tls.domain"
```

When a user passes an invalid name they get `resolving env "../etc/passwd": invalid env name: must match [a-zA-Z0-9_-]+: got "../etc/passwd"`. The error text is adequate, but the flag description gives no hint. An AI agent reading `vibew probe --help` will not know the constraint before constructing the flag value.

Suggested replacement:
```
"environment name (alphanumerics, hyphens, underscores only — e.g. production, staging-eu); loads vibewarden.<name>.yaml and probes tls.domain"
```

Same change applies to `doctor.go` line 203-204 (`--preflight` flag).

---

## Issue 3 (non-blocking, recommended): `llms-full.txt` `vibew probe --env <name>` entry does not mention the allowlist

The entry at the `vibew probe --env <name>` row describes the flag's behaviour but says nothing about which characters are valid in `<name>`. An AI agent constructing a name from a project path or environment variable could silently produce an invalid name and get an opaque error.

Suggested addition to the end of that table cell:
```
Env name must match `[a-zA-Z0-9_-]+` (alphanumerics, hyphens, underscores). Names containing slashes, dots, spaces, or other characters are rejected before any filesystem access with a descriptive error.
```

---

## Surfaces checked

| Surface | Status |
|---|---|
| `CHANGELOG.md` `[Unreleased]` `### Security` | Missing — blocking |
| `internal/cli/cmd/probe.go` `--env` flag text | No allowlist hint — non-blocking |
| `internal/cli/cmd/doctor.go` `--preflight` flag text | No allowlist hint — non-blocking |
| `llms-full.txt` `vibew probe --env <name>` entry | No allowlist mention — non-blocking |
| `vibewarden.reference.yaml` | Not applicable (CLI flag, not config field) — clean |
| `docs/examples/AGENTS-VIBEWARDEN.md` | All examples use `production`/`staging` — clean |
| `docs/getting-started.md` | All examples valid — clean |
| `docs/troubleshooting.md` | All examples valid — clean |
| `docs/agent-kickoff.md` | All examples valid — clean |
| `README.md` | Uses `<name>` template form — clean |
| PR body | Clear summary, test plan present — acceptable |

internal/app/env/resolver_test.go
```diff
@@ -154,3 +154,79 @@ func TestFileResolver_BasePath_IsAbsolute(t *testing.T) {
...
+			if err != nil {
+				t.Errorf("Resolve(%q) unexpected error: %v", tt.envName, err)
+				return
+			}
+			if resolved.EnvName != tt.envName {
+				t.Errorf("EnvName = %q, want %q", resolved.EnvName, tt.envName)
+			}
+		})
+	}
+}
```
tibtof, line 268: **Missing test for symlink escape vector**

The test suite covers the allowlist (11 traversal name cases) and the `filepath.Rel` check indirectly through the allowlist (which makes the `Rel` check vacuous as noted in the other comment). There is no test that creates a _legitimate_ env name whose corresponding file is a symlink pointing outside the project root.

Add a test case to `TestValidateEnvName_RejectsTraversalNames` (or a dedicated `TestFileResolver_Resolve_SymlinkEscape`):

```go
func TestFileResolver_Resolve_SymlinkEscape(t *testing.T) {
    dir := t.TempDir()
    writeFile(t, filepath.Join(dir, "vibewarden.yaml"), minimalBase(8443))

    // canary file outside the project root
    outside := t.TempDir()
    canary := filepath.Join(outside, "secret.txt")
    writeFile(t, canary, "secret")

    // symlink inside root with a valid env name pointing outside root
    link := filepath.Join(dir, "vibewarden.prod.yaml")
    if err := os.Symlink(canary, link); err != nil {
        t.Skip("symlinks not supported:", err)
    }

    r := envpkg.NewFileResolver(dir)
    _, err := r.Resolve("prod")
    if err == nil {
        t.Fatal("Resolve(\"prod\") expected error for symlink escape, got nil")
    }
    if !errors.Is(err, envpkg.ErrInvalidEnvName) {
        t.Errorf("expected ErrInvalidEnvName, got: %v", err)
    }
}
```

This test will currently fail (the symlink is followed silently). It should pass once `filepath.EvalSymlinks` is added per the companion comment on the containment check.

tibtof:

VERDICT: CHANGES REQUESTED

Two blocking issues. Two non-blocking observations.

---

### Blocking 1 — Symlink escape: `filepath.Rel` check is vacuous

**File:** `internal/app/env/resolver.go` (inline comment posted)

By the time the `filepath.Rel` guard runs, `name` has already been validated to match `^[a-zA-Z0-9_-]+$`. The constructed `overridePath` is therefore always `<root>/vibewarden.<safe-name>.yaml` — it can never contain `../` — so `filepath.Rel` always returns a path without a `..` prefix. The guard is dead code for the validated-name path.

The real escape is a **symlink**: `vibewarden.prod.yaml` is a valid name, passes the allowlist, and the `Rel` check sees a path inside root. But `bundleapp.LoadMergedConfig` then reads through the symlink to wherever it points (e.g., `/etc/passwd`, `~/.ssh/id_rsa`).

Fix: call `filepath.EvalSymlinks(overridePath)` after constructing the path, and run the `filepath.Rel` containment check on the _real_ path. Full suggested code in the inline comment. The codebase already uses `EvalSymlinks` for this exact purpose in `internal/app/bundle/input_digest.go:174,228` and `internal/app/ops/image_identity.go:61`.

---

### Blocking 2 — CHANGELOG entry missing

`CHANGELOG.md` → `[Unreleased]` section is empty. Per project standards every security fix must be documented under `### Security` in `[Unreleased]`. This is a high-priority audit finding; it must be chronicled.

Required entry:

```markdown
### Security

- **fix(#1269): `vibew probe --env` and `vibew doctor --preflight` now reject env names containing path-traversal sequences.** `FileResolver.Resolve` validates the name against allowlist `^[a-zA-Z0-9_-]+$` (blocking `../`, slashes, NUL bytes, dots) before building any filesystem path. A second layer calls `filepath.EvalSymlinks` + `filepath.Rel` to reject symlinks that point outside the project root. New sentinel: `ErrInvalidEnvName`.
```

---

### Non-blocking 1 — Missing symlink test

No test exercises the symlink escape path. Inline comment posted on `resolver_test.go` with a complete `TestFileResolver_Resolve_SymlinkEscape` snippet. This test should be added alongside the `EvalSymlinks` fix so the fix is regression-protected.

---

### Non-blocking 2 — `doctor.go` does not surface `ErrInvalidEnvName` distinctly

`doctor.go:104-112` handles `ErrOverrideConfigMissing` and `ErrBaseConfigMissing` explicitly but lets `ErrInvalidEnvName` fall through to `default: fmt.Errorf("loading env config: %w", resolveErr)`. The error message is accurate but not actionable — the user sees no hint that they passed a bad name. `probe.go:129-135` has the same gap. Consider adding a case:

```go
case errors.Is(resolveErr, envapp.ErrInvalidEnvName):
    return fmt.Errorf("invalid env name %q: only letters, digits, hyphens, and underscores are allowed: %w", preflightEnv, resolveErr)
```

Not blocking this cycle; can be done in the same commit or a follow-up.

---

### What is good

- Allowlist `^[a-zA-Z0-9_-]+$` is correct and strict. Empty string, NUL bytes, dots, slashes, backslashes, at-signs, colons, and all Unicode high-bytes are correctly rejected.
- The NUL-byte explicit check before the regex is belt-and-suspenders good practice (regex engines differ in NUL handling).
- Both named callers (`probe.go:127`, `doctor.go:103`) route through `FileResolver.Resolve` and therefore through the validator — no bypass on those paths.
- `ErrInvalidEnvName` is exported with a godoc comment.
- Negative tests correctly use `errors.Is(err, envpkg.ErrInvalidEnvName)` not just `err != nil`.
- Table-driven test structure is clean and covers 11 traversal cases.
- Hexagonal boundaries respected: validation lives in `internal/app/env/`, not in adapters.
- No new dependencies introduced.

tibtof:
Follow-up commit `867fcca` addresses all 6 review points. Summary:

**Reviewer Blocking 1 — symlink escape (CRITICAL)**
`resolver.go`: `filepath.EvalSymlinks` is now called on both the project root and the override file path (when it exists) before the `filepath.Rel` containment check. The root is resolved to its real path to handle platform-level symlinks (e.g. macOS `/var` → `/private/var`). When the override file exists, `EvalSymlinks` resolves it; the containment check then runs on the real path. A symlink pointing outside the root (e.g. `vibewarden.prod.yaml` → `/etc/passwd`) causes the check to fail with `ErrInvalidEnvName`. When the file is absent (fine — missing file means no read), containment is still checked using a path re-joined from `realRoot` so both sides share the same resolution.

**Reviewer Non-Blocking 1 — symlink escape test**
`resolver_test.go`: `TestFileResolver_Resolve_SymlinkEscape` added. Creates a sentinel file in an outside tmpdir, symlinks it into the project root as `vibewarden.prod.yaml`, calls `Resolve("prod")`, asserts `ErrInvalidEnvName`.

**Reviewer Non-Blocking 2 — surface `ErrInvalidEnvName` distinctly**
`probe.go` and `doctor.go`: added `errors.Is(err, envapp.ErrInvalidEnvName)` case that returns a clear actionable error: `"invalid --env value %q: env name must match [a-zA-Z0-9_-]+ (alphanumerics, hyphens, underscores)"`. Generic wrap retained as fallback in the `default` branch.

**Writer Blocking — CHANGELOG**
`CHANGELOG.md`: `### Security` subsection added under `[Unreleased]` with the required entry describing the name-allowlist fix and the `EvalSymlinks` defense-in-depth layer.

**Writer Non-Blocking — CLI help text**
`probe.go` line 85: `--env` usage string updated to include `(alphanumerics, hyphens, underscores only — e.g. production, staging-eu)`.
`doctor.go` flag: `--preflight` usage string updated with the same allowlist hint and examples.

**Writer Non-Blocking — llms-full.txt**
`llms-full.txt` line 1373: appended `Env name must match [a-zA-Z0-9_-]+ (alphanumerics, hyphens, underscores). Names containing slashes, dots, spaces, or other characters are rejected before any filesystem access with a descriptive error.` to the `vibew probe --env <name>` table row.

`make check` passes (golangci-lint, gofmt, goimports, go test -race ./...) — all 14 env-package tests green including the new symlink test.

tibtof:
Documentation review round 2 — VERDICT: APPROVE. All three issues from round 1 addressed.

**Issue 1 (blocker) — RESOLVED.** `CHANGELOG.md` `## [Unreleased]` now contains a `### Security` heading (line 15) with the #1269 entry. No duplicate heading. Entry documents: `^[a-zA-Z0-9_-]+$` allowlist, `ErrInvalidEnvName`, `filepath.EvalSymlinks` defense-in-depth, and both callers (`vibew probe --env`, `vibew doctor --preflight`). Ordering is correct — #1317's entry is in the already-merged versioned section, not duplicated here.

**Issue 2 (non-blocking) — RESOLVED.** `probe.go` flag text now reads: `environment name (alphanumerics, hyphens, underscores only — e.g. production, staging-eu); loads vibewarden.<name>.yaml and probes tls.domain`. `doctor.go` flag text updated identically.

**Issue 3 (non-blocking) — RESOLVED.** `llms-full.txt` `vibew probe --env <name>` row now appends: `Env name must match [a-zA-Z0-9_-]+ (alphanumerics, hyphens, underscores). Names containing slashes, dots, spaces, or other characters are rejected before any filesystem access with a descriptive error.`

Previously-clean surfaces remain clean.

tibtof:
**Reviewer Agent: APPROVED**

Round 2 re-review of commit `867fcca`. All four items from round 1 are resolved. CI (CodeQL) passed at 2026-05-05T16:38:49Z.

**Blocking 1 — Symlink escape: RESOLVED.**
`resolver.go` now calls `filepath.EvalSymlinks` on both the project root and the override file path before the `filepath.Rel` containment check. The root is resolved first to handle macOS `/var` → `/private/var` platform symlinks, eliminating false positives in `t.TempDir()` on CI. The absent-file case is correctly handled: `os.Stat` gates the `EvalSymlinks` call so missing files fall through to the existing `ErrOverrideConfigMissing` path without a spurious escape error. The containment check runs on real paths on both sides of `filepath.Rel`. Pattern matches `internal/app/bundle/input_digest.go:174,228` and `internal/app/ops/image_identity.go:61` exactly.

**Blocking 2 — CHANGELOG: RESOLVED.**
`### Security` subsection present under `## [Unreleased]` with the required `fix(#1269)` entry covering both the allowlist and the `EvalSymlinks` defense-in-depth layer.

**Non-blocking 1 — Symlink test: RESOLVED.**
`TestFileResolver_Resolve_SymlinkEscape` creates sentinel content in an outside `t.TempDir()`, symlinks into project root as `vibewarden.prod.yaml`, calls `Resolve("prod")`, and asserts `errors.Is(err, ErrInvalidEnvName)`. The two-`TempDir()` approach is Mac CI-safe: both dirs resolve under `/private/var/folders/...` but as siblings, so containment check correctly rejects. Sentinel content assertion by error presence is sufficient.

**Non-blocking 2 — CLI error surfacing: RESOLVED.**
Both `probe.go` and `doctor.go` now have `errors.Is(err, envapp.ErrInvalidEnvName)` as the first `switch` case, returning actionable messages naming the allowlist and correct flag names (`--env` vs `--preflight`). Generic `default` wrap retained for other errors.

Additional: `probe_test.go` adds `TestProbeCmd_EnvPathTraversal` as an end-to-end CLI integration test for 9 bad names. `llms-full.txt` updated. `--env` and `--preflight` flag help text updated.

PR #1319, opened 2026-05-05
fix(#1264): strip incoming X-User-* headers + fix public-path prefix bypass

tibtof:
Closes #1264

## Summary

- **Part A — wildcard header strip**: `buildUserHeaderStripHandler` now emits a single Caddy regex delete pattern `~^X-User-` instead of a hardcoded 4-entry list. Both `AuthHandler.ServeHTTP` and `JWTBearerHandler.ServeHTTP` also call `stripXUserHeaders()` as a Go-layer defence-in-depth that fires before any session validation.
- **Part B — public-path prefix boundary**: `isPublicPath` on both handlers now requires either an exact match on the prefix or a trailing slash before continuing the match. A pattern `/auth/*` no longer matches `/auth-evil` or `/authentic`.

## Test plan

- `TestBuildUserHeaderStripHandler_WildcardPattern` — verifies delete pattern is `"~^X-User-"`, single entry
- `TestBuildCaddyConfig_UserHeaderStripDeletesCorrectHeaders` — updated to expect the wildcard pattern
- `TestAuthHandler_isPublicPath_PrefixBoundaryBypass` — sibling paths `/auth-evil`, `/authentic` do not match `/auth/*`
- `TestStripXUserHeaders` — table-driven: all X-User-* variants stripped, unrelated headers kept
- `TestAuthHandler_ServeHTTP_StripsForggedXUserHeaders` — forged values never reach upstream
- `TestJWTBearerHandler_isPublicPath_PrefixBoundaryBypass` — same boundary check for JWT handler
- `TestJWTBearerHandler_ServeHTTP_StripsForggedXUserHeaders` — forged X-User-* stripped on public paths
- `TestJWTBearerHandler_ServeHTTP_NoBearer`, `TestJWTBearerHandler_ServeHTTP_PublicPathBypass` — basic JWT handler coverage
- All 100+ packages: `go test -race ./...` green, `golangci-lint run ./...` 0 issues

### DIFF
--- a/internal/adapters/caddy/config_handlers.go
+++ b/internal/adapters/caddy/config_handlers.go
@@ -18,15 +18,16 @@ var defaultCompressionAlgorithms = []string{"zstd", "gzip"}
 // impersonating an authenticated user by injecting them directly. VibeWarden
 // re-injects them only after a valid session has been verified.
 //
-// The tilde prefix ("~^X-User-") instructs Caddy to treat the value as a regular
-// expression, matching all headers whose names start with "X-User-". This ensures
-// that headers added in the future (e.g. X-User-Name) are also stripped without
-// requiring changes to this function.
+// The pattern "x-user-*" uses Caddy's suffix-wildcard glob syntax (end with "*"
+// means "prefix match"). Caddy lowercases both the pattern and the existing header
+// name before comparing, so "x-user-*" matches X-User-Id, X-User-Email,
+// X-User-Name, and any future X-User-* additions without requiring changes here.
+// See: caddyserver/caddy/v2/modules/caddyhttp/headers/headers.go — Delete field.
 func buildUserHeaderStripHandler() map[string]any {
 	return map[string]any{
 		"handler": "headers",
 		"request": map[string]any{
-			"delete": []string{"~^X-User-"},
+			"delete": []string{"x-user-*"},
 		},
 	}
 }
--- a/internal/adapters/caddy/config_handlers.go
+++ b/internal/adapters/caddy/config_handlers.go
@@ -10,18 +10,23 @@ import (
 // CompressionConfig.Algorithms is empty.
 var defaultCompressionAlgorithms = []string{"zstd", "gzip"}
 
-// buildUserHeaderStripHandler creates a Caddy headers handler that deletes the
-// X-User-Id, X-User-Email, X-User-Verified, and X-User-Role request headers.
+// buildUserHeaderStripHandler creates a Caddy headers handler that deletes all
+// X-User-* request headers.
 //
 // This handler must be placed as the very first handler in every route's chain.
 // Removing these headers on every inbound request prevents a client from
 // impersonating an authenticated user by injecting them directly. VibeWarden
 // re-injects them only after a valid session has been verified.
+//
+// The tilde prefix ("~^X-User-") instructs Caddy to treat the value as a regular
+// expression, matching all headers whose names start with "X-User-". This ensures
+// that headers added in the future (e.g. X-User-Name) are also stripped without
+// requiring changes to this function.
 func buildUserHeaderStripHandler() map[string]any {
 	return map[string]any{
 		"handler": "headers",
 		"request": map[string]any{
-			"delete": []string{"X-User-Id", "X-User-Email", "X-User-Verified", "X-User-Role"},
+			"delete": []string{"~^X-User-"},
 		},
 	}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This initial entry was written by hand to summarise the work leading up to v0.1.
 
 - **fix(#1267): authui — `return_to` parameter validates same-origin to close open-redirect vector.** Server-side `isSafeReturnTo` and client-side `safeReturnTo()` both reject external URLs, protocol-relative URLs, backslash variants, and CRLF-injected paths. Affected pages: `/auth/login`, `/auth/registration`.
 - **fix(#1269): reject env names containing path-traversal sequences in `vibew probe --env` and `vibew doctor --preflight`.** New `validateEnvName` (`^[a-zA-Z0-9_-]+$`) returns `ErrInvalidEnvName`; defense-in-depth: resolved path is verified inside project root after `filepath.EvalSymlinks` to block symlink-escape via legitimately-named override files.
+- **fix(#1264): caddy auth handlers — strip ALL X-User-* headers from incoming requests + fix public-path prefix matching.** Closes identity-spoofing via forged X-User-* values in JWT mode (notably X-User-Name) and public-path bypass via prefix-sibling paths (e.g. /auth-evil matching /auth/*). Both Caddy-layer (config_handlers.go x-user-* glob) and Go-layer (stripXUserHeaders defense-in-depth) defenses now active.
 
 ## [v0.18.7] — 2026-05-05
 
--- a/docs/examples/AGENTS-VIBEWARDEN.md
+++ b/docs/examples/AGENTS-VIBEWARDEN.md
@@ -49,6 +49,10 @@ When auth is enabled, read user identity from these headers:
 - `X-User-Verified` — email verification status ("true"/"false")
 - `X-User-Role` — user role from Kratos identity traits ("user", "admin", or "moderator"; defaults to "user")
 
+The sidecar strips ALL incoming X-User-* headers from the client request
+before validation and re-injects only the verified headers listed above after
+a successful auth check. App code must not perform its own X-User-* stripping.
+
 Role-based path restrictions can be configured in vibewarden.yaml via
 `auth.role_paths` (only when `auth.mode: kratos`). Requests to restricted
 paths by users without the required role receive HTTP 403.
--- a/internal/adapters/caddy/auth_handler.go
+++ b/internal/adapters/caddy/auth_handler.go
@@ -129,11 +129,16 @@ func (h *AuthHandler) UnmarshalJSON(data []byte) error {
 }
 
 // isPublicPath checks if the request path matches any public path pattern.
+//
+// For wildcard patterns ending in "/*" the prefix boundary is enforced: a
+// pattern "/auth/*" matches "/auth", "/auth/", and "/auth/login", but NOT
+// "/auth-evil" or "/authentic". The check requires either an exact prefix
+// match or a match followed by a slash separator.
 func (h *AuthHandler) isPublicPath(reqPath string) bool {
 	for _, p := range h.Config.PublicPaths {
 		if strings.HasSuffix(p, "/*") {
 			prefix := strings.TrimSuffix(p, "/*")
-			if strings.HasPrefix(reqPath, prefix) {
+			if reqPath == prefix || strings.HasPrefix(reqPath, prefix+"/") {
 				return true
 			}
 		}
@@ -146,6 +151,18 @@ func (h *AuthHandler) isPublicPath(reqPath string) bool {
 		}
 	}
 	return false
+}
+
+// stripXUserHeaders removes all request headers whose name starts with
+// "X-User-". It must be called at the top of ServeHTTP before any session
+// validation so that a client cannot impersonate an authenticated user by
+// injecting these headers directly.
+func stripXUserHeaders(r *http.Request) {
+	for key := range r.Header {
+		if strings.HasPrefix(key, "X-User-") {
+			delete(r.Header, key)
+		}
+	}
 }
 
 // kratosWhoamiResponse mirrors the relevant fields from the Kratos
@@ -304,6 +321,11 @@ func (h *AuthHandler) matchRequiredRole(reqPath string) (string, bool) {
 
 // ServeHTTP implements caddyhttp.MiddlewareHandler.
 func (h *AuthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
+	// Strip all X-User-* headers from the incoming request before any session
+	// validation. This is the Go-layer defence-in-depth complement to the
+	// Caddy-layer wildcard delete in buildUserHeaderStripHandler.
+	stripXUserHeaders(r)
+
 	// Public paths: optional auth — check session if cookie present, set
 	// headers if valid, but never redirect or block the request.
 	if h.isPublicPath(r.URL.Path) {
--- a/internal/adapters/caddy/auth_handler_test.go
+++ b/internal/adapters/caddy/auth_handler_test.go
@@ -1418,3 +1418,191 @@ func TestMatchRequiredRole_NoRolePaths(t *testing.T) {
 		t.Error("matchRequiredRole() ok = true with empty RolePaths, want false")
 	}
 }
+
+// ---------------------------------------------------------------------------
+// Security regression tests — #1264
+// ---------------------------------------------------------------------------
+
+// TestAuthHandler_isPublicPath_PrefixBoundaryBypass verifies that a URL-space
+// sibling (e.g. "/auth-evil") is NOT matched by a wildcard public-path rule
+// (e.g. "/auth/*"). Fixes the bare-prefix check that treated strings.HasPrefix
+// as a path match.
+func TestAuthHandler_isPublicPath_PrefixBoundaryBypass(t *testing.T) {
+	h := &AuthHandler{Config: AuthHandlerConfig{
+		PublicPaths: []string{"/auth/*", "/public/*"},
+	}}
+
+	tests := []struct {
+		path    string
+		want    bool
+		comment string
+	}{
+		{"/auth/login", true, "sub-path of /auth/* must match"},
+		{"/auth/", true, "trailing slash of /auth/* must match"},
+		{"/auth", true, "exact prefix /auth must match"},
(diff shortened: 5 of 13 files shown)


tibtof:
## Documentation review — Writer Agent

VERDICT: request-changes (bot-author block prevents `gh pr review --request-changes`; findings recorded here per process)

---

### [REQUIRED] CHANGELOG.md — Security entry missing

**Surface**: `CHANGELOG.md`, `[Unreleased]` > `### Security`

The `[Unreleased]` section has one entry (fix(#1269)). fix(#1264) is absent. Security fixes must have a CHANGELOG entry before merge so the release-note generator includes them in the published release page.

Add under `### Security`:

```
- **fix(#1264): caddy auth handlers — strip ALL X-User-* headers from incoming requests + fix public-path prefix matching.** Defense against identity-spoofing via forged X-User-* values in JWT mode (e.g. X-User-Name) and against route-bypass via prefix-sibling paths (e.g. /auth-evil matching /auth/*).
```

---

### [REQUIRED] llms-full.txt — X-User-* strip guarantee not documented

**Surface**: `llms-full.txt` around lines 344–368 (JWT `claims_to_headers` reference)

The table documents header *injection* but never states that incoming client-supplied `X-User-*` values are stripped unconditionally before auth. An AI agent reading this doc cannot infer the security contract and may implement redundant or incorrect app-side stripping — or worse, may not know the guarantee exists.

Add after the `claims_to_headers` block (before line 370):

```
  Security: VibeWarden strips ALL incoming X-User-* headers from the client
  request before any auth validation. Upstream app code must read only the
  headers VibeWarden injects after validation; any client-supplied X-User-*
  value is discarded unconditionally.
```

---

### [REQUIRED] llms-full.txt — public_paths boundary semantics not documented

**Surface**: `llms-full.txt` line 326

Current: `Glob patterns bypassing auth`

An AI agent configuring `/auth/*` has no signal that `/auth-evil` is excluded. This was the exact bug; the fix is invisible in the docs.

Replace description:

```
  auth.public_paths  list  []  Glob patterns bypassing auth. A trailing /*
                               matches the prefix and any sub-path but NOT
                               URL-space siblings: /auth/* matches /auth,
                               /auth/, and /auth/login but NOT /auth-evil.
```

---

### [REQUIRED] vibewarden.reference.yaml — public_paths comment incomplete after boundary fix

**Surface**: `vibewarden.reference.yaml` line 429

Current: `A trailing /* does prefix matching (e.g. /static/* matches /static/foo/bar).`

Correct but misleadingly incomplete — it implies pure string-prefix matching, which was the bug this PR fixes.

Extend to:

```
# A trailing /* does prefix matching (e.g. /static/* matches /static/foo/bar)
# but NOT URL-space siblings (/auth/* does NOT match /auth-evil or /authentic).
```

---

### [REQUIRED] docs/examples/AGENTS-VIBEWARDEN.md — X-User-* trust contract ambiguous

**Surface**: `docs/examples/AGENTS-VIBEWARDEN.md` §Sidecar-injected headers (lines 44–62) and the template that generates this file

The section tells agents to "read user identity from these headers" but never states that the sidecar strips client-supplied `X-User-*` before re-injecting them. App authors don't know whether they need to defend against header forgery in app code (they don't, but the doc doesn't say so).

Add after line 50:

```
The sidecar strips ALL incoming X-User-* headers from the client request
before validation and re-injects only the verified headers listed above after
a successful auth check. App code must not perform its own X-User-* stripping.
```

Mirror this addition in the vibew init template that generates AGENTS-VIBEWARDEN.md.

---

### Surfaces confirmed consistent — no changes needed

- `README.md` header table: lists only the 4 Kratos-mode injected headers; no claim that client headers pass through. No drift.
- PR body: summary clear, test plan names all regression tests, covers both Part A and Part B. Adequate.

internal/adapters/caddy/config_handlers.go
```diff
@@ -10,18 +10,23 @@ import (
...
+// The tilde prefix ("~^X-User-") instructs Caddy to treat the value as a regular
+// expression, matching all headers whose names start with "X-User-". This ensures
+// that headers added in the future (e.g. X-User-Name) are also stripped without
+// requiring changes to this function.
 func buildUserHeaderStripHandler() map[string]any {
 	return map[string]any{
 		"handler": "headers",
 		"request": map[string]any{
-			"delete": []string{"X-User-Id", "X-User-Email", "X-User-Verified", "X-User-Role"},
+			"delete": []string{"~^X-User-"},
```
tibtof: **Security bug — Caddy-layer header strip does nothing with this pattern**

The `"~^X-User-"` value in the `delete` array is not a recognized tilde-regex syntax in Caddy's `http.handlers.headers` module. The module's `Delete` field supports only glob-style wildcards (`*`-prefix, `*`-suffix, `*`-both-ends). The actual dispatch in `HeaderOps.ApplyTo` (Caddy v2.11.2, `modules/caddyhttp/headers/headers.go` lines 255–282) is:

```go
fieldName = strings.ToLower(repl.ReplaceKnown(fieldName, ""))
// ...
switch {
case strings.HasPrefix(fieldName, "*") && strings.HasSuffix(fieldName, "*"): ...
case strings.HasPrefix(fieldName, "*"): ...
case strings.HasSuffix(fieldName, "*"): ...
default:
    hdr.Del(fieldName)  // ← THIS branch fires for "~^x-user-"
}
```

With `"~^X-User-"` the pattern is lowercased to `"~^x-user-"`, none of the glob cases match, and `hdr.Del("~^x-user-")` is called — which deletes nothing (no header has that literal name). **The Caddy-layer strip is completely inoperative.** Only the Go-layer `stripXUserHeaders` call actually strips headers.

The correct Caddy prefix-wildcard is `"x-user-*"`. Caddy lowercases the pattern before comparing, and the suffix-wildcard case does:
```go
case strings.HasSuffix(fieldName, "*"):
    for existingField := range hdr {
        if strings.HasPrefix(strings.ToLower(existingField), fieldName[:len(fieldName)-1]) {
            delete(hdr, existingField)
        }
    }
```
So `"x-user-*"` correctly strips `X-User-Id`, `X-User-Email`, `X-User-Name`, and any future `X-User-*` headers.

**Fix required:**
```go
"delete": []string{"x-user-*"},
```

The existing tests (`TestBuildUserHeaderStripHandler_WildcardPattern` and `TestBuildCaddyConfig_UserHeaderStripDeletesCorrectHeaders`) only assert the structural value of the delete array — they do not exercise Caddy runtime behavior, so they pass green despite the strip being broken. A note on the tests: the Go-layer `stripXUserHeaders` provides real defense-in-depth and is correctly implemented, but the Caddy-layer fix must also be correct since it is the documented primary mechanism and the only protection for routes not going through `AuthHandler` or `JWTBearerHandler`.

tibtof:

---

## Summary

Two issues, one critical blocker.

### Issue 1 — BLOCKER: Caddy-layer header strip is broken (wrong pattern syntax)

Filed as inline comment on `internal/adapters/caddy/config_handlers.go` line 29.

The `"~^X-User-"` value in the Caddy `delete` array uses tilde-prefix notation, which is **not** a supported syntax in Caddy's `http.handlers.headers` module (Caddy v2.11.2). The module's `Delete` field supports only glob-style wildcards (`*`-prefix, `*`-suffix, `*`-both-ends). Verified directly in `/Users/tibtof/go/pkg/mod/github.com/caddyserver/caddy/v2@v2.11.2/modules/caddyhttp/headers/headers.go` lines 255–282.

The pattern is lowercased to `"~^x-user-"` and falls through to the `default` switch case, which calls `hdr.Del("~^x-user-")` — deleting a header that does not exist. The Caddy-layer strip does nothing.

**Correct pattern:** `"x-user-*"` (Caddy suffix-wildcard: lowercases both sides and checks `strings.HasPrefix(strings.ToLower(existingField), "x-user-")`).

The Go-layer `stripXUserHeaders(r)` in `ServeHTTP` is correctly implemented and does strip headers, but it only covers requests routed through `AuthHandler` or `JWTBearerHandler`. The Caddy-layer strip is the primary mechanism and must be fixed.

The two tests that check this (`TestBuildUserHeaderStripHandler_WildcardPattern` and `TestBuildCaddyConfig_UserHeaderStripDeletesCorrectHeaders`) only assert the structural value of the array — they do not execute the Caddy runtime, so they pass despite the strip being broken. A Caddy-layer integration test is needed to cover this, or at minimum the unit tests need updating to validate the correct glob pattern.

**Required fix:** change `"~^X-User-"` to `"x-user-*"` in `buildUserHeaderStripHandler()` and update the tests and godoc accordingly.

---

### Issue 2 — Missing `### Security` entry in CHANGELOG `[Unreleased]`

The `[Unreleased]` section is empty. This is a high-severity security fix that must be documented:

```markdown
## [Unreleased]

### Security

- **fix(#1264): strip incoming X-User-* headers + fix public-path prefix bypass.** Caddy-layer strip handler now uses `x-user-*` wildcard covering all `X-User-*` headers (previously `X-User-Name` in JWT bearer mode was reachable). Both `AuthHandler` and `JWTBearerHandler` call `stripXUserHeaders(r)` as first statement in `ServeHTTP` for Go-layer defence-in-depth. `isPublicPath` boundary fix prevents `/auth-evil` matching when `/auth/*` is public.
```

---

## What is correct and does not need changes

- **isPublicPath boundary fix**: correct in both `auth_handler.go` and `jwt_bearer_handler.go`. `reqPath == prefix || strings.HasPrefix(reqPath, prefix+"/")` is the right guard.
- **`stripXUserHeaders`**: correctly iterates canonical header map and uses `strings.HasPrefix(key, "X-User-")` — safe because Go's `net/http` canonicalizes all header names to Title-Case before storing in `r.Header`. No case-insensitivity gap exists at the Go layer.
- **`stripXUserHeaders` placement**: called as first statement in both `ServeHTTP` handlers before any other processing.
- **Test coverage for isPublicPath and stripXUserHeaders**: table-driven, covers the attack vectors from issue #1264, test names are descriptive.
- **`jwt_bearer_handler_test.go`**: new file is well-structured; 6 tests cover module metadata, interface guards, prefix boundary bypass, forged header stripping, no-bearer 401, and public path bypass.
- **Architecture**: all new code stays in `internal/adapters/caddy/`. No domain imports. `stripXUserHeaders` is package-private (lowercase), shared within the caddy package only.
- **No new dependencies introduced.**

tibtof:
## Follow-up commit: review fixes (c3617cd)

All 6 review points addressed in a single follow-up commit. No amend — new commit on top.

---

### Reviewer blocker — Caddy strip pattern corrected

**`internal/adapters/caddy/config_handlers.go`** — `buildUserHeaderStripHandler()`
`"~^X-User-"` → `"x-user-*"` (Caddy suffix-wildcard glob).

Root cause: Caddy's `headers.HeaderOps.ApplyTo` has three glob cases (`*`-prefix, `*`-suffix, `*`-both-ends). Tilde-prefix patterns fall through to `default: hdr.Del("~^x-user-")` which is a no-op — no header has that literal name. The suffix-wildcard `"x-user-*"` hits the `strings.HasSuffix(fieldName, "*")` case and correctly calls `strings.HasPrefix(strings.ToLower(existingField), "x-user-")` for every existing header.

---

### Behavioral test (reviewer request)

**`internal/adapters/caddy/config_handlers_behavior_test.go`** (new file)

`TestUserHeaderStripHandler_ActuallyStripsHeaders` — exercises the real `headers.HeaderOps.ApplyTo` runtime (not just JSON structure). Two sub-cases:
- `"x-user-*"`: asserts all four forged X-User-* headers are absent post-strip. PASSES.
- `"~^X-User-"`: asserts headers are still present (documents the no-op). PASSES as "the broken pattern does nothing."

`TestUserHeaderStripHandler_GlobMatchesAllVariants` — verifies 7 X-User-* variants (including X-User-Name, X-User-Custom, X-User-Org) are all stripped; X-Request-Id is not touched.

These tests fail immediately if the production code reverts to any tilde/regex pattern.

---

### Structural test updates

**`internal/adapters/caddy/config_build_test.go`** — `TestBuildUserHeaderStripHandler_WildcardPattern`: `wantPattern` constant `"~^X-User-"` → `"x-user-*"`; godoc explains why the tilde pattern is a no-op.

**`internal/adapters/caddy/config_test.go`** — `TestBuildCaddyConfig_UserHeaderStripDeletesCorrectHeaders`: same `wantPattern` update + godoc clarification.

---

### CHANGELOG

`CHANGELOG.md` `[Unreleased]` — `### Security` section added with the fix(#1264) entry.
Addresses: reviewer non-blocker + writer blocker.

---

### llms-full.txt

Two additions:
1. `auth.public_paths` field description extended with boundary semantics: "A trailing /* matches the prefix and any sub-path but NOT URL-space siblings: /auth/* matches /auth, /auth/, and /auth/login but NOT /auth-evil."
2. Security note added after `claims_to_headers` block: "VibeWarden strips ALL incoming X-User-* headers from the client request before any auth validation..."

Addresses: writer blockers (two required llms-full.txt changes).

---

### vibewarden.reference.yaml

`auth.public_paths` comment updated from "A trailing /* does prefix matching (e.g. /static/* matches /static/foo/bar)" to the path-segment-aware version with NOT-examples.
Addresses: writer blocker.

---

### AGENTS-VIBEWARDEN.md + template

Both `docs/examples/AGENTS-VIBEWARDEN.md` and `internal/cli/templates/agents/agents-vibewarden.md.tmpl` — paragraph added after the X-User-* header list: "The sidecar strips ALL incoming X-User-* headers from the client request before validation and re-injects only the verified headers listed above after a successful auth check. App code must not perform its own X-User-* stripping."
Addresses: writer blocker (both surfaces confirmed, template is the authoritative source).

---

`make check` clean. Pre-push hook ran the full suite. All 9 changed files staged explicitly — no agent scratch directories included.

tibtof:
VERDICT: APPROVE

Documentation review round 2 — all round-1 flags resolved.

1. **CHANGELOG.md** — `[Unreleased]` `### Security` entry present. Covers both fixes: X-User-* header stripping (including the previously-missing X-User-Name) and the public-path prefix-boundary bypass.

2. **llms-full.txt** — Two additions in the correct locations: (a) `auth.public_paths` row now documents the path-segment-aware boundary with /auth-evil NOT-example; (b) X-User-* strip guarantee paragraph added immediately after the `claims_to_headers` block.

3. **vibewarden.reference.yaml** — `public_paths` comment updated with boundary semantics and explicit NOT-examples (/auth-evil, /authentic).

4. **docs/examples/AGENTS-VIBEWARDEN.md** and **internal/cli/templates/agents/agents-vibewarden.md.tmpl** — Both carry the identical trust contract paragraph (strip before validation, re-inject after, app must not self-strip). Template and static example are in sync.

5. No regressions on previously clean surfaces.

Docs consistent with code changes.

tibtof:
**Reviewer Agent: APPROVED** (round 2)

All round-1 blockers and non-blockers resolved. Verified against commit `c3617cd`.

**Checklist verdict:**

1. **Caddy strip fix** — `buildUserHeaderStripHandler()` now emits `"x-user-*"` (suffix-wildcard glob). Confirmed in production code at `c3617cd`. The old hardcoded 4-entry list is gone.

2. **Structural tests updated** — `TestBuildUserHeaderStripHandler_WildcardPattern` (config_build_test.go) and `TestBuildCaddyConfig_UserHeaderStripDeletesCorrectHeaders` (config_test.go) both assert `wantPattern = "x-user-*"` with godoc explaining why the tilde pattern is a no-op.

3. **Behavioral test (config_handlers_behavior_test.go)** — New file verified present at `c3617cd`. Calls real `headers.HeaderOps.ApplyTo` with `caddy.NewReplacer()`. `TestUserHeaderStripHandler_ActuallyStripsHeaders` has two sub-cases: the `"x-user-*"` case asserts headers are absent (fails if code reverts), the `"~^X-User-"` sub-case asserts headers are still present (documents the no-op — would fail if Caddy ever added regex support, which is an acceptable false positive). `TestUserHeaderStripHandler_GlobMatchesAllVariants` exercises 7 X-User-* variants plus a non-User control header (`X-Request-Id`) that must survive. Both tests would fail if production code reverted to any tilde/regex pattern, and would also fail if the production code deleted nothing (control header check in GlobMatchesAllVariants confirms the ops object actually ran).

4. **CHANGELOG** — `[Unreleased] ### Security` entry present and correctly describes both the Caddy-layer and Go-layer fixes.

5. **Documentation surfaces** — All four confirmed at `c3617cd`:
   - `llms-full.txt`: `auth.public_paths` boundary semantics added; X-User-* strip security guarantee added after claims_to_headers block.
   - `vibewarden.reference.yaml`: `public_paths` comment updated with path-segment-aware NOT-examples.
   - `docs/examples/AGENTS-VIBEWARDEN.md`: strip trust contract paragraph present.
   - `internal/cli/templates/agents/agents-vibewarden.md.tmpl`: same paragraph present (template is authoritative source).

6. **Round-1 approved code intact** — `isPublicPath` boundary fix (`reqPath == prefix || strings.HasPrefix(reqPath, prefix+"/")`) present in both `AuthHandler` and `JWTBearerHandler`. `stripXUserHeaders` called as first statement in both `ServeHTTP` handlers.

7. **CI** — Full suite (Lint, Build & Test, Integration, Trivy, Semgrep, Go Vulnerability Check) passed on `789a5ce`. CodeQL Analyze passed on `c3617cd`. All green.

Open review thread from round 1 resolved.

PR #1342, opened 2026-05-06
fix(#1334): bundle — remove disable_mlock from generated openbao config

tibtof:
Closes #1334

## Summary

- Removes `disable_mlock = false` from `internal/config/templates/openbao-config.hcl.tmpl`. OpenBao 2.2.0+ treats this directive as a fatal error and refuses to start: `"OpenBao has dropped support for mlock. Please remove the line "disable_mlock" = false from your config"`.
- Flips the `TestGenerate_OpenBaoConfigHCL_GeneratedForProd` assertion from must-contain to must-not-contain `disable_mlock` — this is the regression test per the artifact-test pattern.
- Adds CHANGELOG entry under `[Unreleased] ### Fixed`.

## Root cause

The directive was originally added to disable memory-locking (mlock) on Linux systems where the container lacks `IPC_LOCK`. OpenBao 2.2.0 removed mlock support entirely rather than deprecating it, so the config key is now unrecognised and causes a startup failure. Since the generated compose already adds the `IPC_LOCK` capability to the openbao service in prod mode, removing the directive has no production impact — mlock would have been enabled, not disabled.

## Test plan

- `go test -race -run TestGenerate_OpenBaoConfigHCL ./internal/app/generate/...` — passes; the not-contains assertion catches any future regression where the directive is reintroduced.
- `make check` — clean (gofmt, golangci-lint, build, full test suite, demo-app).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This initial entry was written by hand to summarise the work leading up to v0.1.
 
 ### Fixed
 
+- **fix(#1334): bundle — remove `disable_mlock = false` from generated `openbao/config.hcl`.** OpenBao 2.2.0+ rejects this directive ("OpenBao has dropped support for mlock"). Every Linux deploy with `secrets.provider: openbao` (i.e., every default deploy) failed at startup. Surfaced by the v0.19 smoke test on demo.vibewarden.dev (Hetzner).
+
 - **fix(#1304): cap TLS retry loop at a hard iteration ceiling.** Previously the retry path spun 500K–750K iterations on persistent failure with no cap, observable in the exhaustion test (the `noSleep` test seam removes the natural pacing of `time.Sleep`). New `MaxIterations(budget, poll)` helper returns `floor(budget/poll) + 2` and gates both `runTLSRetry` and the boot-gap loop. Production iteration counts stay in the low tens (30s budget / 2s poll → 17 max). Returns `ErrTLSRetryExhausted` cleanly after the cap is hit.
 
 ### Added
--- a/internal/app/generate/service_test.go
+++ b/internal/app/generate/service_test.go
@@ -2244,8 +2244,9 @@ func TestGenerate_OpenBaoConfigHCL_GeneratedForProd(t *testing.T) {
 	if !bytes.Contains(data, []byte("api_addr")) {
 		t.Errorf("openbao/config.hcl missing api_addr: %s", data)
 	}
-	if !bytes.Contains(data, []byte("disable_mlock")) {
-		t.Errorf("openbao/config.hcl missing disable_mlock: %s", data)
+	// OpenBao 2.2.0+ rejects the disable_mlock directive entirely; it must not appear.
+	if bytes.Contains(data, []byte("disable_mlock")) {
+		t.Errorf("openbao/config.hcl must not contain disable_mlock: OpenBao 2.2.0+ rejects this directive: %s", data)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


tibtof:
**Reviewer Agent: APPROVED**

All five checklist items pass.

**Template fix** — the diff removes `disable_mlock = false` from `internal/config/templates/openbao-config.hcl.tmpl` and also tightens the alignment of `api_addr` (cosmetic, fine).

**Compose / IPC_LOCK** — `internal/config/templates/docker-compose.yml.tmpl` line 223-224 already carries `cap_add: [IPC_LOCK]` unconditionally under the openbao service block (pinned to image `quay.io/openbao/openbao:2.2.0`). Removing `disable_mlock = false` from the HCL config is not a functional regression: OpenBao will still be granted the `CAP_IPC_LOCK` Linux capability by the container runtime and will use mlock by default, exactly as intended.

**Artifact test** — `TestGenerate_OpenBaoConfigHCL_GeneratedForProd` is a genuine artifact test: it calls `renderComposeAndDir`, reads the rendered file from disk with `os.ReadFile`, and asserts on the raw bytes. The assertion is correctly inverted to `bytes.Contains(...) == true` triggering a failure if `disable_mlock` is present. The error message is descriptive and includes the rendered content for diagnosis.

**OpenBao 2.2.0 compatibility** — OpenBao 2.2.0 dropped the mlock subsystem entirely and treats the `disable_mlock` directive as an unrecognised config key, causing a fatal startup error. Removing the directive (rather than setting it to `true`) is the correct fix: with `IPC_LOCK` granted, the process can lock memory; OpenBao manages that internally without the config key. No positive directive is required.

**CHANGELOG** — `[Unreleased] ### Fixed` entry present with issue reference, root-cause explanation, and smoke-test attribution. Correct position above the existing v0.19 fix entry.

**Sibling check** — `disable_mlock` appears in exactly three non-worktree files: the template (fix applied), the test (assertion inverted), and CHANGELOG.md (new entry). No docs, LLM content files, ADRs, or YAML configs reference `disable_mlock`. Nothing else to clean up.

**CI** — checks in flight (pending); Go Vulnerability Check already passed. No failures at time of review.

tibtof:
Documentation review: docs consistent with code changes.

## Checklist

1. **CHANGELOG.md `[Unreleased] ### Fixed` entry** — present and accurate. The entry correctly names the directive (`disable_mlock = false`), cites the OpenBao 2.2.0+ rejection error string, states the blast radius (every default Linux deploy with `secrets.provider: openbao`), and references the smoke-test surface. No issues.

2. **Cross-doc mlock/disable_mlock search** — zero hits in `docs/`, `llms-full.txt`, `vibewarden.reference.yaml`, `AGENTS-VIBEWARDEN.md`, `README.md`. No stale documentation to clean up.

3. **OpenBao docs mlock claim check** — `docs/secret-management.md` and `llms-full.txt` both reference OpenBao extensively (10+ files) but neither mentions `disable_mlock`, `mlock`, or any claim about the HCL config key. No stale claims.

4. **Operator migration note** — the CHANGELOG entry states the directive is removed from the generated `openbao/config.hcl`, which is sufficient. `openbao-config.hcl` is a generated artifact (not user-edited per the artifact policy); operators who hand-pinned a copy are operating outside the supported workflow. The entry provides the information needed to identify and fix a pinned copy. Adequate.

5. **Cross-contamination check** — CHANGELOG diff adds exactly two lines, both scoped to `#1334`. No other issue numbers in the added content. Clean.

VERDICT: approved

PR #1354, opened 2026-05-07
fix(#1353): demo-app — derive tls_enabled from X-Forwarded-Proto header

tibtof:
Closes #1353

## Summary

- `handleProfile` in `examples/demo-app/main.go` previously computed `tls_enabled` as `profile == "tls" || profile == "prod"`. This is wrong on two axes: the env var was never propagated to the app container, and profile is not the right axis anyway.
- VibeWarden (Caddy) terminates TLS by default in **all** profiles — self-signed in dev, Let's Encrypt in prod. The only time `tls_enabled` should be false is when an external terminator (Cloudflare, ALB) sits in front.
- The correct signal is `X-Forwarded-Proto`. Caddy sets it to `https` on every request it forwards after terminating TLS. The four cases now handled correctly:
  1. `vibew dev` with TLS (default) → Caddy sets header → `tls_enabled: true`
  2. `vibew dev` with `VIBEWARDEN_TLS_ENABLED=false` → no header → `tls_enabled: false`
  3. prod Let's Encrypt → Caddy sets header → `tls_enabled: true`
  4. external terminator (Cloudflare/ALB) → `X-Forwarded-Proto: https` still set by the upstream proxy → `tls_enabled: true` (correctly reflects that TLS is in use end-to-end)
- Stale `VIBEWARDEN_PROFILE=tls` / `VIBEWARDEN_PROFILE=prod` hint copy removed from `static/index.html`.
- `observability_enabled` is left on the profile-name axis — it reflects Compose profile state, not request headers, and there is no better signal.

## Files changed

- `/examples/demo-app/main.go` — `handleProfile` logic
- `/examples/demo-app/main_test.go` — `TestHandleProfile` with 3 table cases
- `/examples/demo-app/static/index.html` — TLS card copy + JS logic
- `/examples/demo-app/README.md` — updated `/profile` example output + note
- `/examples/demo-app/.env.example` — comment clarifying the right axis
- `CHANGELOG.md` — `[Unreleased] ### Fixed` entry

## Test plan

- `TestHandleProfile` covers: `X-Forwarded-Proto: https` → true, `X-Forwarded-Proto: http` → false, no header → false
- `make check` passes clean (lint, build, all tests including demo-app)


tibtof (commented):
Documentation review: one doc-drift item requires a fix before approval.

**Documentation drift: `llms-full.txt` profile values list is stale**

`llms-full.txt` line 260:

```
Values: dev, tls, prod
```

The `.env.example` diff in this PR adds `full` and `observability` as accepted `VIBEWARDEN_PROFILE` values:

```
# Accepted values: dev | tls | prod | full | observability
```

`llms-full.txt` is the primary AI-agent-consumable config reference. An agent reading it will not know `full` or `observability` are valid profile values and will be unable to generate a correct `.env`. Update line 260 to:

```
Values: dev, tls, prod, full, observability
```

---

Everything else checks out:

- CHANGELOG `[Unreleased]` `### Fixed` entry present and scoped to #1353 only.
- `examples/demo-app/README.md` `/profile` example output updated to `tls_enabled: true`. Accurate.
- `.env.example` comment correctly explains that `tls_enabled` is derived from `X-Forwarded-Proto`, not from `VIBEWARDEN_PROFILE`. The axis distinction is clear.
- `static/index.html`: "Start with `VIBEWARDEN_PROFILE=tls`" hint text removed. Replacement copy (external terminator case) is accurate and consistent with the code change.
- Cross-doc grep for `VIBEWARDEN_PROFILE=tls` and `VIBEWARDEN_PROFILE=prod` in `*.md`/`*.html` under the main tree (excluding worktrees): hits in `docs/secret-management.md`, `decisions/pm-log.md`, and `decisions/adr-104-*.md` all use `VIBEWARDEN_PROFILE=prod` in the context of seed-secrets prod-branch detection — not TLS gating — so they are correct and unaffected.
- `vibewarden.reference.yaml`: `profile` field comment mentions Compose profiles, not TLS gating. No update needed.
- `llms-full.txt`: no mention of profile gating TLS. The only drift is the incomplete values list noted above.
- CHANGELOG diff contains no cross-contamination from other issues.

PR #1394, opened 2026-06-11
fix(#1393): close admin-API auth bypass and sidecar crash-loop

tibtof:
Closes #1393

## Summary

Two independent critical bugs existed whenever `admin.enabled: true`:

**1. Admin-API auth bypass (security)** — `buildAdminRoute` in `config_routes.go` emitted a dedicated `/_vibewarden/admin/*` Caddy route whose `handle` chain was only `[reverse_proxy]`. Because Caddy evaluates routes top-to-bottom and stops at the first match, the `vibewarden_admin_auth` handler sitting in the catch-all chain never ran for admin paths. Every `GET /_vibewarden/admin/users` request was proxied to the internal server regardless of the `X-Admin-Key` value — confirmed auth bypass.

**2. Sidecar crash-loop** — `usermgmt.Plugin.ContributeCaddyHandlers()` emitted `{"handler":"admin_auth",...}` with the wrong Caddy module name (registered name is `"vibewarden_admin_auth"`). `caddy.Load` rejected the unknown handler name and crashed on every startup when admin was enabled.

## Changes

- **`internal/adapters/caddy/admin_auth_handler.go`** — `buildAdminAuthHandlerJSON` now serialises `ConfigPath` (previously dropped), so the inlined gate also protects `/_vibewarden/config/*`.
- **`internal/adapters/caddy/config_routes.go`** — `buildAdminRoute` accepts `ports.AdminAuthConfig` and places `vibewarden_admin_auth` **first** in the handle chain before `reverse_proxy`. Returns `(map[string]any, error)`. Stale comment claiming "auth handler has already validated" removed.
- **`internal/adapters/caddy/config_build.go`** — `buildRoutes` threads the error from `buildAdminRoute` and now returns `([]map[string]any, error)`.
- **`internal/adapters/caddy/config.go`** — `BuildCaddyConfig` propagates the `buildRoutes` error.
- **`internal/plugins/usermgmt/plugin.go`** — `ContributeCaddyRoutes()` and `ContributeCaddyHandlers()` are documented no-ops. The divergent handler map and now-unused path constants removed. Struct doc updated.

## Single source of truth

The canonical admin route (`/_vibewarden/admin/*`) is now built exactly once by `buildAdminRoute` in the caddy adapter. No plugin contribution can reintroduce a duplicate unauthenticated route.

## Ordering note for #1391

#1391 (embedded admin UI with tokenless `/_vibewarden/admin/ui` carve-out) must merge **after** this PR. The carve-out is only safe once the gate actually runs in the request path — which this PR fixes. The `adminUIPrefix` logic in #1391's middleware composes correctly on top of this gate.

## Test plan

- Unit: `buildAdminRoute` handle-chain order (`vibewarden_admin_auth` before `reverse_proxy`), `buildAdminAuthHandlerJSON` config_path serialisation, `usermgmt.Plugin.ContributeCaddyRoutes()/ContributeCaddyHandlers()` always-nil locks.
- Integration (`//go:build integration`): `TestAdminAuth_Integration_GateEnforced` asserts through a running Caddy instance:
  - `GET /_vibewarden/admin/users` no token → **401**, upstream not hit
  - `GET /_vibewarden/admin/users` wrong token → **401**, upstream not hit
  - `GET /_vibewarden/admin/users` correct token → **200**, upstream hit
  - `GET /_vibewarden/api/docs` no token → upstream hit (public route unchanged)
- `TestAdminAuth_Integration_CleanStart`: asserts `adapter.Start` succeeds (no crash-loop) and `/_vibewarden/health` returns 200.
- `make check` passed locally (gofmt, golangci-lint, go test -race ./...).


internal/adapters/caddy/config_build.go
```diff
@@ -164,7 +169,11 @@ func buildRoutes(cfg *ports.ProxyConfig, handlers []map[string]any) []map[string
 	}
 
 	if cfg.Admin.Enabled && cfg.Admin.InternalAddr != "" {
-		routes = append(routes, buildAdminRoute(cfg.Admin.InternalAddr))
+		adminRoute, err := buildAdminRoute(cfg.Admin.InternalAddr, cfg.AdminAuth)
+		if err != nil {
+			return nil, fmt.Errorf("building admin route: %w", err)
+		}
+		routes = append(routes, adminRoute)
 		routes = append(routes, buildDocsRoute(cfg.Admin.InternalAddr))
```
tibtof, line 185: **Regression: `/_vibewarden/config/*` route was dropped and not replaced.**

On `main`, `usermgmt.Plugin.ContributeCaddyRoutes()` emitted a dedicated `/_vibewarden/config/*` reverse_proxy route (priority 61) to the internal admin server. This PR removed that contribution (correctly — it was unauthenticated) but did **not** reconstruct a config route here in the caddy adapter the way it did for the admin route. I confirmed against the assembled config: the only routes emitted when admin is enabled are `health`, `ready`, `/_vibewarden/admin/*`, `/_vibewarden/api/docs`, and the catch-all (`match: null`). There is **no** `/_vibewarden/config/*` route.

The internal admin server actually serves these endpoints (`internal/adapters/http/admin_handlers.go:77-78`):
```
POST /_vibewarden/config/reload
GET  /_vibewarden/config
```

Consequences after this PR (admin enabled):
1. `POST /_vibewarden/config/reload` matches the gate prefix `/_vibewarden/config/`, so it is token-gated — but after auth passes it falls through to the catch-all and is reverse-proxied to the **upstream app**, not the internal admin server. The reload endpoint is mis-routed/broken.
2. `GET /_vibewarden/config` has **no trailing slash**, so `strings.HasPrefix(path, "/_vibewarden/config/")` is false — it is **not even gated**, and also lands on the upstream app.

This contradicts the PR's own claim (CHANGELOG + the `ConfigPath` plumbing) that `/_vibewarden/config/*` is protected: the gate config carries `config_path`, but the requests can no longer reach the server they target.

**Fix:** emit a canonical config route here mirroring `buildAdminRoute` — `match: /_vibewarden/config/*`, handle chain `[vibewarden_admin_auth, reverse_proxy → cfg.Admin.InternalAddr]` — placed before the catch-all. And reconcile the prefix vs the `GET /_vibewarden/config` (no-slash) endpoint so both the gate and the route cover it (either match `/_vibewarden/config*` or register the GET endpoint under the trailing-slash prefix). Add an integration assertion: `GET/POST /_vibewarden/config*` with no/wrong token → 401, with correct token → reaches the **internal admin** mock (not the app mock).

PR #1401, opened 2026-06-21
fix(#1279): read user identity from context to eliminate per-user rate-limit ordering dependency

tibtof:
Closes #1279

## Summary

- **Root cause investigated:** `RateLimitMiddleware` resolved the authenticated user ID solely from the `X-User-Id` header (set by `IdentityHeadersMiddleware`). In the standalone `net/http` wiring path, if `RateLimitMiddleware` ran before `IdentityHeadersMiddleware`, the header was absent and per-user limiting was silently skipped. Authenticated users fell through to per-IP limits only.
- **Caddy path confirmed correct:** `vibewarden_authentication` at priority 40, `vibewarden_rate_limit` at priority 50 — ordering is already correct in production Caddy wiring. The bug was latent fragility in the `net/http` path.
- **Fix:** Added `resolveUserID()` helper that prefers the domain `Identity` in the request context (set by `AuthMiddleware` via `IdentityFromContext`) and falls back to the `X-User-Id` header. This eliminates the ordering dependency entirely.
- **CHANGELOG:** Entry under `## [Unreleased] → ### Fixed` referencing #1279.

## Test plan

- `TestRateLimitMiddleware_ContextIdentityUsedBeforeHeader` — table-driven, covers: context-only (no header), header-only (no context), context-wins-over-header precedence, unauthenticated (neither source).
- `TestRateLimitMiddleware_ContextIdentityTriggersUserLimit_BehavioralGuard` — crux behavioral guard: authenticated user exceeding per-user limit gets 429 even when `X-User-Id` header is absent; user limiter called with the correct user ID from context. Guards against the silent no-op regression.
- All existing rate limit tests continue to pass (`make check` clean).


internal/middleware/ratelimit.go
```diff
@@ -84,7 +96,12 @@ func RateLimitMiddleware(
 			}
```
tibtof, line 96: **Pre-existing audit gap (not introduced by this PR, not blocking):** When a per-user rate limit fires, `emitAuditRateLimitHit` (line 92 in the pre-PR main) is called as `emitAuditRateLimitHit(r, auditLogger, drops, "user", clientIP, userResult)`. The function signature takes `clientIP` as the `clientIP` field in the `Actor`, but does not accept or record the `userID`. That means audit events for user-triggered rate limit hits have `actor.user_id = ""` — a forensics gap.

This is pre-existing behavior, unchanged by this PR (the PR only changes how `userID` is *resolved*, not what is done with it downstream). Recommend tracking a follow-up issue to pass `userID` into `emitAuditRateLimitHit` and include it in `audit.Actor.UserID`, but this must not block the current fix.

tibtof: Acknowledged — pre-existing and out of scope for this PR. Tracked as #1402. Resolving.

PR #1403, opened 2026-06-21
fix(#1281): reject path-traversal injection in secret:// URIs (OWASP A03)

### LINKED ISSUE #1281: audit(security): secret:// URI path is not validated — path traversal via malformed URIs in config fields
## Finding

**Severity**: medium
**OWASP Category**: A03:2021 Injection

### Description

`internal/domain/secret/uri.go` (`ParseURI`) accepts any path component in a `secret://path/key` URI without sanitizing `..` segments or absolute path markers. The parsed `uri.Path()` is then passed directly to the secret store backend:

```go
data, err := store.Get(ctx, uri.Path())
```

For the **builtin** secret store (`internal/adapters/builtin/store.go`), the `Get` method uses the path as a map key (no filesystem access), so there is no direct traversal risk in the builtin backend.

However, the **OpenBao** adapter (`internal/adapters/openbao/`) constructs HTTP paths from the URI path component and sends them to the OpenBao server. A crafted URI such as `secret://../sys/mounts/secret/key` could cause the adapter to send requests to unintended OpenBao API paths (e.g. `GET /v1/../sys/mounts/secret` → `GET /v1/sys/mounts/secret`).

Additionally, the `List` operation in the builtin store uses a bare `strings.HasPrefix(path, prefix)` check (line 131):

```go
if !strings.HasPrefix(path, prefix) {
    continue
}
```

This is the same prefix-extension bug: listing prefix `"auth"` also returns entries for `"auth-evil"`. While only a logic bug (no security boundary crossed in a trusted store), it confirms the pattern.

### Suggested fix

1. In `ParseURI`, reject paths containing `..`, absolute path markers, or encoded slashes (`%2F`):
```go
if strings.Contains(body, "..") || strings.HasPrefix(body, "/") {
    return URI{}, fmt.Errorf("secret URI %q contains disallowed path component", raw)
}
```
2. In the builtin `List`, use `strings.HasPrefix(path, prefix+"/") || path == prefix` to avoid the prefix-extension issue.
3. In the OpenBao adapter, use `path.Clean` / `url.PathEscape` when constructing API request URLs.

### Source

- `internal/domain/secret/uri.go` — `ParseURI` (no path sanitization, lines 32–58)
- `internal/adapters/builtin/store.go` — `List` (bare `HasPrefix`, line 131)
- `internal/adapters/openbao/adapter.go` — uses `uri.Path()` to build OpenBao request URLs

tibtof:
Closes #1281

## Summary

Three coordinated hardening changes for OWASP A03 Injection in `secret://` URI handling:

- **`ParseURI` (`internal/domain/secret/uri.go`):** Rejects URIs containing `..` path segments (path traversal), a leading `/` (absolute paths), or percent-encoded slashes (`%2F`/`%2f`). Segment-based check via `strings.Split` so a key legitimately named `a..b` is accepted while the traversal segment `..` alone is blocked. Closes the primary vector: `secret://../sys/mounts/secret/key` could reach unintended OpenBao API paths.

- **`List` (`internal/adapters/builtin/store.go`):** Replaces bare `strings.HasPrefix(path, prefix)` with `path == prefix || strings.HasPrefix(path, prefixDir)` where `prefixDir` is the prefix normalised to end with `"/"`. Fixes prefix-extension: prefix `"auth"` no longer matches stored path `"auth-evil"`.

- **`validateSecretPath` (`internal/adapters/openbao/adapter.go`):** New unexported helper enforcing the same three rules; called at the top of `Get`, `Put`, `Delete`, `List`, and `GetMetadata`. Defense-in-depth for callers that supply a path directly without going through `ParseURI`.

## Test plan

All tests assert observable behavior, not string shape:

- `TestParseURI` — 8 new cases: `..` traversal (3 variants), leading `/`, `%2F` uppercase, `%2f` lowercase, valid `a..b` substring, valid `.` segment
- `TestStore_List_PrefixExtension` — populates `auth`, `auth-evil`, `auth/token`; asserts `List("auth")` never returns `"-evil"` and `List("auth/")` returns `"token"` only
- `TestAdapter_Get_MaliciousPath` — 5 crafted paths; httptest server fails the test if any request arrives (adapter must error before HTTP call)
- `TestAdapter_Put_MaliciousPath` — same pattern for `Put`

`make check` (golangci-lint + go test -race ./...): all checks passed.


internal/domain/secret/uri.go
```diff
@@ -39,6 +43,26 @@ func ParseURI(raw string) (URI, error) {
 		return URI{}, errors.New("secret URI is empty after scheme: expected secret://path/key")
 	}
 
+	// Reject percent-encoded slashes — they are decoded by HTTP clients into
+	// path separators and bypass segment-based validation downstream (e.g. in
+	// the OpenBao adapter, which builds HTTP API paths from the URI path).
```
tibtof: **Security gap: double-encoded slash (`%252F`) bypasses this check.**

`strings.Contains(body, "%2F")` catches a literal percent-encoded slash, but not a *double*-encoded one. When `body` contains `auth%252F../sys`, this check passes. Go's `net/http` client calls `url.Parse` on the constructed URL, which decodes `%25` → `%` in the path field, leaving `auth%2F../sys`. Whether OpenBao then treats `%2F` as a path separator depends on its router, but double-encoding bypass is a named attack category (OWASP A03) and must be explicitly tested or explicitly excluded.

**Concrete fix:** reject any `%` character in the body, since legitimate KV secret names never need percent-encoding:

```go
if strings.ContainsRune(body, '%') {
    return URI{}, fmt.Errorf("secret URI %q contains a percent-encoded character; use only literal characters in the path", raw)
}
```

This collapses the `%2F`/`%2f` check into a single, future-proof rule that also blocks `%2e` → `.` and any other encoding tricks.

internal/adapters/openbao/adapter.go
```diff
@@ -571,6 +586,28 @@ func (a *Adapter) RevokeLease(ctx context.Context, leaseID string) error {
...
 
+// validateSecretPath checks that p is safe to interpolate into an OpenBao KV
+// API path. It rejects paths that contain ".." segments, a leading "/", or
+// percent-encoded slashes (%2F / %2f).
+//
+// ParseURI enforces the same constraints at parse time, so this function is
+// defense-in-depth for callers that pass a path directly to the adapter
+// without going through ParseURI.
+func validateSecretPath(p string) error {
+	if strings.Contains(p, "%2F") || strings.Contains(p, "%2f") {
```
tibtof: **Same double-encoding bypass as in `ParseURI`.**

`%252F` passes this check and reaches `doRaw`, where `url.Parse` decodes `%25` → `%`, producing a path containing a literal `%2F` that OpenBao may further decode. As the defense-in-depth layer, this function should be at least as strict as `ParseURI`.

Apply the same fix: reject any `%` instead of only `%2F`/`%2f`:

```go
if strings.ContainsRune(p, '%') {
    return fmt.Errorf("openbao: path %q contains percent-encoded characters (not allowed)", p)
}
```

Also add a test case to `TestAdapter_Get_MaliciousPath` for `"auth%252Fgoogle"` to prove this guard is in place.

internal/adapters/openbao/adapter_test.go
```diff
@@ -366,3 +366,63 @@ func TestDynamicCredentials_ExpiresAt(t *testing.T) {
...
+	defer srv.Close()
+
+	a := newTestAdapter(srv.URL, openbao.AuthMethodToken, "test-token")
+	_ = a.Authenticate(context.Background())
+
+	err := a.Put(context.Background(), malicious, map[string]string{"k": "v"})
+	if err == nil {
+		t.Errorf("Put(%q) expected error for malicious path, got nil", malicious)
+	}
+}
```
tibtof, line 441: **Missing malicious-path tests for `Delete`, `List`, and `GetMetadata`.**

`validateSecretPath` is called in all five methods but only `Get` and `Put` have a guard test. If the call were accidentally removed from `Delete`, `List`, or `GetMetadata`, no test would catch the regression.

Per CLAUDE.md: "Unit tests for all domain logic and middleware." Add three more tests following the same httptest pattern — one malicious path per method is enough to prove the guard is wired up:

```go
func TestAdapter_Delete_MaliciousPath(t *testing.T) { ... }
func TestAdapter_List_MaliciousPath(t *testing.T) { ... }
func TestAdapter_GetMetadata_MaliciousPath(t *testing.T) { ... }
```

A single sub-case (e.g., `"../sys/mounts/secret"`) per function is sufficient.

internal/adapters/openbao/adapter_test.go
```diff
@@ -366,3 +366,63 @@ func TestDynamicCredentials_ExpiresAt(t *testing.T) {
...
+	malicious := "../sys/mounts/secret"
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		t.Errorf("adapter sent HTTP request for malicious path %q; URL path = %q", malicious, r.URL.Path)
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer srv.Close()
+
+	a := newTestAdapter(srv.URL, openbao.AuthMethodToken, "test-token")
+	_ = a.Authenticate(context.Background())
```
tibtof: **Swallowed error — inconsistent with `TestAdapter_Get_MaliciousPath`.**

`_ = a.Authenticate(...)` silently drops the return value. For `AuthMethodToken` this can never fail in practice, but the pattern violates CLAUDE.md ("No swallowed errors") and is inconsistent with the `Get` malicious-path test directly above, which properly calls `t.Fatalf` on authenticate failure.

Change to:
```go
if err := a.Authenticate(context.Background()); err != nil {
    t.Fatalf("Authenticate() error = %v", err)
}
```

PR #1404, opened 2026-06-21
fix(#1274): extend symlink-containment check to staleness walker (OWASP A01)

### LINKED ISSUE #1274: audit(security): staleness walker missing symlink escape check — sibling of #1223 fix not applied
## Finding

**Severity**: medium
**OWASP Category**: A01:2021 Broken Access Control

### Description

The symlink escape fix applied to `input_digest.go` in PR #1223 was **not applied** to the staleness walker in `internal/app/bundle/staleness.go`.

`staleness.go` uses `filepath.WalkDir` but does not resolve symlinks or verify that symlink targets remain within the project root. An attacker who can create a symlink inside the project root pointing outside of it (e.g. `project/.vw-link -> /etc/`) can cause the staleness walker to:

1. Count files outside the project root toward the "changed files" counter.
2. Include modification times of sensitive system files when computing the freshness threshold.

While the staleness walker only reads file `mtime` (not content), the side effect is that a symlink to a frequently-updated external file (e.g. `/var/log/syslog`) always marks the project as "stale", breaking `vibew dev` blocking logic that depends on this signal.

Compare with the fix in `input_digest.go` lines 226–243:

```go
if resolved != absRoot && !strings.HasPrefix(resolved, absRoot+string(os.PathSeparator)) {
    slog.Debug("input-digest: symlink escapes project root, skipping", ...)
    return nil
}
```

No equivalent check exists in `staleness.go`.

### Reproduction

```bash
cd /my/project
ln -s /var/log /my/project/evil-link
# vibew dev will now count /var/log file mtimes and may behave unpredictably
```

### Why it matters

While not a direct data-exfiltration path (no content is read), this is a sibling bug to a previously-fixed vulnerability (#1223). The fix was applied inconsistently — the audit found that only `input_digest.go` received it. A complete fix requires the same separator-terminated prefix check in all walkers.

### Suggested fix

In `FileSystemStalenessWalker.NewestMTime`, resolve symlinks for each entry and check containment before processing:

```go
if d.Type()&os.ModeSymlink != 0 {
    resolved, err := filepath.EvalSymlinks(path)
    if err != nil {
        return nil
    }
    if resolved != root && !strings.HasPrefix(resolved, root+string(os.PathSeparator)) {
        return nil // skip: symlink escapes root
    }
}
```

Note: `filepath.WalkDir` does not follow symlinks by default, so the check is only needed when the entry type includes `os.ModeSymlink`. Verify the walk mode being used.

### Source

- `internal/app/bundle/staleness.go` — `FileSystemStalenessWalker.NewestMTime` (line 69), no symlink escape check
- `internal/app/bundle/input_digest.go` — lines 226–243, reference fix from #1223

tibtof:
Closes #1274

## Summary

- PR #1223 added a symlink-escape containment check to `computeInputDigest` (`input_digest.go`) but did not apply it to `FileSystemStalenessWalker.NewestMTime` (`staleness.go`).
- A symlink inside the project root pointing outside (e.g. to `/var/log`) caused the walker to count the symlink's own mtime (via `d.Info()` / Lstat), which could produce a spurious always-stale signal in `vibew dev`.
- This PR applies the identical `filepath.EvalSymlinks` + separator-terminated prefix check to the staleness walker, resolving `absRoot` once outside the walk for efficiency. Escaping symlinks are skipped with a debug-log; in-root symlinks are unaffected.

**No shared containment helper existed** — the check in `input_digest.go` is inline. The fix is added inline in `staleness.go` to match that exact pattern (same variable names, same log keys, same separator-terminated prefix guard against the `/proj-secret` prefix-extension attack).

## Test plan

Three behavioral tests added to `staleness_test.go`:

- `TestStalenessWalker_SymlinkOutsideRootSkipped` — symlink inside root pointing to an outside file is skipped; `changedCount` stays 0 despite the symlink having `mtime=now`. FAILS without the fix, PASSES with it.
- `TestStalenessWalker_SymlinkPrefixExtensionRejected` — symlink to a sibling directory whose name extends the root name (`/proj-secret` when root is `/proj`) is rejected by the separator-terminated prefix check. FAILS without the fix, PASSES with it.
- `TestStalenessWalker_InRootSymlinkNotSkipped` — symlink whose resolved target stays within the root is NOT skipped; `changedCount > 0` as expected.

`make check` passed clean (0 lint issues, all tests green including `-race`).


internal/app/bundle/staleness.go
```diff
@@ -101,15 +108,36 @@ func (w *FileSystemStalenessWalker) NewestMTime(root string, threshold time.Time
...
-			slog.Debug("staleness walk: cannot stat file", "path", path, "err", err)
+		// Resolve symlinks to catch out-of-root targets (OWASP A01 / #1274).
+		// filepath.WalkDir uses Lstat, so symlink entries are reported with
+		// ModeSymlink set rather than being followed automatically. We resolve
+		// the target here and skip any entry whose resolved path escapes
+		// absRoot. The exact-directory check (absRoot + separator) prevents
+		// the prefix-extension attack (e.g. /proj-secret admitted when root
+		// is /proj via a bare strings.HasPrefix). Mirrors the equivalent
+		// containment check in computeInputDigest (input_digest.go).
+		if d.Type()&os.ModeSymlink != 0 {
```
tibtof, line 125: **Non-blocking / DRY note**: The separator-terminated prefix containment check is now duplicated inline in both `computeInputDigest` (`input_digest.go`) and `NewestMTime` (`staleness.go`). The PR description explicitly acknowledges this and justifies it by matching the existing inline pattern from #1223 — that's the right call for a targeted security fix. Accepted here.

A follow-up task worth creating: extract a `containsPath(absRoot, resolved string) bool` helper into `internal/app/bundle/containment.go` with its own table-driven tests. If a third walker or a future change to the separator logic is needed, there will be a single place to fix. The helper is small (~3 lines of logic) and the extraction is low risk.

tibtof: Accepted as non-blocking — keeping the check inline to match the existing #1223 pattern and not delay the security fix. Tracked the helper extraction in #1405. Resolving.

PR #1406, opened 2026-06-21
fix(#1276): add health.expose_version to suppress version fingerprinting (OWASP A05)

tibtof:
Closes #1276

## Summary

- Adds `health.expose_version` config field (default `true` — no behavior change for existing deployments). Set to `false` to omit the `"version"` key from `/_vibewarden/health` JSON.
- Decouples port-ownership detection from the version string: the health handler now always emits `X-Vibewarden: 1` response header, which `port_owner.go` checks first. The old body-prefix match (`{"status":"ok","version":`) is kept as a backward-compat fallback for pre-fix sidecars during rolling upgrades.
- New `internal/config/health.go` with `HealthConfig` struct; `Health HealthConfig` field added to `Config`; default registered in `setDefaults`.
- `HealthHandler` gains `suppressVersion bool` field (zero = show version, backward compat). `RuntimeServices` gains `SuppressVersion bool`. Wired end-to-end in `buildRuntimeServices` from `cfg.Health.ExposeVersion`.
- `vibewarden.reference.yaml` and `llms-full.txt` updated with the new `health.expose_version` entry. CHANGELOG updated under `## [Unreleased] > ### Security`.

## Test plan

- `TestHealthHandler_ExposeVersion`: assert `version` key present when `suppressVersion=false`, absent (not merely empty) when `suppressVersion=true`.
- `TestHealthHandler_IdentityHeader`: `X-Vibewarden: 1` present regardless of `suppressVersion`.
- `TestHealthHandler_ProvisionWith_SuppressVersion`: `SuppressVersion` wired correctly from `RuntimeServices`.
- `TestVibeWardenHealthProbe_SuppressedVersion`: sidecar with no version in body still detected as VibeWarden via header.
- `TestVibeWardenHealthProbe_LegacyBackwardCompat`: older sidecar (version in body, no header) still detected via body-prefix.
- `TestLoad_HealthExposeVersionDefault`, `TestLoad_HealthExposeVersion`, `TestLoadStrict_HealthExposeVersion_Accepted`: config parsing, defaults, and strict schema validation.
- `TestReferenceYAML_UnmarshalsCleanly` and `TestSetDefaults_EmptyYAML` updated with `health.expose_version` sentinels.
- `make check` passes (gofmt, golangci-lint, go build, go test -race ./...).


internal/adapters/ops/port_owner.go
```diff
@@ -9,13 +9,29 @@ import (
 	"strings"
 	"time"
 
+	caddyadapter "github.com/vibewarden/vibewarden/internal/adapters/caddy"
```
tibtof: **Architecture violation**: `internal/adapters/ops/port_owner.go` imports `internal/adapters/caddy` (line 12 of the new file) to pull in the `HealthIdentityHeader` constant.

In the project's hexagonal model adapters are peers — they implement ports and depend only on `internal/ports/` and `internal/domain/`. An ops adapter importing a caddy adapter creates a lateral coupling that is forbidden by the project constitution: "Hexagonal architecture: domain layer has zero external dependencies. All I/O goes through ports."

Why it matters: if the caddy adapter is ever replaced or refactored, the ops adapter silently picks up that change. It also means you cannot compile the ops adapter without pulling in caddy's entire dependency tree.

**Fix:** move the constant to `internal/ports/` (e.g. a new `internal/ports/health.go`):

```go
// HealthIdentityHeader is the response header always emitted by the
// /_vibewarden/health endpoint. It is the stable ownership marker used
// by port_owner probes to detect a VibeWarden sidecar without relying
// on the version string in the body.
const HealthIdentityHeader = "X-Vibewarden"
```

Then import `internal/ports` in both `internal/adapters/caddy/health_handler.go` (use `ports.HealthIdentityHeader`) and `internal/adapters/ops/port_owner.go` (same, no alias needed). The caddy adapter already imports `internal/ports`, so no new import edge is created.

PR #1422, opened 2026-08-04
fix(#1421): CI-blocking set — fsnotify debounce race, grpc v1.82.1, toolchain go1.26.5

tibtof:
Closes #1421

## Summary

Three independent failures that each keep a required check red. They land together because any subset still blocks merges.

1. **Data race in `internal/adapters/fsnotify/watcher.go`** (Build & Test with `-race`). The debounce `time.AfterFunc` callback selected over `<-done` and `ch <- struct{}{}`; when both cases are ready Go picks at random, so the `done` guard never actually prevented a late send racing the watch goroutine's `defer close(ch)`. Fixed structurally: the callback now only signals an internal buffered `fired` channel that is never closed, and the watch goroutine is the sole sender on and closer of `ch`. The `done` channel is gone; pending timers are stopped on shutdown via `defer`. Debounce and shutdown semantics (channel closes on ctx cancel / unrecoverable error) are unchanged.
2. **govulncheck**: `google.golang.org/grpc` v1.81.1 -> v1.82.1 (indirect, via the OTLP exporters), `go mod tidy` clean.
3. **Trivy**: Go toolchain go1.26.4 -> go1.26.5, pinned consistently in `go.mod` (`toolchain`), `Dockerfile` (`golang:1.26.5-alpine`) and the CI workflow `go-version` inputs — same set of pins the go1.26.4 bump touched, so the toolchain stays single-sourced and the image scan does not stay red on the old base image.

CHANGELOG updated (Fixed + Security).

## Test plan

- `/usr/bin/make check` passes on both modules (gofmt, golangci-lint, build, `go test -race ./...`, demo-app).
- New `TestWatcher_CancelDuringDebounce` cancels the context exactly at the debounce deadline over 50 iterations. Verified behavioral: against the pre-fix `watcher.go` it reports `WARNING: DATA RACE` (`closechan` in `Watch.func1.deferwrap1` vs `chansend` in the timer callback); against the fixed version `go test -race -count=2 ./internal/adapters/fsnotify/...` is green.
- Existing `TestWatcher_Debounce`, `TestWatcher_DetectsWriteEvent`, `TestWatcher_ContextCancellation` unchanged and still pass.


go.mod
```diff
@@ -221,7 +221,7 @@ require (
 	google.golang.org/api v0.277.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20260526163538-3dc84a4a5aaa // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260526163538-3dc84a4a5aaa // indirect
-	google.golang.org/grpc v1.81.1 // indirect
+	google.golang.org/grpc v1.82.1 // indirect
```
tibtof, line 224: **Blocking: the CI-blocking set is incomplete — two required checks are still red on this head SHA.**

Issue #1421 is explicit that the whole set must land in one PR because "any subset leaves required checks red". On `77d7c74`, both `Go Vulnerability Check` and `Trivy Vulnerability Scan` (both **required** by the `main` ruleset) are `completed/failure`. The grpc and toolchain bumps did their job — the grpc finding and the go1.26.4 stdlib finding are both gone — but a single remaining dependency keeps both gates red:

- govulncheck: `GO-2026-5970` — infinite loop on invalid input in `golang.org/x/text@v0.37.0`, fixed in `v0.39.0`. Reported as **called** from our code (`caddy.Adapter.Reload` -> `caddy.Load` -> `norm.Form.*`, and `postgres.NewAuditAdapter` -> `sql.Open` -> `norm.Form.*`), so it is not suppressible as unreachable.
- Trivy fs: `CVE-2026-56852` (HIGH) — same `norm.Iter` infinite loop, `golang.org/x/text v0.37.0`, fixed in `0.39.0`.

`golang.org/x/text` is at `v0.37.0` on both `main` and this branch, so this is not a regression introduced here — it is a newly published advisory that landed after #1421 was filed. It still belongs in this PR: as-is this PR cannot merge, and merging it would not unblock the queue, which is its entire purpose.

Fix — one bump clears both gates:

```
go get golang.org/x/text@v0.39.0 && go mod tidy
```

I verified this in a throwaway worktree off `77d7c74`. It transitively upgrades `x/mod 0.35.0->0.37.0`, `x/net 0.55.0->0.56.0`, `x/sync 0.20.0->0.21.0`, `x/tools 0.44.0->0.47.0` (all BSD-3, approved). After it:

- `go build ./...` OK
- `trivy fs --severity HIGH,CRITICAL ... --skip-dirs examples .` -> 0 vulnerabilities, exit 0
- `govulncheck ./...` -> "No vulnerabilities found", exit 0

Please also extend the CHANGELOG Security entry to cover the `x/text` bump — it currently asserts that the grpc + toolchain bumps "clear currently-red repo-wide CI gates that block every merge", which is not true without this third bump.

PR #1435, opened 2026-08-04
fix(#1338): align vibew logs/dev help service names with the generated compose file

tibtof:
Closes #1338

## Summary

`vibew logs --help` advertised `postgres` as a default-stack service, but the generated `docker-compose.yml` names the Kratos database `kratos-db`. An agent following the help text ran `vibew logs postgres` and hit an "unknown service" dead end. (`mailslurper` had already been removed in an earlier pass.)

- **`internal/cli/cmd/logs.go`** — service list now matches the compose template exactly (`vibewarden`, `app`, `kratos`, `kratos-migrate`, `kratos-db`, `seed-users`, `openbao`, `seed-secrets`, `redis`), with the `vibewarden.yaml` condition that emits each one in parentheses, plus a pointer to `vibew obs` for the observability services and a note that an unknown name prints the exact list for the current stack.
- **`internal/cli/cmd/dev.go`** — baseline-stack list now gives the compose service name next to each component instead of the bare product name "PostgreSQL"; the `--volumes` prose/flag text says `kratos-db data` rather than `Postgres data`.
- **`docs/troubleshooting.md`** — two `vibew logs postgres --tail 100` examples corrected to `vibew logs kratos-db`.
- **`internal/cli/cmd/help_service_names_test.go`** (new) — artifact test: parses the service names out of the *rendered* cobra help text and asserts each is declared under `services:` in `internal/config/templates/docker-compose.yml.tmpl`. Also guards the observability names in both directions and keeps a stale-name blocklist (`postgres`, `mailslurper`) with word-boundary matching so prose like "PostgreSQL database" does not false-positive.

No behaviour change beyond help text; `ErrUnknownService` already validated service names at runtime.

## Test plan

- `/usr/bin/make check` passes (gofmt, goimports, golangci-lint, `go test ./...`, demo-app).
- `go test ./internal/cli/cmd/ -run 'Help|LogsCmd|DevCmd'` — the new drift guard fails if the help text and the compose template disagree (verified by running it against the pre-fix help text, which failed on `postgres`).
- Manual: `vibew logs --help` and `vibew dev --help`.


internal/cli/cmd/logs.go
```diff
@@ -39,11 +39,23 @@ Flags:
...
-  kratos       Ory Kratos identity server
-  postgres     PostgreSQL database
+Services in the generated stack. The exact set depends on which features are
+enabled in vibewarden.yaml — the condition is shown in parentheses:
+  vibewarden      VibeWarden sidecar (proxy, auth, TLS)
+  app             your application container
+  kratos          Ory Kratos identity server (auth.mode: kratos)
+  kratos-migrate  one-shot Kratos schema migration (auth.mode: kratos)
+  kratos-db       PostgreSQL database for Kratos (auth.mode: kratos)
+  seed-users      one-shot seeding of auth.users (auth.users set)
```
tibtof: **Non-existent config key in help text**: `auth.users` is not part of the config schema.

The real key is `auth.seed_demo_users` (bool, default `false`) — see `internal/config/auth.go:339` (`SeedDemoUsers bool \`mapstructure:"seed_demo_users"\``) and `vibewarden.reference.yaml:477`. `auth.users` appears nowhere else in the repo.

Why this matters: it reproduces the exact failure mode this PR fixes. The strict loader reports nested unknown keys (`internal/config/strict.go`, cf. `strict_test.go:191` → `config vibewarden.yaml: unknown key(s): tls.dmain`), so an agent that reads this line and adds `auth.users:` to `vibewarden.yaml` gets a hard `vibew validate` / `vibew bundle` failure — a worse dead end than `vibew logs postgres`, because it blocks the bundle rather than printing a recoverable list.

The stated condition is also wrong on its own terms. The compose guard for this service is `auth.mode == kratos && !kratos.external` (`internal/config/templates/docker-compose.yml.tmpl:273`) — it is emitted regardless of any seeding flag. Only the mounted `scripts/seed-users.sh` is gated on `seed_demo_users` (`internal/app/generate/helpers.go:62`). So `seed-users` shows up in `docker compose ps` for every local-Kratos stack, and a reader told "(auth.users set)" will not expect it.

Suggested fix:

```go
  seed-users      one-shot seeding of demo Kratos identities (auth.mode: kratos)
```

and, if you want to keep the seeding-flag detail, put it in the description rather than the condition column (the script is only written when `auth.seed_demo_users: true`).

PR #1437, opened 2026-08-04
fix(#1339): route OTel SDK export errors to slog and dedupe repeats

tibtof:
Closes #1339

## Summary

In dev mode the `observability` compose profile is off, so `otel-collector` is not running. The OTel SDK reports background export failures through `otel.Handle`, whose default handler writes to the stdlib `log` package — and Caddy redirects the stdlib logger into its own logger at **info** level. Result: one `failed to upload metrics: ... lookup otel-collector ...: no such host` line every export interval (~30s), burying real errors in `vibew logs vibewarden`.

- New `internal/adapters/otel/errorhandler.go`: `ErrorHandler` implements `otel.ErrorHandler` over an injected `*slog.Logger`. The first occurrence of each distinct error message is logged at `warn` (a genuinely misconfigured collector stays visible); every later occurrence of that message is demoted to `debug` with a `repeat_count` attribute. Dedup state is kept per message in a table bounded at 64 entries (cleared wholesale on overflow), not just for the previous message: the generated `docker-compose.yml` enables the metric and log OTLP exporters together, so their two export loops fail on different intervals and interleave — single-slot dedup would treat every alternation as new and warn forever. Safe for concurrent use.
- `InstallErrorHandler(logger)` registers it as the global OTel error handler; the metrics plugin calls it in `Init` before any exporter is created, so metric, trace and log export failures all funnel through slog instead of the stdlib logger.
- No new dependencies, no config surface change, no ADR (bug fix, below the ADR threshold).

## Test plan

- `internal/adapters/otel/errorhandler_test.go` — table-driven coverage of the level/dedup ladder (nil error ignored, first = WARN, repeats = DEBUG with counter, interleaved A,B,A,B = WARN, WARN, DEBUG, DEBUG), `TestErrorHandler_TwoExportersStayBounded` replaying the reported two-exporter scenario (log error every tick + metric error every 30th over 120 ticks -> exactly 2 WARN lines), `TestErrorHandler_DedupTableIsBounded` asserting the dedup table stays within its bound under 6400 distinct messages, nil-logger fallback to `slog.Default`, concurrent `Handle` under `-race`, and `otel.Handle` routing after `InstallErrorHandler`.
- `internal/plugins/metrics/plugin_test.go` — `TestPlugin_Init_InstallsSlogOTelErrorHandler` asserts the global handler is the slog-backed one after plugin `Init`.
- `/usr/bin/make check` passes (gofmt, goimports, golangci-lint, full `go test -race`).
- Manual: `vibew dev` without `COMPOSE_PROFILES=observability`, then `vibew logs vibewarden` — expect a single warn line about the unreachable collector instead of one info line per interval.


internal/adapters/otel/errorhandler.go
```diff
@@ -0,0 +1,82 @@
...
+// logged at debug level with a repeat count. Nil errors are ignored.
+func (h *ErrorHandler) Handle(err error) {
+	if err == nil {
+		return
+	}
+
+	msg := err.Error()
+
+	h.mu.Lock()
+	first := msg != h.lastMsg
```
tibtof: **Dedup degenerates to no dedup when two exporters fail at once — which is the exact configuration #1339 reports.**

`internal/config/templates/docker-compose.yml.tmpl` emits `VIBEWARDEN_TELEMETRY_OTLP_ENABLED=true` and `VIBEWARDEN_TELEMETRY_LOGS_OTLP=true` **together** under `{{ if .Observability.Enabled }}`. So every stack that can hit this bug has two independent export loops aimed at the missing `otel-collector`, both funnelling into `otel.Handle`:

- metric `PeriodicReader` — `sdk/metric@v1.44.0/periodic_reader.go:203`, interval `telemetry.otlp.interval` (30s)
- log `BatchProcessor` — `sdk/log@v0.20.0/exporter.go:172`, `dfltExpInterval = 1s`

Their messages differ (`failed to upload metrics: …` vs `failed to upload logs: …`), so with a single-slot `lastMsg` every alternation makes `first == true` and logs at WARN. I ran that sequence against this handler (log error every tick, metric error every 30th tick, 120 ticks ~ 2 minutes of dev traffic): **9 WARN lines, growing linearly and unbounded**. With per-message state it would be 2 WARN lines, ever.

Net effect in the reported scenario: the change swaps one INFO line per interval for a permanent stream of WARN lines — higher severity, still unbounded — so `vibew logs vibewarden` stays noisy and the "logged once at warn" contract in the godoc above and in the CHANGELOG entry does not hold.

The table test only covers `A,A,B,B`, which is why this passes green.

Suggested fix: key the dedup on the message instead of remembering only the previous one — e.g. replace `lastMsg`/`repeats` with `seen map[string]int` guarded by `h.mu`, reset wholesale once it exceeds a small bound (the realistic ceiling is a handful of distinct SDK errors, and the bound keeps it from growing on high-cardinality messages). Then add an `A,B,A,B` case to `TestErrorHandler_Handle` asserting `WARN, WARN, DEBUG, DEBUG`, so this class of regression fails the build.

tibtof: Fixed in 834948c.

Dedup is now keyed on the message: `seen map[string]int` guarded by `h.mu`, cleared wholesale once it exceeds `maxTrackedMessages` (64), so memory stays flat if an exporter ever emits high-cardinality messages.

Tests added:
- `TestErrorHandler_Handle/interleaved_messages_warn_once_each_then_stay_at_debug` — the A,B,A,B case asserting `WARN, WARN, DEBUG, DEBUG` with counts `0,0,1,1`.
- `TestErrorHandler_TwoExportersStayBounded` — replays your scenario (log error every tick, metric error every 30th, 120 ticks): 124 records, exactly 2 WARN, one per distinct message.
- `TestErrorHandler_DedupTableIsBounded` (white-box) — 6400 distinct messages, asserts `len(h.seen) <= maxTrackedMessages`.

Godoc and the CHANGELOG entry now state the per-message contract and why single-slot dedup is wrong here. `/usr/bin/make check` passes.

PR #1438, opened 2026-08-04
docs: add zero-to-secure terminal demo GIF to README top

tibtof:
First thing visitors now see is a 45-second real terminal demo (vibew init → generated files → security config with TLS/headers/rate-limit enabled).

- `docs/assets/demo.tape` — VHS script, checked in so the demo is regenerable and never rots (artifact policy: real, owned, reproducible)
- `docs/assets/demo.gif` — 668K rendered output
- README: GIF embedded under the tagline

No CHANGELOG entry: docs-only, no runtime change. Part of the promote playbook Tier 1 (notes/promote-playbook.md).


docs/assets/demo.tape
```diff
@@ -0,0 +1,45 @@
...
+
+Type "ls -1"
+Enter
+Sleep 2.5s
+
+Type "grep -A2 -E '^(tls|security_headers|rate_limit):' vibewarden.yaml"
+Enter
+Sleep 4s
+
+Type "# TLS, auth, rate limiting, WAF — configured. Now: vibew dev"
```
tibtof: **Incorrect claim in the demo — auth is not configured by a `vibew init` scaffold.**

This is the closing frame of the GIF (verified by extracting the last frame), i.e. the last thing a README visitor reads, and it states that auth is configured. It is not:

- `internal/cli/templates/init-vibewarden.yaml.tmpl` emits no `auth:` block at all — I ran the tape's exact commands against a binary built from this branch and the scaffold contains only `name`, `server`, `upstream`, `app`, `log`, `security_headers`, `tls`, `rate_limit`.
- With no `auth:` block, `internal/config/config.go:341` applies `v.SetDefault("auth.mode", "none")` — the auth middleware is off. Auth requires an explicit `vibew add auth`.
- The repo already corrected this exact claim elsewhere one commit ago: the CHANGELOG entry for fix(#1338) states verbatim that "a default `vibew init` scaffold has no `auth:` block".

A visitor's takeaway from the hero image is that `vibew init` gives them authentication. That is a false security claim, and it contradicts the grep output shown directly above it in the same frame (which lists only tls / security_headers / rate_limit).

Secondary nuance on the same line: `waf.enabled` does default to `true`, but `waf.mode` defaults to `"detect"` (`internal/config/config.go:482-483`) — log-only, not blocking — and the WAF is likewise absent from the file being grepped, so "configured" oversells it too.

Suggested fix (then re-render with `vhs docs/assets/demo.tape` so the GIF matches):

```
Type "# TLS, security headers, rate limiting — on by default. Next: vibew add auth, vibew dev"
```

or, if you want to keep the WAF mention, make the state explicit: `WAF in detect mode`. Either way, drop `auth` or reframe it as the next step rather than something already done.

